PR #2, opened 2017-04-21
Added hostnet/form-handler-bundle/1.2

iltar:
Added the hostnet/form-handler-bundle: https://github.com/hostnet/form-handler-bundle/wiki

### DIFF
--- a/hostnet/form-handler-bundle/1.2/manifest.json
+++ b/hostnet/form-handler-bundle/1.2/manifest.json
@@ -1,0 +1,5 @@
+{
+    "bundles": {
+        "Hostnet\\Bundle\\FormHandlerBundle\\FormHandlerBundle": ["all"]
+    }
+}


fabpot:
Thank you @iltar.

stof:
@fabpot this recipe is now useless

fabpot:
deleted in 3dba6fb

iltar:
Something as simple like this will now be auto-configured?

stof:
@iltar if you have `type=symfony-bundle` in your composer.json, flex will now try to generate a default recipe when there is none in the repository. This recipe is only registering the bundle for all environments.
This works if flex manages to guess the bundle class name (which I think is the case for this bundle)

stof:
Well, actually, the bundle does not define this type, so it would still need a change. But I think this can be done easily in the bundle

iltar:
@stof indeed it's missing, I'll add it.

PR #3, opened 2017-04-21
Added features bundle

yannickl88:
Q | A
-- | --
License | MIT

### DIFF
--- a/yannickl88/features-bundle/1.0/etc/packages/features.yaml
+++ b/yannickl88/features-bundle/1.0/etc/packages/features.yaml
@@ -1,0 +1,21 @@
+features:
+    tags:
+#        my.feature:
+#            request: ["beta", "on"]
+
+services:
+#    app.features.request_resolver:
+#        class: App\Feature\RequestResolver
+#        arguments:
+#            - "@request_stack"
+#        tags:
+#            # Registering the resolver "request" so it can be used in the config
+#            - { name: features.resolver, config-key: request }
+#
+#    app.some.service:
+#       class: App\Some\Service
+#       arguments:
+#           - "@features.tag"
+#       tags:
+#           # Using the defined tag
+#           - { name: features.tag, tag: my.feature }
--- a/yannickl88/features-bundle/1.0/manifest.json
+++ b/yannickl88/features-bundle/1.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Yannickl88\\FeaturesBundle\\FeaturesBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    }
+}


fabpot:
What about also adding a default config file as an example on how to use the bundle?

yannickl88:
I can't really add much other then commented stuff since the bundle does not have any revolvers.

sstok:
Service example filename is wrong, should be something like features (not your the Vendor name) 😉

yannickl88:
@sstok you are right, my bad

iltar:
@fabpot seems like https://pr3.contrib.symfony.sh/ is not working anymore: 
image

If I try: https://pr1.contrib.symfony.sh/ it "works", pr2 is down as well.

image

stof:
@iltar the URL used by the server changed. It is now https://symfony.sh/r/contrib/3

As the PR has not been updated after the backend change, the URL reported to github is still the old one

iltar:
Ah that works for me :+1:

yannickl88/features-bundle/1.0/etc/packages/features.yaml
```diff
@@ -0,0 +1,21 @@
+features:
+    tags:
+#        my.feature:
+#            request: ["beta", "on"]
+
+services:
```
stof: @fabpot should we really have the service definition and the semantic configuration in the same file ? Shouldn't services be defined in a separate file ?

Pierstoval: In this case, the goal is to make sure services that depend on this package are removed when removing the package, else you would have to remove them manually.

It seems fine to me :/

iltar: The classes it refers to, do not exist, they are merely examples. I do agree with keeping them in here, I'm just not entirely sure if the examples should be in here.

e-moe: do we really need to have this file? see similar comment from @stof in sibling PR: https://github.com/symfony/recipes-contrib/pull/6#discussion_r113647743

yannickl88:
Either way, this is just example config (and service definition) since the bundle does not provide any implementations. I added this on request of @fabpot (see first comment)

Anyhow, the bundle does not throw any errors when nothing is configured so everything is optional, hence why I commented everything.

Let me know if I need to remove the config file, change it or keep it as is.

fabpot:
I think having commented configuration to help people get started seems like a good idea to me.

yannickl88:
so... what can this be merged in it's current state or do I need to change anything?

yannickl88/features-bundle/1.0/etc/packages/features.yaml
fabpot: These services should be removed, as they should be added by the dev in the main `container.yaml` file.

yannickl88: `container.yaml` is something that is part of the app? or is it something I need to add?

Nyholm:
Thank you for this PR and sorry for the long wait. 
In order to get merged I need you to provide some good default config that will be good for 90% of the users.

yannickl88:
@Nyholm That is the thing, I cannot add a good default since this bundle only adds infrastructure. The functional code has to be made by the person using this.

It's like asking for some default config for form component, you cannot add those without implementing a form type.

So at best I can add some commented service definitions, but that is about it.

See the [readme for an example](https://github.com/yannickl88/features-bundle).

Nyholm:
I understand. Then I suggest closing this PR since Flex is smart enough to install this bundle without a recipe.

Nyholm:

PR #4, opened 2017-04-23
Add EkinoNewrelicBundle

GromNaN:
* Is it possible to add conditional configuration when an other bundle is installed? This bundle has interactions with [SonataBlockBundle](https://github.com/ekino/EkinoNewRelicBundle#integration-with-sonatablockbundle).
* Configuration could be made interactive with questions like "What transaction naming do you want to use?" or "Are you using Symfony HTTP Cache?"

### DIFF
--- a/ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
+++ b/ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
@@ -1,12 +1,12 @@
 ekino_new_relic:
     # enabled: true                         # Defaults to true
-    # application_name: My App              # default value in newrelic is "PHP Application", or whatever is set as php ini-value
+    # application_name: PHP Application     # default value in newrelic is "PHP Application", or whatever is set as php ini-value
     # deployment_names: ~                   # default value is 'application_name', supports string array or semi-colon separated string
-    # api_key: "%newrelic_api_key%"         # New Relic API
+    # api_key: "%env(NEWRELIC_API_KEY)%"    # New Relic API
     # license_key:                          # New Relic license key (optional, default value is read from php.ini)
     # xmit: false                           # if you want to record the metric data up to the point newrelic_set_appname is called, set this to true (default: false)
     # logging: false                        # If true, logs all New Relic interactions to the Symfony log (default: false)
-    # instrument: false                     # If true, uses enhanced New Relic RUM instrumentation (see below) (default: false)
+    # instrument: false                     # If true, uses enhanced New Relic RUM instrumentation (default: false)
     # log_exceptions: false                 # If true, sends exceptions to New Relic (default: false)
     # log_commands: true                    # If true, logs CLI commands to New Relic as Background jobs (default: true)
     # using_symfony_cache: false            # Symfony HTTP cache (default: false)
--- a/ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
+++ b/ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
@@ -1,17 +1,4 @@
 ekino_new_relic:
-    # enabled: true                         # Defaults to true
-    # application_name: PHP Application     # default value in newrelic is "PHP Application", or whatever is set as php ini-value
-    # deployment_names: ~                   # default value is 'application_name', supports string array or semi-colon separated string
-    # api_key: "%env(NEWRELIC_API_KEY)%"    # New Relic API
-    # license_key:                          # New Relic license key (optional, default value is read from php.ini)
-    # xmit: false                           # if you want to record the metric data up to the point newrelic_set_appname is called, set this to true (default: false)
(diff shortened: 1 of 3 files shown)


ekino/newrelic-bundle/1.3/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "Ekino\\Bundle\\NewRelicBundle\\NewRelicBundle": ["all"]
```
fabpot: Do we really want to enable it for all envs? Or just the `prod` one?

fabpot:
There is no difference between configuring a bundle with a recipe or by hand. So, the answer is no to the "conditional configuration". That should be done via a compiler pass.

I'm against interactive questions as the goal is to **auto**configure things. You can display some information about the next steps at the end of the install if needed. But that should not replace proper documentation.

ekino/newrelic-bundle/1.3/manifest.json
```diff
@@ -0,0 +1,8 @@
...
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+	},
```
fabpot: Not valid JSON

ekino/newrelic-bundle/1.3/manifest.json
GromNaN: Indeed, I'm changing it and moving the config file to prod directory.

ekino/newrelic-bundle/1.3/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "Ekino\\Bundle\\NewRelicBundle\\NewRelicBundle": ["prod"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+	}
```
fabpot: Looks like there is an indent issue

ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
```diff
@@ -0,0 +1,17 @@
+ekino_new_relic:
```
GromNaN: Naming the yaml file with the config root key could be considered as a best practice.

fabpot:
This one is quite interesting as this is one that could be in the "main" repository. /cc @symfony/deciders

ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
fabpot: Indeed, I will add a check for that in the validator.

ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
```diff
@@ -0,0 +1,17 @@
...
+    # enabled: true                         # Defaults to true
+    # application_name: PHP Application     # default value in newrelic is "PHP Application", or whatever is set as php ini-value
+    # deployment_names: ~                   # default value is 'application_name', supports string array or semi-colon separated string
+    # api_key: "%env(NEWRELIC_API_KEY)%"    # New Relic API
+    # license_key:                          # New Relic license key (optional, default value is read from php.ini)
+    # xmit: false                           # if you want to record the metric data up to the point newrelic_set_appname is called, set this to true (default: false)
+    # logging: false                        # If true, logs all New Relic interactions to the Symfony log (default: false)
+    # instrument: false                     # If true, uses enhanced New Relic RUM instrumentation (see below) (default: false)
```
javiereguiluz: The *"(see below)"* mention in this line, what does it refer to?

javiereguiluz:
In my opinion, before considering to include this bundle in the main repo, the documentation should be completely updated. It only talks about "Symfony2" and it evens explains how to install the bundle using the "deps" file of Symfony 2.0.x.

ekino/newrelic-bundle/1.3/etc/packages/prod/ekino_new_relic.yaml
GromNaN: Copy/paste from the readme.

stof:
I don't think the file in the recipe should have all available options commented. It should suggest only the important ones (`api_key` for instance).
Advanced options are useless for most people and so should not be here. Th doc is the place where they should be explained

iltar:
What is exactly the workflow to get something into the main repository? symfony/recipes#29

Pierstoval:
> What is exactly the workflow to get something into the main repository? symfony/recipes#29

It will be decided by the core team (a.k.a @symfony/deciders) , I think.

GromNaN:
Good idea @stof. I've updated the PR.
I agree that this bundle need some cleanup.

fabpot:
👍

rande:
I have updated the readme file to remove the obsolete deps installation reference.

Thanks

PR #5, opened 2017-04-25
Add LexikJWTAuthenticationBundle

chalasr:
For JWT authentication.
See https://github.com/lexik/LexikJWTAuthenticationBundle

### DIFF
--- a/lexik/jwt-authentication-bundle/2.3/Makefile
+++ b/lexik/jwt-authentication-bundle/2.3/Makefile
@@ -1,0 +1,8 @@
+generate-jwt-keys:
+ifeq (, $(shell which openssl))
+$(error "Unable to generate keys (needs OpenSSL)")
+endif
+	mkdir -p etc/jwt
+	openssl genrsa -passout pass:${JWT_PASSPHRASE} -out ${JWT_PRIVATE_KEY_PATH} -aes256 4096
+	openssl rsa -passin pass:${JWT_PASSPHRASE} -pubout -in ${JWT_PRIVATE_KEY_PATH} -out ${JWT_PUBLIC_KEY_PATH}
+	@echo "\033[32mRSA key pair successfully generated\033[39m"
--- a/lexik/jwt-authentication-bundle/2.3/manifest.json
+++ b/lexik/jwt-authentication-bundle/2.3/manifest.json
@@ -1,0 +1,13 @@
+{
+    "bundles": {
+        "Lexik\\Bundle\\JWTAuthenticationBundle\\LexikJWTAuthenticationBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%"
+    },
+    "env": {
+        "JWT_PRIVATE_KEY_PATH": "%ETC_DIR%/jwt/private.pem",
+        "JWT_PUBLIC_KEY_PATH": "%ETC_DIR%/jwt/public.pem",
+        "JWT_PASSPHRASE": "%generate(secret)%"
+    }
+}
--- a/lexik/jwt-authentication-bundle/2.3/post-install.txt
+++ b/lexik/jwt-authentication-bundle/2.3/post-install.txt
@@ -1,0 +1,1 @@
+<bg=blue;fg=white>Now</>, run <comment>make generate-jwt-keys</> to generate a key pair for LexikJWTAuthenticationBundle.
--- a/lexik/jwt-authentication-bundle/2.3/etc/packages/lexik_jwt_authentication.yaml
+++ b/lexik/jwt-authentication-bundle/2.3/etc/packages/lexik_jwt_authentication.yaml
@@ -1,0 +1,4 @@
+lexik_jwt_authentication:
+    private_key_path: "%env(JWT_PRIVATE_KEY_PATH)%"
+    public_key_path:  "%env(JWT_PUBLIC_KEY_PATH)%"
+    pass_phrase:      "%env(JWT_PASSPHRASE)%"


javiereguiluz:
I'm 👍  for adding this bundle. However, what if we leverage more Symfony Flex features? Quick thoughts:

1. Define a minimal default config file with these two config options only:

```
lexik_jwt_authentication:
    private_key_path: '%jwt_private_key_path%'
    public_key_path:  '%jwt_public_key_path%'
```

2. Define two env vars instead of the params: `jwt_private_key_path` and `jwt_public_key_path`

The rest of the config would be hard to automatize, because it depends on the current contents of `security.yml`.

chalasr:
@javiereguiluz config added, thanks!

Pierstoval:
> 2. Define two env vars instead of the params: `jwt_private_key_path` and `jwt_public_key_path`

Best practice IMO

lexik/jwt-authentication-bundle/2.3/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Lexik\\Bundle\\JWTAuthenticationBundle\\LexikJWTAuthenticationBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%"
+    },
+    "env": {
+        "JWT_PRIVATE_KEY_PATH": "%kernel.project_dir%/var/jwt/private.pem",
```
javiereguiluz: I'm not sure if this will work ... but there's no `%VAR_DIR%` value (only ETC, CONF, BIN, SRC and WEB). I'm going to ask about this in the Flex repository.

chalasr: Ah indeed the `var` dir doesn't exist anymore. Maybe do you know where the cache would go using Flex?

javiereguiluz: I explained myself wrong. The var/ dir still exists in Flex apps ... but you can't use `%VAR_DIR%` in your recipes because that variable is not defined. But if we define variables for everything (bin/, etc/, src/, web/), why shouldn't we provide it for var/ too? I've proposed to add this in the Flex repo (which is still private, that's why I don't link it here).

chalasr: Ok understood, I think this should work anyway (inspired from https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/etc/packages/doctrine.yaml#L12)

javiereguiluz: The difference is that you linked to a YAML file (there it will keep working) but here you are using a Symfony param inside a env var. I don't know if that would work.

chalasr:
Given this bundle is a reference regarding the need it solves, could it be subject to the main recipes repository?

lexik/jwt-authentication-bundle/2.3/manifest.json
sstok: IMO a Key is configuration and should be kept in etc directory. Var should only be kept for files that can be easily purged without breaking the application (cache, storage, logs).

sstok:
Do you think it would a good idea to provide a Makefile job that will generate these keys?

lexik/jwt-authentication-bundle/2.3/manifest.json
chalasr: I wasn't comfortable with putting this in  the`app/config` directory of the SE as this is not something editable, but I agree and it feels right to me now that this dir is called `etc`.
Updated to `%ETC_DIR%/jwt/*.pem`.

chalasr:
@sstok The commands to run (2) both need an interaction for generating the passphrase, a bit annoying. The good news is that we can define the env var as a `JWT_PASSPHRASE=ChangeMe`, set it using a `%env(JWT_PASSPHRASE)%` in the bundle config and consume the env var from the target, passing the passphrase as an option to these commands. Even more automated! 👍  Will do tomorrow

chalasr:
Make task added

chalasr:
See https://github.com/symfony/recipes/pull/32 for submission to the main recipes repository.

lexik/jwt-authentication-bundle/2.3/Makefile
```diff
@@ -0,0 +1,5 @@
+generate-rsa-keys:
+	mkdir -p etc/jwt
+	openssl genrsa -passout pass:${JWT_PASSPHRASE} -out etc/jwt/private.pem -aes256 4096
```
javiereguiluz: To provide a good user experience, we could use here the same strategy as Symfony (https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/Makefile) to display a good error message when the `openssl` command is not installed in the local computer.

lexik/jwt-authentication-bundle/2.3/Makefile
```diff
@@ -0,0 +1,5 @@
+generate-rsa-keys:
```
javiereguiluz: `generate-jwt-keys` could be a better name? `RSA` looks a bit generic, but `JWT` should be clear enough.

chalasr: Renamed

lexik/jwt-authentication-bundle/2.3/Makefile
chalasr: Error message added

fabpot:
What about adding a small output text to hint people to run `make generate-jwt-keys`?

chalasr:
post-install output added, I'll update the bundle docs to skip some manual configuration steps when installed via Flex.

lexik/jwt-authentication-bundle/2.3/post-install.txt
```diff
@@ -0,0 +1,7 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
```
fabpot: I would use a on-liner here instead of a block (to avoid having several such blocks when installing several packages with post-install output). What about something along the lines of:

```
<bg=blue;fg=white> Next  </>Generate a key pair for LexikJWTAuthenticationBundle
```

lexik/jwt-authentication-bundle/2.3/post-install.txt
```diff
@@ -0,0 +1,7 @@
...
+
+  * <fg=blue>Execute</> the <comment>make generate-jwt-keys</> command to generate your key pair
+
+  * <fg=blue>Read</> the documentation at <comment>https://github.com/lexik/LexikJWTAuthenticationBundle#documentation</>
```
fabpot: I would remove this as we won't list doc URLs for each bundle you install.

lexik/jwt-authentication-bundle/2.3/post-install.txt
javiereguiluz: We could even make the entire `post-install.txt` file a one-liner:

```
Now, run <comment>make generate-jwt-keys</> to generate the JWT keys for LexikJWTAuthenticationBundle.
```

fabpot: Indeed, probably even better

fabpot: We would still need some color to catch the eyes of the dev

chalasr: updated

fabpot:
Using `%ETC_DIR%` in env var values is not supported. Added it in fabpot/flex@350ba9f

fabpot:
Closing in favor of symfony/recipes#32

PR #6, opened 2017-04-27
e-moe/guzzle6-bundle

e-moe:
Added: https://packagist.org/packages/e-moe/guzzle6-bundle

### DIFF
--- a/e-moe/guzzle6-bundle/1.1/manifest.json
+++ b/e-moe/guzzle6-bundle/1.1/manifest.json
@@ -1,0 +1,5 @@
+{
+  "bundles": {
+    "Emoe\\GuzzleBundle\\EmoeGuzzleBundle": ["all"]
+  }
+}


e-moe/guzzle6-bundle/1.1/etc/packages/emoe_guzzle_bundle.yaml
```diff
@@ -0,0 +1,4 @@
+emoe_guzzle:
+    log:
+        enabled: true # Logging requests to Monolog
+        format: 'Guzzle: [{ts}] "{method} {uri} HTTP/{version}" {code}' # Optional log format customization
```
stof: does the bundle actually require to provide a log format ? If it has a format making sense, it should be the default one

e-moe: yes, it have default values and these settings are optional in my bundle. should I remove them from this file?

stof: well, if the default value is fine, yes. Why make the project repeat the default configuration ? It does not make sense

e-moe: hint for developers: how this option can be used. I will comment it out so default value will be used and developers will be able to change it

javiereguiluz: I'm divided about this 🤔   I think both of you are right: 1) this is the default config, so why repeat it?; 2) it's nice to tell users how to customize this without forcing them to look for in the docs.

stof: @javiereguiluz but then, it forces all developers joining a project to wonder "why is this config commented in our config file ?".
So I'm more in favor of removing the config entirely. People wanting to customize the log format would then read the doc (to know that it is possible, but also to know the available placeholder, which still requires reading the doc anyway)

e-moe: @stof , done

fabpot:
I'm closing this PR as Symfony Flex is able to automatically register a bundle for all environments if there is no associated recipe.

PR #9, opened 2017-05-02
Add pomm bundle recipe

sanpii:
https://github.com/pomm-project/pomm-bundle

### DIFF
--- a/pomm-project/pomm-bundle/2.3/etc/packages/pomm.yaml
+++ b/pomm-project/pomm-bundle/2.3/etc/packages/pomm.yaml
@@ -1,0 +1,4 @@
+pomm:
+    configuration:
+        db:
+            dsn: "%env(DATABASE_URL)%"
--- a/pomm-project/pomm-bundle/2.3/etc/routing/dev/pomm.yaml
+++ b/pomm-project/pomm-bundle/2.3/etc/routing/dev/pomm.yaml
@@ -1,0 +1,3 @@
+_pomm:
+    resource: "@PommBundle/Resources/config/routing.yml"
+    prefix:   /_pomm
--- a/pomm-project/pomm-bundle/2.3/manifest.json
+++ b/pomm-project/pomm-bundle/2.3/manifest.json
@@ -1,0 +1,11 @@
+{
+    "bundles": {
+        "PommProject\\PommBundle\\PommBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    },
+    "env": {
+        "DATABASE_URL": "pgsql://user:pass@127.0.0.1:5432/my_db_name"
+    }
+}


pomm-project/pomm-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "alias": "pomm",
+    "bundles": {
+        "PommProject\\PommBundle\\PommBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    },
+    "env": {
+        "DATABASE_DSN": "pgsql://user:pass@127.0.0.1:5432/my_db_name"
```
javiereguiluz: In Doctrine's recipe we use the `DATABASE_URL` env var. See https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/manifest.json  Do you need a different env var from Doctrine or can you reuse it?

pomm-project/pomm-bundle/2.0/etc/pomm.yaml
```diff
@@ -0,0 +1,10 @@
+pomm:
+    configuration:
+        my_db1:
+            dsn: '%env(DATABASE_DSN)%'
+#            pomm:default: true
```
javiereguiluz: In other PRs, we were discussing about the utility of adding commented config. @stof said that it's better to not add this commented config. Others think it's OK. I'm divided. Maybe @fabpot can share his opinion about this.

chanmix51: It is possible to uncomment it since the first database session declared is the default one anyway.

sanpii: I saw, I don’t have a strong opinion about that. It’s maybe easier to maintain without comment (each new configuration require a new pull request, errors in comment are silently).

pomm-project/pomm-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "alias": "pomm",
```
Pierstoval: I think aliases are not allowed in the contrib repository.

@javiereguiluz / @fabpot ?

javiereguiluz: I don't know if it's not allowed (the error page doesn't show it: https://symfony.sh/r/contrib/9) but it's sure that it will be ignored until the package is moved to symfony/recipes.

Pierstoval: IMO the flex validation server should check this and forbid an alias on contrib repos, because aliases will be decided on the symfony/recipes only

sanpii: @Pierstoval the validation boot doesn’t failed, but the alias seems doesn’t work:
```
$ SYMFONY_ENDPOINT=https://symfony.sh/r/contrib/9 composer req pomm                                                         
                                                                                                                                                       
                                                                                                                                                       
  [UnexpectedValueException]    
  "pomm" is not a valid alias.  
                           
```

fabpot: It's not "alias", but "aliases" as an array.

fabpot: Sorry, forgot to tell you that aliases are indeed not supported in the contrib repo (validation fails).

sanpii: Ok, the bot fails fine with the right key:
2017-05-02-152517_1676x993_scrot

pomm-project/pomm-bundle/2.3/etc/packages/pomm.yaml
```diff
@@ -0,0 +1,4 @@
+pomm:
+    configuration:
+        my_db1:
```
fabpot: I would give it a better name, like `db`

pomm-project/pomm-bundle/2.3/etc/packages/pomm.yaml
```diff
@@ -0,0 +1,4 @@
...
+            dsn: '%env(DATABASE_URL)%'
```
fabpot: I think we are using `"` instead of `'`

pomm-project/pomm-bundle/2.0/manifest.json
Pierstoval: @fabpot Isn't there a way to make sure there is no wrong key in the recipe file? Here, we have a hidden issue: "alias" is not available but the bot doesn't care. Maybe Symfony/Config can be used to validate any recipe file and make sure there is no trailing or unintended data in the file :)

fabpot: I've added this feature on my todo-list, I thought I implemented that already but that's not the case apparently (keep in mind that everything is in alpha mode for now). I won't be able to use symfony/config as the server is not written in PHP :)

Pierstoval: Anyway it's good news that it's on the todo list 👍

pomm-project/pomm-bundle/2.3/etc/packages/pomm.yaml
sanpii: Fixed.

pomm-project/pomm-bundle/2.3/etc/packages/pomm.yaml
sanpii: Fixed too, thank you.

fabpot:
Thank you @sanpii.

PR #10, opened 2017-05-02
Add DoctrineMongoDBBundle recipe

Lctrs:


doctrine/mongodb-odm-bundle/3.3/etc/packages/doctrine_mongodb.yaml
```diff
@@ -0,0 +1,18 @@
+doctrine_mongodb:
+    auto_generate_proxy_classes: "%kernel.debug%"
+    auto_generate_hydrator_classes: "%kernel.debug%"
```
stof: I suggest using single quotes. Double quotes in YAML require escaping lots of things.

Lctrs: In the officiel recipe for `doctrine/doctrine-bundle`, we use double quotes for this parameter.

See https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/etc/packages/doctrine.yaml#L5.

doctrine/mongodb-odm-bundle/3.3/etc/packages/doctrine_mongodb.yaml
```diff
@@ -0,0 +1,18 @@
...
+    default_database: "%env(MONGODB_DBNAME)%"
+    document_managers:
+        default:
+            auto_mapping: true
+            mappings:
+                App:
+                    is_bundle: false
+                    type: annotation
+                    dir: "%kernel.project_dir%/src/Document"
+                    prefix: "App\\Document\\"
```
Pierstoval: We often voluntarily remove quotes for class names and would just use `prefix: App\Document\` in this case, it should work

doctrine/mongodb-odm-bundle/3.3/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "Doctrine\\Bundle\\MongoDBBundle\\DoctrineMongoDBBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/",
+        "src/": "%SRC_DIR%/"
+    },
+    "env": {
+        "MONGODB_URL": "mongodb://localhost:27017",
```
Pierstoval: Can you suffix the database name to keep consistency with Doctrine ORM recipe and then simply have
`"MONGODB_URL": "mongodb://localhost:27017/symfony",` ?

Lctrs:
One thing I don't know how to translate is the 2 lines that we needed to add to the old `app/autoload.php` (see the [docs](https://symfony.com/doc/current/bundles/DoctrineMongoDBBundle/index.html#register-the-annotations-and-the-bundle) for the lines in question). Any idea ?

doctrine/mongodb-odm-bundle/3.3/etc/packages/doctrine_mongodb.yaml
Lctrs: That's not inline with the official Doctrine ORM recipe.

Pierstoval:
The ORM annotation driver was lazy-loaded in symfony/symfony#21837, maybe something similar could be done for the ODM, I don't know 😕

stof:
Actually, as the ODM bundle uses the annotation_reader defined by Symfony, it automatically benefits from the change done in Symfony.

Lctrs:
Yeah. I tested it in a fresh application and it works as expected.

Lctrs:
Should be good now.

fabpot:
Thank you @Lctrs.

PR #11, opened 2017-05-03
Add Behat Symfony2Extension recipe

QcRafal:
Due to directories structure changes Behat Symfony2Extension needs some configuration tweaks


behat/symfony2-extension/2.1/Makefile
```diff
@@ -0,0 +1,5 @@
+behat-init:
+	@test -f bin/behat && bin/behat --init || echo "cannot init behat"
+
+behat:
+	@test -f bin/behat && bin/behat || echo "cannot run behat"
```
fabpot: What's the benefit of defining these in the Makefile instead of just using `bin/behat` directly?

QcRafal: Basically none. Just alias.

I was rather playing around.

behat/symfony2-extension/2.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "bin/": "%BIN_DIR%",
+        "behat.yaml.dist": "behat.yaml.dist"
```
QcRafal: Can this configuration be based on `.yml` files not `.yaml` (it does not pass validation)?  For Behat [`v3.3.0`](https://github.com/Behat/Behat/blob/v3.3.0/src/Behat/Behat/ApplicationFactory.php#L116) it checks only `.yml` files. As I can see it's improved on [`maser`](https://github.com/Behat/Behat/blob/master/src/Behat/Behat/ApplicationFactory.php#L116)

behat/symfony2-extension/2.1/Makefile
QcRafal: I found that it can be useful now in this way: `@test -f bin/behat && bin/behat --config behat.yaml` because of [this](https://github.com/symfony/recipes-contrib/pull/11/files#r114464402)

behat/symfony2-extension/2.1/manifest.json
fabpot: If that's a limitation of Behat, keep `.yml`

behat/symfony2-extension/2.1/Makefile
```diff
@@ -0,0 +1,5 @@
+behat-init:
```
stof: I don't think this deserves an alias in the Makefile, as you will only run it once in the life of the project (or never, if you write the config files by hand)

behat/symfony2-extension/2.1/behat.yaml.dist
```diff
@@ -0,0 +1,7 @@
+default:
+    extensions:
+        Behat\Symfony2Extension:
+            kernel:
+                bootstrap: vendor/autoload.php
```
stof: useless. The composer autoloader is already loaded automatically by Behat. Adding a bootstrap file is useful when a *custom* file needs to be loaded

behat/symfony2-extension/2.1/behat.yaml.dist
```diff
@@ -0,0 +1,7 @@
...
+                path: src/Kernel.php
```
stof: this is useless as well. The Kernel class is autoloadable since a long time (we changed it in symfony-standard 2.8) so there is no need to make the Behat extension require the kernel file directly. The autoloader will handle it.

behat/symfony2-extension/2.1/bin/behat
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env php
+<?php
+
+/*
+ * This file is part of the Behat.
```
stof: Replacing the Behat test runner by your own script is a bad idea, as it means we will then have to keep it in sync with the one in Behat (which is not even the package for which we have a recipe).

so this is a big -1 from me.

If the goal is to load the .env file, a better way is to create a DotenvExtension for Behat instead, for cases where you want to use a .env file.

behat/symfony2-extension/2.1/Makefile
stof: We already merged support for using behat.yaml natively in Behat (and with the automatic fallback to behat.yaml.dist, which is not supported by your alias, which *requires* overwriting the config locally). This is a matter of release Behat 3.4.0.

behat/symfony2-extension/2.1/Makefile
fabpot: or just `source .env`

behat/symfony2-extension/2.1/behat.yaml.dist
QcRafal: I agree that behat loads it, but extension tries to load file from default configuration (`app/autoload.php`) and throws ProcessingException (`Could not load bootstrap file. Please check your configuration at "kernel.bootstrap"`)

I know the purpouse of this option and I prefer to not do such workarounds but if the recipe is to pre-configure this is needed.

behat/symfony2-extension/2.1/behat.yaml.dist
QcRafal: You're right.

behat/symfony2-extension/2.1/behat.yaml.dist
stof: then set it to null, to load nothing

stof: another option: make it point to a file containing *only* the Dotenv loading 😄 (but a reusable extension for Behat would be cleaner)

fabpot: instead of creating a file with only dotenv loading, why not prefix the command with `source .env`?

stof: well, isn't the Makefile sourcing it already now btw ?

QcRafal: I tried this way: `@test -f vendor/bin/behat && vendor/bin/behat || echo "cannot run behat (needs behat/behat)"` with `kernel.bootstrap` set to `null`and all I'm getting is `Environment variable not found: "DATABASE_URL".`

it worked in this way: `@test -f vendor/bin/behat && DATABASE_URL=$(DATABASE_URL) vendor/bin/behat -v || echo "cannot run behat (needs behat/behat)"` and i see that Symfony Kernel is initialized with variables from behat configuration so APP_ENV and APP_DEBUG

I'll try to test it more complex

stof: well, the Symfony2Extension does not yet use APP_ENV and APP_DEBUG to configure the kernel.

QcRafal:
@stof @fabpot It's not the cleanest, but it works properly without errors.
I think that, for now boostrap file with `.env` loader is the simpliest solution (and the best against my knowledge)

I tried to make it with makefile only, but I had to give up (I have to learn more about Makefile and unix staff)

I was also thinking more about what will happend when You try to install more extenstions like this, for example `MinkExtension` which, as it is describied in docs, also needs some configuration in behat.yml file. But i think this is topic for future considerations

behat/symfony2-extension/2.1/post-install.txt
```diff
@@ -0,0 +1,11 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * <fg=blue>Init Behat</>:
+    1. Change to the project directory
+    2. Execute the <comment>vendor/bin/behat --init</> command;
```
stof: I don't think we need to document how to use behat when they installed an extension for behat (thus, the way to start using Behat may change, and what people should do to use behat is described in the Behat doc.
so I suggest removing this message.

behat/symfony2-extension/2.1/Makefile
```diff
@@ -0,0 +1,3 @@
+behat:
+	@test -f vendor/bin/behat && vendor/bin/behat || echo "cannot run behat (needs behat/behat)"
```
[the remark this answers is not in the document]
fabpot: This file must be removed now as we don't support Makefiles anymore.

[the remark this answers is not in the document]
fabpot: This file must be removed now as we don't support Makefiles anymore.

Nyholm:
@QcRafal Could you please make some updates to this recipe so it is compatible with sf4 and the latest changes in flex?

Nyholm:

PR #12, opened 2017-05-04
Added OrbitaleCmsBundle

Pierstoval:
| Q             | A
| ------------- | ---
| License       | MIT

=> https://github.com/Orbitale/CmsBundle

Lightweight CMS with already built-in Page and Category entities.

I'd like to propose a ready-to-use EasyAdmin integration, but I think I might need to create a "pack" as proposed by @fabpot in his post.

What do you think?

### DIFF
--- a/orbitale/cms-bundle/2.0/manifest.json
+++ b/orbitale/cms-bundle/2.0/manifest.json
@@ -1,0 +1,9 @@
+{
+    "bundles": {
+        "Orbitale\\Bundle\\CmsBundle\\OrbitaleCmsBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/",
+        "src/": "%SRC_DIR%/"
+    }
+}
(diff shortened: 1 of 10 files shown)


orbitale/cms-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "bundles": {
+        "Orbitale\\Bundle\\CmsBundle\\OrbitaleCmsBundle": [
+            "all"
+        ]
```
fabpot: should be on one line

Pierstoval: Fixed.
I'm gonna release a 3.0 of this bundle for Symfony3 only + php7 requirements, with some BC breaks (routing config, different dependencies), please not merge until I also have put 3.0 manifest 🙂

Pierstoval:
Done, updated 2.0 and 3.0 versions of the bundle, and added 2.0 and 3.0 recipes for these specific versions (as 3.0 is SF3 & PHP7 only, I had to create a new major version, and as there are some adjustments in routing I needed two manifests, I guess there is no other way around even if there is some duplication)

Pierstoval:
Needs review 🙂

Pierstoval:
Thanks @javiereguiluz ❤️

PR #14, opened 2017-05-06
Add Imagine Bundle Recipe

robfrawley:
This pull request adds a Symfony Flex recipe for [`LiipImagineBundle`](https://github.com/liip/LiipImagineBundle), a popular image manipulation bundle that leverages the [Imagine Library](https://github.com/avalanche123/Imagine) internally.

I am waiting for some reviews from my fellow bundle maintainers, so do not merge this immediately. Also, as there isn't much information or many examples on writing these recipes yet, I look forward to fielding reviews or change requests from this repository's maintainers, as well.

Hopefully, this bundle will land in the exciting Symfony Flex ecosystem shortly!

### DIFF
--- a/liip/imagine-bundle/1.8/etc/packages/imagine.yaml
+++ b/liip/imagine-bundle/1.8/etc/packages/imagine.yaml
@@ -1,0 +1,36 @@
+#liip_imagine:
+#    
+#    # valid drivers options include "gd" or "gmagick" or "imagick"
+#    driver: "gd"
+#
+#    # define your filter sets under this option
+#    filter_sets:
+#
+#        # an example thumbnail transformation definition
+#        # https://symfony.com/doc/current/bundles/LiipImagineBundle/basic-usage.html#create-thumbnails
+#        squared_thumbnail:
+#
+#            # set your image quality defaults
+#            jpeg_quality:          85
+#            png_compression_level: 8
+#            
+#            # setup the filter steps to apply for this transformation
+#            filters:
+#
+#                # auto rotate the image using EXIF metadata
+#                auto_rotate: ~
+#
+#                # strip the image of all metadata
+#                strip: ~
+#
+#                # scale and square the image to the given dimensions
+#                thumbnail:
+#                    size:          [253, 253]
+#                    mode:          outbound
+#                    allow_upscale: true
+#
+#                # create border by placing image on larger black background
+#                background:
+#                    size:     [256, 256]
+#                    position: center
+#                    color:    '#fff'
--- a/liip/imagine-bundle/1.8/manifest.json
+++ b/liip/imagine-bundle/1.8/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Liip\\ImagineBundle\\LiipImagineBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    }
+}
--- a/liip/imagine-bundle/1.8/post-install.txt
+++ b/liip/imagine-bundle/1.8/post-install.txt
@@ -1,0 +1,6 @@
+  <bg=blue;fg=white>         Getting started using </><bg=blue;fg=white;options=bold>liip/imagine-bundle</><bg=blue;fg=white>          </>
+
+  <fg=blue;options=bold>Configure</> <fg=blue>your transformations:</>
+    1. You <options=bold>MUST</> verify and uncomment the configuration in <comment>%ETC_DIR%/packages/imagine.yaml</>.
+    2. You <options=bold>MAY</> configure your image transformation library (<comment>gmagick</>, <comment>imagick</>, or <comment>gd</> [default]).
+    3. You <options=bold>MAY</> define custom transformation definitions under the <comment>filter_sets</> option.
--- a/liip/imagine-bundle/1.8/etc/routing/imagine.yaml
+++ b/liip/imagine-bundle/1.8/etc/routing/imagine.yaml
@@ -1,0 +1,2 @@
+_liip_imagine:
+    resource: "@LiipImagineBundle/Resources/config/routing.yaml"


robfrawley:
I'm a little confused by the automated pull request checker, which has determined that the `liip/imagine-bundle` package does not exist with any available versions, despite its existence with a plethora of versions: what am I getting wrong here?

> The following errors occurred while validating the pull request: package "liip/imagine-bundle" does not have any available version on Packagist

Also, I was attempting to test the recipe by setting the `SYMFONY_ENDPOINT` environment variable and requiring Imagine Bundle, but the routing, configuration, and post-install message are not properly handled. Is this the expected behavior of the environment variable or am I using it incorrectly?

```
SYMFONY_ENDPOINT=https://github.com/robfrawley/recipes-contrib/ composer req liip/imagine-bundle:^1.7
```

antoligy:
Just to check, is it possible that this is because of our requirement on ext-gd?

fabpot:
There is a bug in the validator (the issue here is that there is no `dev-master` version on Packagist). I will fix soon.

antoligy:
Ah, thanks @fabpot!

liip/imagine-bundle/1.7/post-install.txt
```diff
@@ -0,0 +1,14 @@
...
+  * <fg=blue>Configure</> your transformations:
+    - Verify and uncomment the config options in <comment>imagine.yaml</comment>
+    - You MUST specify the image manipulation library, one of <comment>gd</comment>, <comment>gmagick</comment>, or <comment>imagick</comment>
+    - You MAY specify additional transformation definitions under the <comment>filter_sets</comment> option
+
+  * <fg=blue>Use</> your transformations:
+    - Twig: <comment>{{ imagine_filter('my_thumbnails') }}</comment>
+    - PHP:  <comment>$this->get('liip_imagine.cache.manager')->getBrowserPath('/relative/path/to/image.jpg', 'my_thumbnails');</comment>
+
+  * <fg=blue>Read</> the documentation at <comment>http://symfony.com/doc/current/bundles/LiipImagineBundle</>
```
javiereguiluz: Please use `https://` when linking to symfony.com. Thanks!

fabpot: I will add a check for that :)

robfrawley: Fixed; thanks!

liip/imagine-bundle/1.7/etc/routing/imagine.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+liip_imagine_filter_runtime: 
+  path: "/media/cache/resolve/{filter}/rc/{hash}/{path}"
+  defaults: 
+    _controller: "%liip_imagine.controller.filter_runtime_action%"
+  methods: 
+    - GET
+  requirements: 
+    filter: "[A-z0-9_\\-]*"
+    path: .+
```
Pierstoval: Maybe use 4 spaces instead of 2 to keep consistency with the rest of the codebase?

liip/imagine-bundle/1.7/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "Liip\\ImagineBundle\\LiipImagineBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    }
+}
```
Pierstoval: Missing newline here

liip/imagine-bundle/1.7/etc/routing/imagine.yaml
robfrawley: Fixed

liip/imagine-bundle/1.7/manifest.json
robfrawley: Fixed; thanks for the review!

Pierstoval: 👍

liip/imagine-bundle/1.7/etc/routing/imagine.yaml
```diff
@@ -0,0 +1,19 @@
...
+        path: .+
+
+liip_imagine_filter_runtime: 
+    path: "/media/cache/resolve/{filter}/rc/{hash}/{path}"
+    defaults: 
+        _controller: "%liip_imagine.controller.filter_runtime_action%"
+    methods: 
+        - GET
+    requirements: 
+        filter: "[A-z0-9_\\-]*"
```
Pierstoval: Maybe use single quotes everywhere too? 😆 It'll help in backslash escaping too

robfrawley: Good idea; updated.

liip/imagine-bundle/1.7/etc/routing/imagine.yaml
```diff
@@ -0,0 +1,19 @@
+liip_imagine_filter:
```
Koc: is it possible inclide `@LiipImagineBundle/Resources/config/routing.xml` instead of copy-pasting it content here?

robfrawley: Assuming adding the following addition to the `manifest.json` results in the same auto-registration of the file, I agree, aside from my comment below:

```yaml
"copy-from-package": {
    "Resources/config/routing.xml": "%ETC_DIR%/routing/"
}
```

My only concern is that the use of YAML here was an intentional choice made to lower the entry barrier for someone looking to, perhaps, edit the route paths; XML is generally more difficult for newcomers to follow or edit.

Pierstoval: I agree that you could just use 

```yaml
I agree that you could just use 

```yaml
liip_imagine:
    resource: '@LiipImagineBundle/Resources/config/routing.xml'
```

Is fairly enough to add routing config with flex

robfrawley: Ooh; good idea.

Koc: so convert LiipImagineBundle routing to yaml and copy it?

robfrawley: That isn't something we can do until the next major release, as people rely on the current name for their own usage in pre-flex applications.

robfrawley:
@fabpot Is there a standard for enabling multiple package versions and avoiding duplication of files; every single recipe file for this bundle is exactly the same for both version `1.7` and `2.0`.

For example, is the solution found at the following link acceptable, where I created a `.default` directory with all the recipe files and then symlinked `./1.7` and `./2.0` to `./.default/`?

[github.com/robfrawley/recipes-contrib/tree/feature-include-version-1.x-and-2.x/liip/imagine-bundle](https://github.com/robfrawley/recipes-contrib/tree/feature-include-version-1.x-and-2.x/liip/imagine-bundle)

Is this a suitable strategy, or is there an alternative standard, or should the files simply be duplicated in each version folder?

Pierstoval:
@robfrawley Take example of the PHPUnit recipe: https://github.com/symfony/recipes/blob/master/phpunit/phpunit/4.7/manifest.json 

You can specify other versions that are compatible with this recipe.

But if files change, you'll have to create a new directory

robfrawley:
@Pierstoval Oh, wonderful; I don't know how I missed that. Thanks!

robfrawley:
Please do not merge until liip/LiipImagineBundle#925 is merged and version `1.8.0` is released; will update once we're ready to go on our end.

fabpot:
Bug fix, validation does not pass now because the 1.8 version does not exist yet, which is expected.

robfrawley:
This should be all ready to go, but it doesn't seem to work properly when I test it

```bash
composer create-project symfony/skeleton:3.3.x-dev symfony-flex
cd symfony-flex
SYMFONY_ENDPOINT=https://symfony.sh/r/contrib/14 composer req "liip/imagine-bundle:^1.8"
ls etc/**
```
Results in the following output
```
etc/bundles.php  etc/container.yaml  etc/routing.yaml

etc/packages:
app.yaml  dev  framework.yaml  test
```

Where is the `packages/imagine.yaml` configuration or the `routing/imagine.yaml` routing file? Are these supposed to exist there or are they handled "magically" elsewhere? Thanks!

liip/imagine-bundle/1.8/etc/packages/imagine.yaml
```diff
@@ -0,0 +1,36 @@
+#liip_imagine:
+#    
+#    # valid drivers options include "gd" or "gmagick" or "imagick"
```
robfrawley: Thanks @symfony-bot - all set!

liip/imagine-bundle/1.8/post-install.txt
```diff
@@ -0,0 +1,14 @@
+<bg=blue;fg=white>                                 </>
+<bg=blue;fg=white> What's next for Imagine Bundle? </>
+<bg=blue;fg=white>                                 </>
```
robfrawley: Thanks @symfony-bot - all set!

javiereguiluz:
@robfrawley I can confirm the error that you reported. ImagineBundle is correctly added to `etc/bundles.php`, but nothing else works: the default config and routing files are not created and the `post-install.txt` message is not displayed.

Koc:
Also this bundle should register form theme. Is it possible?

Pierstoval:
> Also this bundle should register form theme. Is it possible?

In the config file you're copying to `etc/`, you can add a `twig:` instruction and register the form themes, I think they are merged, but not sure, should be tested 🤔 

But the best option may be to do this in your bundle and register the forms from a Compiler Pass, if possible

liip/imagine-bundle/1.8/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Liip\\ImagineBundle\\LiipImagineBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    },
+    "copy-from-package": {
+    	"Resources/config/routing.yaml": "%ETC_DIR%/routing/imagine.yaml"
```
stof: why not simply importing the route from the bundle itself ? This way, people won't alter the route definitions by mistake (making it a maintenance nightmare for you as you won't know what the actual content of the file is).

Btw @fabpot, is the file copied from the package updated when updating the package later ? I don't think so. So this would be another argument for importing properly.

liip/imagine-bundle/1.8/etc/packages/imagine.yaml
stof: @fabpot it would be great to put the comment on a line containing the error, so that it disappears when being fixed.

liip/imagine-bundle/1.8/manifest.json
robfrawley: My understanding was that this would work, and if it doesn't I believe it should. As for the choice to expose our full routes, this decision was made after a discussion by different contributors and maintainers; we are intentionally exposing the full routing configuration.

stof: I checked the code of the flex plugin, and the recipe is only taken into account for the *initial* installation of the package. So when updating from a version to the next one, the routing file will **not** be updated (which makes sense as it would overwrite user modifications otherwise, and files added in ETC_DIR are meant to be modified to tweak the config).
So IMO, this is the worse decision maintainers could have taken.

The few users wanting to redefine routes fully can copy/paste the file and use this to replace the import, instead of forcing everyone to copy/paste them on each update.

robfrawley: I'll bring this up and see what they want to do.

robfrawley: Hm, so changing this (per your suggestion) doesn't actually cause this recipe to work, though. Any idea why the recipe doesn't apply the configuration file or the routing file, using either method?

fabpot:
Works for me.

I think you forgot to run `composer config extra.symfony.allow-contrib true` in your project. In which case, the behavior is correct as recipes from contrib are only available when you set `allow-contrib` to `true` in your project's `composer.json` file.

I will update the deployment page to add a warning about that for contrib pull requests.

fabpot:
I've just updated the page -> https://symfony.sh/r/contrib/14, should be better hopefully.

liip/imagine-bundle/1.8/post-install.txt
```diff
@@ -0,0 +1,14 @@
...
+
+  * <fg=blue>Configure</> your transformations:
+    - Verify and uncomment the config options in <comment>imagine.yaml</>
```
fabpot: Could be `in %ETC_DIR%/packages/imagine.yaml`.

robfrawley:
@fabpot You were right; I had thought the composer configuration was global, not per project; thanks for the clarification on that. The new PR deployment page is great, BTW.

At this point, this looks good from our end. I was able to successfully test both version `1.8` and `2.0`. Let me know if you require any additional changes.

robfrawley:
Does anyone following this thread want to approve this (other than me) now that auto-merging is enabled?
 See symfony/recipes-contrib#21 for additional information.

fabpot:
@robfrawley I would prefer to be the one who will approve this PR, the rules are still changing a bit, so better to wait a bit :)

robfrawley:
@fabpot Apologies; I misunderstood the [updated `README.rst`](https://github.com/symfony/recipes-contrib/pull/21/files#diff-88b99bb28683bd5b7e3a204826ead112R24). No worries; take your time!

fabpot:
@robfrawley I think everything is ready now. So, for that PR to be merged, a Symfony Core team member or a member of the Liip organization must approve the PR (the member must be publicly listed on the Liip organization).

robfrawley:
@lsmith77 Care to approve this PR for our Symfony Flex recipe?

robfrawley:
Hm; thanks for giving that a try @Pierstoval :1st_place_medal:

liip/imagine-bundle/1.8/post-install.txt
```diff
@@ -0,0 +1,17 @@
...
+    2. You <options=bold>MAY</> configure your image transformation library (<comment>gmagick</>, <comment>imagick</>, or <comment>gd</> [default]).
+    3. You <options=bold>MAY</> define custom transformation definitions under the <comment>filter_sets</> option.
+
+  * <fg=blue;options=bold>Use</> <fg=blue>your transformations:</>
+    - Using Twig: <comment>{{ imagine_filter('squared_thumbnail') }}</>
+    - Using PHP : <comment>$this->get('liip_imagine.cache.manager')->getBrowserPath('/path/to/image.jpg', 'squared_thumbnail')</>
+
+  * <fg=blue;options=bold>Read</> <fg=blue>the documentation and</> <fg=blue;options=bold>ask</> <fg=blue>for help:</>
+    - Documentation: <comment>https://symfony.com/doc/current/bundles/LiipImagineBundle</>
+    - Issue Tracker: <comment>https://github.com/liip/LiipImagineBundle/issues</>
```
fabpot: This block is way too long. It should not replace proper documentation. It should only be used to indicate the immediate next steps to make it work, if any.

robfrawley: Brought it down from 18 to 10 lines: is that more appropriate?

liip/imagine-bundle/1.8/post-install.txt
```diff
@@ -0,0 +1,10 @@
+  <bg=blue;fg=white>         Getting started using </><bg=blue;fg=white;options=bold>liip/imagine-bundle</><bg=blue;fg=white>          </>
+
+  * <fg=blue;options=bold>Configure</> <fg=blue>your transformations:</>
+    1. You <options=bold>MUST</> verify and uncomment the configuration in <comment>%ETC_DIR%/packages/imagine.yaml</>.
+    2. You <options=bold>MAY</> configure your image transformation library (<comment>gmagick</>, <comment>imagick</>, or <comment>gd</> [default]).
+    3. You <options=bold>MAY</> define custom transformation definitions under the <comment>filter_sets</> option.
+
+  * <fg=blue;options=bold>Read</> <fg=blue>the documentation and</> <fg=blue;options=bold>ask</> <fg=blue>for help:</>
+    - Documentation: <comment>https://symfony.com/doc/current/bundles/LiipImagineBundle</>
+    - Issue Tracker: <comment>https://github.com/liip/LiipImagineBundle/issues</>
```
fabpot: The last paragraph should be removed.

robfrawley: Done!

PR #15, opened 2017-05-07
Docker support

dkarlovi:


dkarlovi:
This doesn't work, it says https://symfony.sh/r/contrib/15/m/dkarlovi/symfony-flex-docker/dev-master doesn't exist. /cc @javiereguiluz @fabpot

dkarlovi:
Problem above was: if doing `SYMFONY_ENDPOINT=XXX composer require vendor/package` as suggested in docs, it will require `dev-master` by default (because minimum stability in Composer), but the Flex server does not make it available. 

`SYMFONY_ENDPOINT=XXX composer require vendor/package:^<version>` works as expected.

fabpot:
@dkarlovi Great suggestion. I've just made the change.

dkarlovi/symfony-flex-docker/0.1/Makefile
```diff
@@ -0,0 +1,35 @@
+docker-start:
+	docker-compose -f ./docker-compose.yaml -f ./.infra/docker-compose/docker-compose.$(APP_ENV).yaml up -d
+.PHONY: docker-up
+
+docker-stop:
+	docker-compose down
+.PHONY: docker-down
+
+docker-start-dev:
```
jderusse: what's about using arguments instead of multiplicating calls

dkarlovi/symfony-flex-docker/0.1/Makefile
```diff
@@ -0,0 +1,35 @@
+docker-start:
+	docker-compose -f ./docker-compose.yaml -f ./.infra/docker-compose/docker-compose.$(APP_ENV).yaml up -d
+.PHONY: docker-up
+
+docker-stop:
```
jderusse: `stop` and `down` are 2 differents things

dkarlovi/symfony-flex-docker/0.1/manifest.json
```diff
@@ -0,0 +1,18 @@
+{
+    "copy-from-package": {
+        ".infra/": ".infra",
+        ".dockerignore": ".dockerignore",
+        "docker-compose.yaml": "docker-compose.yaml"
+    },
+    "env": {
+        "COMPOSE_PROJECT_NAME": "symfony",
```
dkarlovi: Agree, but this is a placeholder.

I'd like to get some information from `composer.json` (such as project name), hope Flex will provide access to it while copying files. /cc @fabpot

dkarlovi: This will also be important when I add image tags in `docker-compose.yaml`, need to know how to name them properly in order to tag images being built. Push and pull also need this.

dkarlovi/symfony-flex-docker/0.1/manifest.json
```diff
@@ -0,0 +1,18 @@
...
+        "CACHE_HOST": "cache",
+        "CACHE_PORT": "6379",
+        "DATABASE_HOST": "db",
+        "DATABASE_PORT": "3306",
+        "DATABASE_NAME": "symfony",
+        "DATABASE_USER": "symfony",
+        "DATABASE_PASSWORD": "%generate(secret)%",
+        "DATABASE_ROOT_PASSWORD": "%generate(secret)%"
```
dkarlovi: It's tougher to discover for the user, IMO. This way, you can setup your database GUI tool just by opening `.env`. If you want, you can also change the password and have a place to do it.

dkarlovi/symfony-flex-docker/0.1/Makefile
dkarlovi: I know, I've named targets `start` & `stop` because that's what I expect non-`docker-compose` users to look for first, but the difference between `up`/`start`, `down`/`stop` is not significant enough here.

Also, from my personal experience, I've never once used `start`/`stop`, only `up`/`down.

dkarlovi/symfony-flex-docker/0.1/Makefile
dkarlovi: You can already do 

     make docker-start APP_ENV=prod

I just feel 

    make docker-start-prod

is easier to type. But you might be right, the additional targets might not be significant enough (or used frequently enough) to have all these shortcuts in.

dkarlovi/symfony-flex-docker/0.1/Makefile
jderusse: people should know that runing docker-stop will delete networks and volumes too. This is not harmless.

dkarlovi: It doesn't delete volumes, it just stops and removes all containers (which is throw-away, recreated) and network (which is throw-away, recreated).

Basically, `docker-stop` will undo what `docker-start` did, except volumes and images themselves.

dkarlovi/symfony-flex-docker/0.1/Makefile
Pierstoval: And some shells have makefile autocomplete, so it's nice to have a dedicated command IMO

dkarlovi/symfony-flex-docker/0.1/Makefile
```diff
@@ -0,0 +1,51 @@
...
+docker-push:
+	docker-compose push
+.PHONY: docker-push
+
+docker-logs:
+	docker-compose logs -f
+.PHONY: docker-logs
+
+docker-init:
+	mkdir -p                            var/cache var/logs var/sessions
```
dkarlovi: Needs https://github.com/symfony/flex/issues/89

dkarlovi:
OK, so how does this work? Who needs to approve this?

sstok:
> OK, so how does this work? Who needs to approve this?

Anyone who isn't the author 😛 btw how would this if don't want to use MySQL (but PostgreSQL)?

dkarlovi:
@sstok 

>how would this if don't want to use MySQL (but PostgreSQL)?

I was already thinking about that, it's a bit tricky as the app container would need to have a PHP module for PostgreSQL instead of MySQL. I'd need to build a small utility which would know how to generate proper Dockerfiles and then split the repo to: base (with auto-generating Dockerfiles for app container), db-mysql, db-postgresql, cache-redis, cache-memcached, etc. It would be quite a bit of work.

It's easy to do manually for now, though:

1. change the [DB Dockerfile](https://github.com/dkarlovi/symfony-flex-docker/blob/master/.infra/docker/db/Dockerfile) to say `FROM postgres:9.6-alpine` instead (comment out other stuff for now)
2. replace the env var names in main [docker-compose.yaml](https://github.com/dkarlovi/symfony-flex-docker/blob/master/docker-compose.yaml#L26-L30) to match names from the [image docs](https://hub.docker.com/_/postgres/)
3. mount the [volume](https://github.com/dkarlovi/symfony-flex-docker/blob/master/docker-compose.yaml#L34) at `/var/lib/postgresql` instead of `/var/lib/mysql`
4. add `pdo_pgsql` instead of `pdo_mysql` in [app Dockerfile](https://github.com/dkarlovi/symfony-flex-docker/blob/master/.infra/docker/app/Dockerfile#L7)

You should be all set now.

dkarlovi/symfony-flex-docker/0.1/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+    "copy-from-package": {
+        ".infra/": ".infra",
```
jderusse: Directory missing in this PR?

dkarlovi/symfony-flex-docker/0.1/Makefile
```diff
@@ -0,0 +1,51 @@
...
+docker-start-prod:
+	docker-compose -f ./docker-compose.yaml -f ./.infra/docker-compose/docker-compose.prod.yaml up -d
+.PHONY: docker-start-prod
+
+docker-start-test:
+	docker-compose -f ./docker-compose.yaml -f ./.infra/docker-compose/docker-compose.test.yaml up -d
+.PHONY: docker-start-test
+
+docker-build:
+	docker-compose build
```
jderusse: what if `/.infra/docker-compose/docker-compose.env.yml` file contains something to build which is not présent (or not with the same parameters) than ` ./docker-compose.yaml`?

dkarlovi/symfony-flex-docker/0.1/bin/dphp
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
jderusse: What's the point of this file? Why not simply using a "tools" service with docker-compose for such task?

dkarlovi/symfony-flex-docker/0.1/manifest.json
dkarlovi: Nope, it's in [the package](https://github.com/dkarlovi/symfony-flex-docker) from which this recipe is copying.

dkarlovi/symfony-flex-docker/0.1/Makefile
dkarlovi: There is no "environment-specific build", the idea is to use **exactly the same images on production** you use in development, by pushing them (from your CI server) to a [Docker registry](https://docs.docker.com/registry/) and pulling them from production (say, Kubernetes or Openshift).

The environment-specific Compose files are for runtime-only, not for build.

dkarlovi/symfony-flex-docker/0.1/bin/dphp
dkarlovi: Because this is not part of the project, you use a [preconfigured utility Docker container](https://github.com/dkarlovi/php-toolbox) which carries all the QA and such tools so your `vendor/` folder is not bloated (with this approach, I've reduced my `vendor/` folder size by 50%), also removing some nasty conflicts in my Composer dependencies.

This way, you build your app without development dependencies (again, there is no "environment-specific environment", you build once and run everywhere) and ship your development tools in a separate container, managed elsewhere.

Also, the toolbox container invocation script is setup so that you can use it like a custom PHP interpreter in PhpStorm, ie. you point PhpStorm directly to this file as a PHP interpterer and it automatically can run your tests and tools INSIDE your local container cluster.

teohhanhui:
I'm -1. See https://github.com/symfony/recipes/issues/78#issuecomment-305100020 for reasoning.

Don't get me wrong. I'd love to see better Docker support, as raised in symfony/flex#54. (I'm currently adapting my existing project with a Docker Compose setup to Symfony Flex.) But I think this is not the right approach.

dkarlovi:
@teohhanhui thanks, I see your point, but strongly disagree with it.

You don't want Flex to be coupled to a 3rd party technology, especially in the Docker world where things are made and broken every month. This is true for Docker, but also for stuff like Heroku (which Symfony is tied to in a few places). If you couple to a tech, you not only have a possible SPOF, you also subtly align your solution to that tech which might make [using other, "non-approved" tech harder](https://github.com/symfony/recipes/issues/64).

I think the approach of having Flex be a generic "do stuff to your Symfony app" tool, with recipes being able to hook into the process and do powerful transformations is the best way to go.

What I DO think is a mistake is having this repo be a hub for making your own recipes instead of [shipping it with the package itself](https://github.com/symfony/flex/issues/70). This is visible with this PR where a very basic recipe is being reviewed for a week.

sstok:
If this one works it can be approved (by anyone but the author), this recipe is part of contrib. So in this case it's not a matter if whether you need it, but whether this is good to work 👍

PR #16, opened 2017-05-08
Add Enqueue recipes.

makasim:
| Q             | A
| ------------- | ---
| License       | MIT

This pull request adds a Symfony Flex recipe for [Enqueue](https://github.com/php-enqueue/enqueue-dev) bundle and two transports `filesystem` and `amqp`. 

Enqueue provides solutions to deal with message queues.


enqueue/amqp-ext/0.3/etc/packages/enqueue-amqp.yaml
```diff
@@ -0,0 +1,23 @@
+enqueue:
```
robfrawley: [Imagine Bundle](https://github.com/liip/LiipImagineBundle) could benefit from such selective config, as well.

makasim: > Maybe this shows something we could propose as a new feature for flex: if specific package is installed, install more configuration, else, just "suggest" it.

This might be a solution, though I support the point not to adding conditions to recipes. 

The packs could be a solution to my problem. You do not install the bundle and amqp-ext separately but a `enqueue/amqp-pack` package. It installs the bundle and the transport and puts required configuration where needed

enqueue/fs/0.3/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/",
+        "var/": "%VAR_DIR%"
+    },
+    "env": {
+        "ENQUEUE_FS_STORAGE": "%VAR%/queue"
+    },
+    "gitignore": [
+        "/var/queue"
```
stof: the whole content of `var` is already ignored in git, so this seems useless

enqueue/fs/0.3/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/",
+        "var/": "%VAR_DIR%"
+    },
+    "env": {
+        "ENQUEUE_FS_STORAGE": "%VAR%/queue"
```
stof: are you sure about `%VAR%` here ?

enqueue/fs/0.3/manifest.json
makasim: Yeah, I know. It is done in framework bundle and what if someone did not install it. That is here just for that case.

enqueue/fs/0.3/manifest.json
makasim: fixed

enqueue/fs/0.3/manifest.json
makasim: I could remove it if the case is not something we should worry about.

fabpot: Framework bundle is a requirement, it cannot be uninstalled or not installed, so this is safe to remove.

enqueue/amqp-ext/0.3/etc/packages/enqueue_amqp.yaml
```diff
@@ -0,0 +1,23 @@
+enqueue:
+    transport:
+        default: 'amqp'
+        amqp:
+            host: '%env(ENQUEUE_AMQP_HOST)%'
+            port: '%env(ENQUEUE_AMQP_PORT)%'
+            login: '%env(ENQUEUE_AMQP_LOGIN)%'
+            password: '%env(ENQUEUE_AMQP_PASSWORD)%'
+            vhost: '%env(ENQUEUE_AMQP_VHOST)%'
```
fabpot: I've have had a look, but do you support a URL/DSN as well? That's what we are using for Doctrine and Swiftmailer, so that would be great if possible to do the same here as well.

enqueue/fs/0.3/manifest.json
stof: btw, your own enqueue-bundle depends on FrameworkBundle 😄

enqueue/enqueue-bundle/0.3/post_install.txt
```diff
@@ -0,0 +1 @@
+[Enqueue] Run <comment>make consume</> command to consume messages
```
fabpot: I would remove this as the post install script if for things you need to do to make things work. It should not replace proper documentation of the package.

enqueue/amqp-ext/0.3/etc/packages/enqueue_amqp.yaml
stof: agreed here. PaaS expose credentials as a URL, so this config is not usable drop-in on such PaaS.

enqueue/fs/0.3/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/",
+        "var/": "%VAR_DIR%"
```
fabpot: Is the `var/queue/.gitkeep` really needed? The package should be able to create it when it does not exist, right?

enqueue/fs/0.3/manifest.json
makasim: I overcomplicated things. fixed.

enqueue/amqp-ext/0.3/etc/packages/enqueue_amqp.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+#        If you use RabbitMQ broker you can replace amqp transport with this one. It enables specific features like priorities, delays.
+#        default: 'rabbitmq_amqp'
+#        rabbitmq_amqp:
+#
+#            host: '%env(ENQUEUE_AMQP_HOST)%'
+#            port: '%env(ENQUEUE_AMQP_PORT)%'
+#            login: '%env(ENQUEUE_AMQP_LOGIN)%'
+#            password: '%env(ENQUEUE_AMQP_PASSWORD)%'
+#            vhost: '%env(ENQUEUE_AMQP_VHOST)%'
```
fabpot: The whole config seems to be the same as the default one, except for the option. What about adding the comment on the default config with the explanation you have above?

makasim: The transport name (a key) is different: `amqp` and `rabbitmq_amqp`.

Pierstoval: Then the `If you use RabbitMQ broker ...` part could just say _"rename `amqp` with `rabbitmq_amqp` and the bundle will enable specific features ..."_

fabpot:
Did you close on purpose?

enqueue/enqueue-bundle/0.3/Makefile
```diff
@@ -0,0 +1,3 @@
+consume:
+	@test -f bin/console && bin/console enqueue:consume --setup-broker || echo "cannot consume messages (needs symfony/console)"
+.PHONY: consume
```
fabpot: I think this can be removed as well. We don't want to wrap all Symfony CLI commands as make tasks. We are using Makefile tasks for when this is not practical or not possible to use the Symfony CLI.

makasim:
> https://github.com/symfony/recipes-contrib/pull/16#issuecomment-300883395

it was closed when  I merged https://github.com/php-enqueue/enqueue-dev/pull/78. I wrote there "fixes XXX".

enqueue/enqueue-bundle/0.3/etc/packages/test/enqueue.yaml
```diff
@@ -0,0 +1,6 @@
+enqueue:
+    transport:
+        default: 'null'
```
javiereguiluz: This `'null'` value must be a string or could be a normal `null` value instead?

enqueue/enqueue-bundle/0.3/etc/packages/test/enqueue.yaml
```diff
@@ -0,0 +1,6 @@
...
+        'null': ~
```
javiereguiluz: I don't understand this line: `'null': ~`

enqueue/fs/0.3/manifest.json
```diff
@@ -0,0 +1,9 @@
...
+    },
+    "env": {
+        "ENQUEUE_FS_STORAGE": "%VAR_DIR%/queue"
```
javiereguiluz: Just asking. If the bundles is called "enqueue", should the var/ dir be called "enqueue/" too? The "queue" name is too generic ... what if someday Symfony releases a Queue component or bundle?

enqueue/enqueue-bundle/0.3/etc/packages/test/enqueue.yaml
stof: it defines the configuration for the NullTransport. But the key must be quoted in YAML, as unquoting it would parse it as `null`, not as the string `'null'`

javiereguiluz: This is extremely confusing to me: a "default transport" which is a null string ... which then is a "null transport" with a null value. Can we think a better approach for this? A new config option for tests maybe? Thanks!

makasim: @javiereguiluz In this context 'null' is the name of the transport. The transport does not send nor receive anything. It is there just to mock a messaging API and useful in tests for example. 

I've not come up with a better name for the transport.

enqueue/amqp-ext/0.3/etc/packages/enqueue_amqp.yaml
makasim: fixed

enqueue/amqp-ext/0.3/etc/packages/enqueue_amqp.yaml
makasim: fixed

enqueue/enqueue-bundle/0.3/Makefile
makasim: fixed

enqueue/enqueue-bundle/0.3/etc/packages/test/enqueue.yaml
stof: Noop ? Blackhole ?

makasim: Maybe, but not now. That's a huge change and would take time. I'll do it later on. 

Is it a blocker?

makasim:
I address all the review comments but the PR still red. It says (@stof stof requested changes). What should I do to this?

fabpot:
@stof needs to dismiss his review and someone else (including stof) need to approve the PR.

enqueue/enqueue-bundle/0.3/etc/packages/test/enqueue.yaml
robfrawley: Having a "null" object is very common and is, of course, a [well know design pattern](https://en.wikipedia.org/wiki/Null_Object_pattern); why are we trying to rename something that already describes exactly what the transport is?

makasim:
> @stof needs to dismiss his review and someone else (including stof) need to approve the PR.

@stof I kindly remind you.

enqueue/fs/0.4/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "env": {
+        "ENQUEUE_DSN": "file:/%VAR_DIR%/enqueue"
```
xabbuh: Will the library take care to create the directory if it does not exist yet?

makasim: yes

javiereguiluz: Shouldn't we use here `file://` instead of `file:/` ?

makasim: I followed the syntax I saw in [chrome's  address bar](https://www.dropbox.com/s/2ajw18yjqp04otu/Screenshot%202017-05-24%2009.47.06.png?dl=0)

If `%VAR_DIR%` is `/home/foo/acme/var` the dsn become `file://home/foo/acme/var/enqueue`

Pierstoval: It's a problem. On Unix system, using `file://` will result in paths like `file:///var/www/dev/enqueue` which might not work, whereas on windows it will result in `file://e:/www/dev/enqueue` which should work.

Dunno how to fix this 😱

javiereguiluz: I don't know about Linux, but on macOS `file:///var/www/dev/enqueue` not only works, but it's the only way this can work: 3 `/` after `file:` is normal.

Pierstoval: Ok nevermind, I tested on my ubuntu distro and `file:///` works, once again I should check before talking

makasim: I opened the issue. It'will be fixed.

PR #18, opened 2017-05-12
Added a recipe for IvoryCKEditorBundle

javiereguiluz:
About the bundle:

* 1.2 million downloads
* Code: https://github.com/egeloen/IvoryCKEditorBundle
* Docs: https://symfony.com/doc/master/bundles/IvoryCKEditorBundle/index.html

---

**First** I have a question. I still don't understand how the versioning of recipes work. These are the branches and tags of the bundle:

| Branches | Tags
| --- | ---
| branches | tags

I've chosen `2.0` as the version of the recipe. Is it correct or wrong? How can I know?

---

**Second**, I have a question. The bundle uses by default a very complex toolbar for the WYSIWYG editor (it shows all options available in CKEditor). Instead of relying on the default config, I'm proposing that the recipe provides a "better" default config ("better" is just my opinion, of course). This is how it'd look the editor now:

better-default-config

So, my question: is allowed/expected that recipes "improve" (or at least change) the default config of the bundle to provide a better user experience?


javiereguiluz:
There are validation errors but I can't see them:

recipe-errors

stof:
@javiereguiluz if your recipe works for 2.0 of the bundle, it is fine to start the versioning at 2.0. but then, you need to add version aliases for next major versions.

@fabpot does flex server require having aliases for all minor versions or could it assume semver (maybe with an option in the recipe to opt-out) and automatically try to find a recipe for a previous minor version (of the same major one) if it does not have a recipe for the requested version ? This would avoid having to open a PR on the recipes repository each time the minor version is bumped to add new features in a BC way. This would be especially useful for cases where the recipe is not contributed by the bundle maintainer themselves (as is the case for this PR AFAIK)

stof:
However, as the version 2.0 of CKEditorBundle is for Symfony 2.0.x, I would not try to make it flex-ready. Start with 5.x only IMO (not having a maintenance branch for 4.x probably means that it is considered EOL).

Pierstoval:
> So, my question: is allowed/expected that recipes "improve" (or at least change) the default config of the bundle to provide a better user experience?

I think it's an awesome idea, providing a default toolbar that shows the most necessary stuff. It is also great in the fact that this is totally the place where you can add a link to toolbar configuration documentation 🙂

fabpot:
@javiereguiluz I've fixed the error page, you can now see the error messages.

fabpot:
The package name is `egeloen/ckeditor-bundle` apparently.

PR #19, opened 2017-05-12
Added HTTPlug manifest

Nyholm:
Added manifest for HTTPlugBundle 

We have a custom profiler page so we need to run assets:install. 

Also, is it okey to have "get started"-config like this? (commented)
The bundle works with zero config but most user probably want to add something.


fabpot:
Having the default configuration is being discussed in another PR. The conclusion seems to be that we don't want to duplicate the configuration. So, you should only keep config settings that people will probably change. If none, just remove the file.

php-http/httplug-bundle/1.5/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "bundles": {
+        "Http\\HttplugBundle\\HttplugBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    },
+    "composer-scripts": {
+        "assets:install --symlink --relative %WEB_DIR%": "symfony-cmd"
```
fabpot: This is not needed as already part of framework-bundle, so it's aways installed.

Pierstoval:
> So, you should only keep config settings that people will probably change. If none, just remove the file.

If the file is removed, no recipe is needed as Flex already configures bundles if they have `symfony-bundle` as "type" option in the `composer.json` file

Nyholm:
I've updated the PR according to feedback. 
The config is no longer a comment. I believe it is a "good to get started" config. 

@dbu or @fbourigault might want to say 👍  or suggest changes to the config.

php-http/httplug-bundle/1.6/etc/packages/httplug.yaml
```diff
@@ -0,0 +1,16 @@
+httplug:
+    plugins:
+        logger:
+            enabled: true
```
stof: shouldn't this be `%kernel.debug%` ?

Nyholm: Excellent question. I want to log my HTTP request in production. 

Should I put your suggestion in a comment?

dbu: or maybe use %kernel.debug% as default and add a comment that you can set this to true if you want to log HTTP requests in production?

php-http/httplug-bundle/1.6/etc/packages/httplug.yaml
```diff
@@ -0,0 +1,16 @@
...
+        redirect:
+            preserve_header: true
+
+    discovery:
+        client: 'auto'
+
+    clients:
+        acme:
```
dbu: is it best practice to make a default client called acme?

stof:
> and do we need to add a new recipe each time we tag a new minor version of httplug-bundle?

no. the flex server will reuse this recipe for any future version (even major one), until you create a new recipe for a version. So you only need to create a new recipe when you need a different content.

Nyholm:
> the bundle depends on a virtual package to provide an adapter...

Good question. Will flex handle this and suggest to install implementations of the virtual package? Or should we improve the recipe somehow to install implementations ourselves?

php-http/httplug-bundle/1.6/etc/packages/httplug.yaml
Pierstoval: I would rather go for `app`

stof:
> should the recipe install guzzle6 adapter by default so that there is an actual client? or how exactly does that work?

The recipe cannot alter the require command to inject a new dependency. It cannot alter requirements of a package (and it should not). Recipes are only downloaded after the solver ran and selected the appropriate package.

And making your bundle depend explicitly on the guzzle6-adapter would be a bad idea, as it would make it impossible to use the guzzle5-adapter (you cannot install both together due to dependency conflicts for guzzle).
The solution here is probably a documentation one on your side: keep recommending `composer require php-http/httplug-bundle php-http/guzzle6-adapter`, which is what you already need today.

Nyholm:
Thank you @stof. 

Im happy with this PR now if there is no other feedback.

Nyholm:
Thank you for the reviews and for merging.

PR #21, opened 2017-05-12
Add a note about PRs being merged automatically

fabpot:


yceruto:
What if I'm no ready to merge the PR? if someone approve it I lose the PR?

What do you think about a label to deny the auto-merge, something like `wip`? later when the author remove this label, @symfony-bot merges then.

yceruto:
... or maybe better by using the PR title, if '[WIP]...' don't auto-merge?

fabpot:
I've just updated the checks to be more strict and predictable.

robfrawley:
@fabpot Out of interest, (in your language preferences), what did you write @symfony-bot in?

fabpot:
@robfrawley It's written in Go :)

stof:
@fabpot does it also handle the case where 1 reviewer requested changes and another one approves ? The PR should not be merged until the first reviewer approves again, or their review gets dismissed.

fabpot:
@stof Yes, taht's taken into account as we are using the Github green button to merge, which enforces that rule for us.

robfrawley:
@fabpot What about also allowing repository "collaborators" (those with push access) to approve these recipe PRs (to cause an auto-merge); this may be important in some circumstances.

For example, with regard to #14, no members of @liip actively maintain [liip/LiipImagineBundle](https://github.com/liip/LiipImagineBundle); instead, it has been me and two or three others developing and maintaining the repository. None of us are members of @liip but we all have "collaborator" status. I suspect this situation isn't uncommon for other projects, as well.

stof:
@robfrawley The list of repo collaborators is not a public resource in the Github API. Only people with push access to the repo can get the list. So the symfony-bot cannot use this info.

PR #23, opened 2017-05-14
Added DoctrineFixturesBundle

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT

This recipe is an experiment to see if you agree with it. It's OK to create an empty fixtures file when installing the bundle ... or do you prefer just to create an empty `src/DataFixtures/ORM/` dir ... or even just `src/DataFixtures/` to not upset those not using ORM?


fabpot:
Having the PHP skeleton file sounds good to me.

xabbuh:
Doesn't the bundle provide a command to create a fixtures skeleton? In that case, I would be in favour to remove the example, but only create the directory.

javiereguiluz:
OK, I've removed the fixture file for now. We can revisit this decision if this PR is merged and the best practices for creating fixtures change a bit: https://github.com/doctrine/DoctrineFixturesBundle/pull/189

javiereguiluz:
I don't understand this error from Symfony Bot:

symfony-bot-error

Pierstoval:
@javiereguiluz Tried adding the MIT notice as in the [PR template](https://github.com/symfony/recipes-contrib/blob/master/.github/PULL_REQUEST_TEMPLATE.md)?

PR #24, opened 2017-05-14
Added SncRedisBundle

javiereguiluz:
About the bundle:

* The most popular Redis bundle for Symfony
* 3 million downloads
* https://github.com/snc/SncRedisBundle

### DIFF
--- a/snc/redis-bundle/2.0/etc/packages/snc_redis.yaml
+++ b/snc/redis-bundle/2.0/etc/packages/snc_redis.yaml
@@ -1,0 +1,6 @@
+snc_redis:
+    clients:
+        default:
+            type: predis
+            alias: default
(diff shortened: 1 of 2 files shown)


snc/redis-bundle/2.0/etc/packages/snc_redis.yaml
```diff
@@ -0,0 +1,6 @@
+snc_redis:
+    clients:
+        default:
+            type: predis
+            alias: default
+            dsn: "%env(REDIS_URL)%"
```
xabbuh: IIRC correctly, SncRedisBundle does not support configuring the DSN through a single environment variable.

xabbuh: see snc/SncRedisBundle#302

PR #25, opened 2017-05-14
Added FOSJsRoutingBundle

javiereguiluz:


maidmaid:
What about move ``FOSJsRoutingBundle`` in ``symfony/recipes``?

stof:
@maidmaid I'm actually not a fan of exposing the url generator to the JS code itself. I used this bundle in the past, but then I stopped and decided to write my JS code in a way relying on receiving a URL from the backend (generally through data attributes) rather than just the route parameters. I found it much more maintainable, as the JS code is decoupled from the backend routing.
So I'm not sure it should be included in the main recipes repo yet (I could change my mind if presented with arguments to convince me though, it is not a permanent veto). But anyway, this should not be discussed on the merged PR (other people will probably miss the discussion)

PR #28, opened 2017-05-14
(WIP) Added GregwarCaptchaBundle

javiereguiluz:
About the bundle:

* 600,000 downloads
* https://github.com/Gregwar/CaptchaBundle

---

@Gregwar as the author of the bundle, I'd like to ask you if the routing configuration should always be enabled by default or if it' better to add it commented and let the users enable it if they need it. Thanks!

### DIFF
--- a/gregwar/captcha-bundle/2.0/etc/packages/gregwar_captcha.yaml
+++ b/gregwar/captcha-bundle/2.0/etc/packages/gregwar_captcha.yaml
@@ -1,0 +1,4 @@
+gregwar_captcha:
+    width: 200
+    height: 50
+    length: 6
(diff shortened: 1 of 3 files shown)


Gregwar:
@javiereguiluz it could be optional, because the captcha can also be embedded using inlined base64 data

What you suggest is having the routing configuration in the README and ask the user to add a piece of configuration if they want to use the routes?
Can we discuss this on an issue on the bundle repo?

sstok:
> Symfony2 bundle implementing a "captcha" form type.

You might want to change this 😉

javiereguiluz:
If we remove the loading of the routing file, this recipe is no longer needed. Si I'm closing for now and we can rethink this later or discuss it in the original repo, as @Gregwar suggested. Thanks!

robfrawley:
Can you add a reference to the discussion thread in the original repo? Doing this in general when these are closed for further discussion would be very helpful, as well.

PR #29, opened 2017-05-14
(WIP) Added WebuniCommonMarkBundle

javiereguiluz:
About the bundle:

* https://github.com/webuni/commonmark-bundle

---

Although the bundle is still tiny in number of downloads, it integrates the popular CommonMark PHP library (https://github.com/thephpleague/commonmark) which is the reference implementation of Markdown in PHP and other popular (and great!) extensions (https://github.com/webuni/commonmark-attributes-extension and https://github.com/webuni/commonmark-table-extension).

@hason as the author of the bundle and the extensions, could you please review this recipe? Thanks!

### DIFF
--- a/webuni/commonmark-bundle/0.2/etc/packages/webuni_commonmark.yaml
+++ b/webuni/commonmark-bundle/0.2/etc/packages/webuni_commonmark.yaml
@@ -1,0 +1,6 @@
+webuni_commonmark:
+    extensions:
+        # adds support for defining attributes on HTML elements (e.g. 'id', 'class', 'style')
+        attributes: true
+        # adds support for parsing Markdown tables
+        table: true
--- a/webuni/commonmark-bundle/0.2/manifest.json
+++ b/webuni/commonmark-bundle/0.2/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Webuni\\Bundle\\CommonMarkBundle\\WebuniCommonMarkBundle": ["all"]
+    },
+    "copy-from-recipe": {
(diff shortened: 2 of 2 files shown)


webuni/commonmark-bundle/0.2/etc/packages/webuni_commonmark.yaml
```diff
@@ -0,0 +1,6 @@
+webuni_commonmark:
+    extensions:
+        # adds support for defining attributes on HTML elements (e.g. 'id', 'class', 'style')
+        attributes: true
+        # adds support for parsing Markdown tables
+        table: true
```
stof: this repeats the default values.

Thus, it does not always enable it: setting it to true here will have an effect only if the extension is installed in your project (which btw is a weird behavior. It would be better to change the default value based on installation status and then enforce to have the extension when enabling it. Much less WTF)

javiereguiluz: Yes, but the problem is that the bundle doesn't have docs yet. This exposes the options so you can know about them: enabling the tables extension is a no-brainer to me, but the attributes extension is debatable.

About your comment, I thought the bundle installed both extensions ... but they are only in "require-dev". So, are you asking me to do changes in this reicpe or is your proposal something to be fixed in the bundle? Thanks!

stof: I think the change should be done in the bundle itself (and doc should be contributed there too btw)

javiereguiluz:
closing as won't fix because of the reasons given by Christophe. Thanks!

PR #30, opened 2017-05-14
Added MisdPhoneNumberBundle

javiereguiluz:
About this bundle:

* ~700,000 downloads
* https://github.com/misd-service-development/phone-number-bundle

---

This recipe is another experiment. What is the right way to register the PhoneNumberType DBAL type dynamically? I guess the provided `libphonenumber.yaml` is wrong and we'd need some PHP file that detectes if Doctrine exists and then registers the type. Can anyone help me doing that? Thanks!

### DIFF
--- a/misd/phone-number-bundle/1.2/etc/packages/libphonenumber.yaml
+++ b/misd/phone-number-bundle/1.2/etc/packages/libphonenumber.yaml
@@ -1,0 +1,4 @@
+doctrine:
+    dbal:
+        types:
+            phone_number: Misd\PhoneNumberBundle\Doctrine\DBAL\Types\PhoneNumberType
--- a/misd/phone-number-bundle/1.2/manifest.json
+++ b/misd/phone-number-bundle/1.2/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Misd\\PhoneNumberBundle\\MisdPhoneNumberBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    }
+}


javiereguiluz:
I'm not sure about this merge :(  I asked a question which I think it's important to solve, because currently this recipe is broken. Will the doctrine DBAL config defined in this recipe work? I don't think so.

Of course I caused this issue because I forgot to add "WIP" to the title of this recipe. Can we revert the merge please? Thanks!

sstok:
Whoops, sorry 😅

fabpot:
Reverted

fabpot:
@javiereguiluz You will need to recreate a new PR.

xabbuh:
@javiereguiluz I think now that merges are done automatically, in the future you will have to file a review requesting changes to prevent automatic merges (in case @stof's suggestion made in https://github.com/symfony/recipes-contrib/pull/21#issuecomment-301419942 is taken into account which IMO would be a good idea).

stof:
well, for experiments, the solution is to mark it as WIP

javiereguiluz:
Yes, it's easier for everybody to just add "WIP" in the title. I'm doing that for most of my PRs.

yceruto:
> well, for experiments, the solution is to mark it as WIP

Add "WIP" in the title already avoid auto-merge?

xabbuh:
> Add "WIP" in the title already avoid auto-merge?

yes, see https://github.com/symfony/recipes-contrib/#contributing

yceruto:
@xabbuh Great! I missed these changes, thank you :)

PR #31, opened 2017-05-14
Added BazingaGeocoderBundle

javiereguiluz:
| License       | MIT

About the bundle:

* ~550,000 downloads
* https://github.com/geocoder-php/BazingaGeocoderBundle

---

@Nyholm as you are the new maintainer of the bundle, I'd like to ask you something about the `fake_ip` config recommended for the "dev" environment. Does it have to be a wrong IP as the example showed in https://github.com/geocoder-php/BazingaGeocoderBundle/blob/master/Resources/doc/index.md#killer-feature ... or can be a valid IP?

### DIFF
--- a/willdurand/geocoder-bundle/4.1/etc/packages/bazinga_geocoder.yaml
+++ b/willdurand/geocoder-bundle/4.1/etc/packages/bazinga_geocoder.yaml
@@ -1,1 +1,0 @@
-bazinga_geocoder: ~
--- a/willdurand/geocoder-bundle/4.1/etc/packages/bazinga_geocoder.yaml
+++ b/willdurand/geocoder-bundle/4.1/etc/packages/bazinga_geocoder.yaml
@@ -1,0 +1,1 @@
+bazinga_geocoder: ~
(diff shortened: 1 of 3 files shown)


Nyholm:
It should be a valid IP. The docs are wrong. Thank you. 

https://github.com/geocoder-php/BazingaGeocoderBundle/pull/118

willdurand/geocoder-bundle/4.1/etc/packages/dev/bazinga_geocoder.yaml
```diff
@@ -0,0 +1,2 @@
+bazinga_geocoder:
+    fake_ip: 123.123.123.123
```
fabpot: Would it make sense to add a comment linking to the docs about this feature?

willdurand/geocoder-bundle/4.1/etc/packages/bazinga_geocoder.yaml
```diff
@@ -0,0 +1 @@
+bazinga_geocoder: ~
```
stof: should it really be there as there is no config in it ?

Nyholm:
FYI: A version 5 is coming out this month. This PR is still valid but I need to create a new PR for 5.0. The changes to the config is minimal but we have a multi-repo approach now. Which means that one have to install a bunch of geocoder providers. 

Just FYI..

PR #32, opened 2017-05-14
(WIP) Added LexikFormFilterBundle

javiereguiluz:
About this bundle:

* ~400,000 downloads
* Voted in the Top 30 Most Useful Bundles (https://symfony.com/blog/the-30-most-useful-symfony-bundles-and-making-them-even-better)
* https://github.com/lexik/LexikFormFilterBundle

---

This recipe is another experiment. I don't know it it works. I want to add a new form theme to Twig when the bundle is installed. Of course I don't want to lose the existing form themes configured in other files. How can I do that? Thanks!


javiereguiluz:
Closing because of no feedback and because the recipe is clearly not going to work.

PR #35, opened 2017-05-14
Added AlgoliaAlgoliaSearchBundle

javiereguiluz:
About the bundle:

* ~70,000 downloads
* It integrates the amazing Algolia search service (we use it on symfony.com for example)
* https://github.com/algolia/AlgoliaSearchBundle

---

I have a question about the env vars. As you can see, in the default config I've used "realistic values" for the APP_ID and API_KEY (the placeholders are similar to the real values that you'll put there). Do you like that ... or do you prefer to put some generic placeholder? Example:

```json
"env": {
    "ALGOLIA_APP_ID": "you app ID",
    "ALGOLIA_API_KEY": "your API key"
}
```

License MIT

### DIFF
--- a/algolia/algolia-search-bundle/2.2/etc/packages/algolia.yaml
+++ b/algolia/algolia-search-bundle/2.2/etc/packages/algolia.yaml
@@ -1,0 +1,3 @@
+algolia:
+    application_id: "%env(ALGOLIA_APP_ID)%"
+    api_key: "%env(ALGOLIA_API_KEY)%"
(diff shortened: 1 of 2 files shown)


xabbuh:
I would prefer some descriptive placeholders

javiereguiluz:
@xabbuh the idea is to give the env vars names that you cannot misuse. In this case, `ALGOLIA_APP_ID` and `ALGOLIA_API_KEY` "can't" be misused, right?

The question was: should we use a description as the value of the env var ... or should we use a value that resembles the real value?

xabbuh:
I think both (the var name and its placeholder) should be descriptive. Having values that look "real" makes it a bit harder to spot if you already adjusted the config according to your needs.

javiereguiluz:
👍  that's a very good point

PR #36, opened 2017-05-14
(WIP) Added MisdPhoneNumberBundle

javiereguiluz:
About this bundle:

* ~700,000 downloads
* https://github.com/misd-service-development/phone-number-bundle

---

**My question**: which is the right way to register the PhoneNumberType DBAL type dynamically?

I guess the provided `libphonenumber.yaml` is wrong and we'd need some PHP file that detectes if Doctrine exists and then registers the type. Can anyone help me doing that? Thanks!

### DIFF
--- a/misd/phone-number-bundle/1.2/etc/packages/libphonenumber.yaml
+++ b/misd/phone-number-bundle/1.2/etc/packages/libphonenumber.yaml
@@ -1,0 +1,4 @@
+doctrine:
+    dbal:
+        types:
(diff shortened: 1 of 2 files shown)


stof:
@javiereguiluz I think the solution would be for the bundle to use a PrependExtension to inject the config in the DoctrineBundle config

javiereguiluz:
Closing because this recipe is not going to work ... and the change can be made in the upstream repository. Thanks!

PR #38, opened 2017-05-17
Add recipe for Wisembly AMQP bundle

Taluu:
| Q             | A
| ------------- | ---
| License       | MIT

Integration for Wisembly's AMQP bundle, which offers a slightly different (well not so different now...) than the SwarrotBundle. Works with Swarrot.

### DIFF
--- a/wisembly/amqp-bundle/1.3/manifest.json
+++ b/wisembly/amqp-bundle/1.3/manifest.json
@@ -1,0 +1,11 @@
+{
+    "bundles": {
+        "Wisembly\\AmqpBundle\\WisemblyAmqpBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "AMQP_URL": "amqp://guest:guest@localhost:5672"
+    }
+}
--- a/wisembly/amqp-bundle/1.3/config/packages/wisembly_amqp.yaml
+++ b/wisembly/amqp-bundle/1.3/config/packages/wisembly_amqp.yaml
@@ -1,0 +1,16 @@
+wisembly_amqp:
+    console_path: '%kernel.project_dir%/bin/console'
+    broker: 'oldsound'
+    connections:
+        default: '%env(AMQP_URL)%'
+    gates:
+        # here you should declare all your "gates", which are the cornerstone
+        # of this amqp bundle.
+        #
+        # For each gates declares which queue and which exchange it should
+        # communicate with, if these should be auto_declared, which connection
+        # it should use, ... and so on.
+        acme:
+            auto_declare: true
+            queue: acme_queue
+            exchange: acme_exchange


wisembly/amqp-bundle/1.3/post-install.txt
```diff
@@ -0,0 +1,8 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * <fg=blue>Add</> "gates" to the bundle configuration
```
javiereguiluz: Is this phrase easy to understand for people installing the bundle? I know I don't understand a word of it (*"Add gates to the bundle config"* ?!?!?!) but maybe it's because I'm not skilled in this AMQP thing.

wisembly/amqp-bundle/1.3/etc/packages/amqp.yaml
```diff
@@ -0,0 +1,7 @@
+wisembly_amqp:
```
javiereguiluz: We never add blank lines in the config ... and in this case the config is super small. I think we should remove those blank lines.

wisembly/amqp-bundle/1.3/Makefile
```diff
@@ -0,0 +1,16 @@
...
+	$(error GATE not set)
+endif
+
+gate:
+
+consume: gate $(CONSOLE)
+	$(CONSOLE) wisembly:amqp:consume $(GATE) $(OPTIONS)
+
+publish: gate $(CONSOLE)
+	$(CONSOLE) wisembly:amqp:publish $(GATE) "$(MESSAGE)"
```
stof: why making people define env variables when calling the command instead of using arguments ?

Btw, if arguments are used, I don't think wrapping the Symfony command in the makefile makes much sense, as they would just be proxied directly

Taluu: It's used as in `make consume GATE=my-gate` or `make publish GATE=my-gate MESSAGE=""`, but I also thought it would be overkill (the `GATE` env var is required though).

This is the only way I could find to use "arguments" in the Makefile, but then you are probably right by saying that this would be overkill, and better to directly use the command instead... but then, should in the Makefile only be commands without required arguments ?

wisembly/amqp-bundle/1.3/post-install.txt
Taluu: It's actually the concept of the bundle, not AMQP. 

I thought I documented that, but it doesn't seem to be the case, so indeed, only those familiar with the bundle (not a lot of people currently TBH) would understand that. I'll document that.

wisembly/amqp-bundle/1.3/etc/packages/amqp.yaml
Taluu: fixed

wisembly/amqp-bundle/1.3/Makefile
stof: well, this is also why we don't want to proxy all commands through the Makefile: we loose the possibility to use options to refine the command (and so people still need to know the actual command to use it with options or to see the help of the command, which relies on passing the `-h` option).
We provide Makefile commands for tasks which involve multiple Symfony commands, or additional command not being part of the Symfony console

wisembly/amqp-bundle/1.3/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "Wisembly\\AmqpBundle\\WisemblyAmqpBundle": ["all"]
+    },
+
+    "copy-from-recipe": {
```
Pierstoval: No need for blank lines in the manifest either

Taluu:
Finally took some time to update it. I also added an acme "gate" with some explanation in the generated config file.

wisembly/amqp-bundle/1.3/post-install.txt
Taluu: Documented here : https://github.com/Wisembly/AMQPBundle#concept-of-gates

Taluu: But as I added a acme gate example and documented it a bit more in the config file, I think that the post-install is now unnecessary now.

fabpot: Please remove it then. We try to limit display during installation as much as possible as it's not permanent.

fabpot:

wisembly/amqp-bundle/1.3/config/packages/amqp.yaml
```diff
@@ -0,0 +1,16 @@
+wisembly_amqp:
+    console_path: '%kernel.project_dir%/bin/console'
+    broker: 'oldsound'
+    connections:
+        default: '%env(AMQP_URI)%'
```
fabpot: We are using `DATABASE_URL` or `MAILER_URL`, so what about using `AMQP_URL` here?

Taluu: Changed.

PR #40, opened 2017-05-17
(WIP) Added SentryBundle

GromNaN:
References:
* https://sentry.io/for/symfony/
* https://docs.sentry.io/clients/php/integrations/symfony2/

This bundle is not required to get error reported to Sentry. The Symfony logger has a listener on uncaught exceptions and Monolog has the `RavenHandler` that send messages to Sentry.

---

I have a question regarding the "release" config. This is more related to the version of the code that an environment variable. 

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sentry/sentry-symfony/0.7/etc/packages/prod/sentry.yaml
+++ b/sentry/sentry-symfony/0.7/etc/packages/prod/sentry.yaml
@@ -1,0 +1,3 @@
+sentry:
+    dsn: "%env(SENTRY_DSN)%"
+    release: "%env(APP_RELEASE)%"
--- a/sentry/sentry-symfony/0.7/manifest.json
+++ b/sentry/sentry-symfony/0.7/manifest.json
@@ -1,0 +1,13 @@
+{
+    "bundles": {
+        "Sentry\\SentryBundle\\SentryBundle": ["prod"]
+    },
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    },
+    "env": {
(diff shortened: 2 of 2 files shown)


GromNaN:
I've submitted this PR to fix the bundle: https://github.com/getsentry/sentry-symfony/pull/57

sentry/sentry-symfony/0.7/etc/packages/prod/sentry.yaml
```diff
@@ -0,0 +1,3 @@
+sentry:
+    dsn: "%env(SENTRY_DSN)%"
+    release: "%env(APP_RELEASE)%"
```
fabpot: How is it used by Sentry? The name seems too generic. Can't we just use `APP_ENV` instead?

GromNaN: This is the version of the application, not the environment. Should contain something like a git tag or commit hash.

fabpot:

ruudk:

ruudk:
@fabpot How come version 0.7 is used in the [Flex Server Test](https://symfony.sh/r/github.com/symfony/recipes-contrib/40) while there is also a 0.8.7 version? https://github.com/getsentry/sentry-symfony/releases

ruudk:
See https://github.com/symfony/recipes-contrib/pull/128 for a new PR that uses v0.8 that passes validation.

ruudk:
This can be closed now that #128 is merged

PR #44, opened 2017-05-27
(WIP) Add mykiwi/docker-* | Bug ?

mykiwi:
| Q             | A
| ------------- | ---
| License       | MIT


Pierstoval:
You must release at least one stable version for your package, else the bot refuses the PR

vonalbert:
Just a question: does make sense to include the recipes for versions of php < 7.1 since Flex will require at least PHP 7.1.3?
I mean all the docker-symfony-php5* and docker-symfony-php70* packages

mykiwi:
Good point.  
Symfony 3 still works with php 5.5 so I will keep php 5.6 & 7.0; but I think I will add an other container for composer with php 7.1.

vonalbert:
So if I'm using symfony 3.3.x (which requires php 5.5.9), as long I run all the composer commands from my host machine with php 7.1 (required only by flex ATM), everything should be fine for a development environment.

Anyway I'm not sure the dependencies could be resolved for a production env where php is 5.5, because of flex is not in required-dev section (and flex requires php 7.1).

However I cannot test it right away so I'm not 100% sure :)

mykiwi:
In fact, I  will remove 5.6 and 7.0 since 7.1 is in require when using flex.

PR #47, opened 2017-05-29
Add FOSRestBundle recipe

20uf:
| Q             | A
| ------------- | ---
| License       | MIT

About this bundle:
Is a popular bundle for the rapid development of RESTful applications and applications.
8.2 million downloads
https://github.com/FriendsOfSymfony/FOSRestBundle
@FriendsOfSymfony as the authors of the bundle

### DIFF
--- a/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
+++ b/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
@@ -15,13 +15,4 @@ fos_rest:
 #    format_listener:
 #        rules:
 #            - { path: ^/api, prefer_extension: true, fallback_format: json, priorities: [ json, html ] }
-#    body_listener:
-#        default_format:       null
-#        throw_exception_on_unsupported_content_type:  false
-#        decoders:
-#
-#            # Prototype
-#            name:                 ~
-#        array_normalizer:
-#            service:              null
-#            forms:                false
+#    body_listener: true
--- a/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
+++ b/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
@@ -1,5 +1,5 @@
 # Read the documentation: https://symfony.com/doc/master/bundles/FOSRestBundle/index.html
-fos_rest:
+fos_rest: ~
 # Uncomment to activate features
 #
 #    param_fetcher_listener:  true
@@ -15,4 +15,5 @@ fos_rest:
 #    format_listener:
 #        rules:
 #            - { path: ^/api, prefer_extension: true, fallback_format: json, priorities: [ json, html ] }
-#    body_listener: true
+#    body_listener:
+#        enabled:              true
--- a/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
+++ b/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
@@ -15,5 +15,3 @@ fos_rest: ~
 #    format_listener:
 #        rules:
 #            - { path: ^/api, prefer_extension: true, fallback_format: json, priorities: [ json, html ] }
-#    body_listener:
-#        enabled:              true
--- a/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
+++ b/friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
@@ -1,70 +1,20 @@
+# Read the documentation: https://symfony.com/doc/master/bundles/FOSRestBundle/index.html
 fos_rest:
-    disable_csrf_role:    null
-#    access_denied_listener:
+# Uncomment to activate features
 #
-#        # Prototype
-#        name:                 ~
-#    unauthorized_challenge:  null
-#    param_fetcher_listener:  false
-#    cache_dir:            '%kernel.cache_dir%/fos_rest'
-#    allowed_methods_listener:  false
-#    routing_loader:
-#        default_format:       null
-#        include_format:       true
-#    body_converter:
-#        enabled:              false
-#        validate:             false
-#        validation_errors_argument:  validationErrors
-#    service:
-#        router:               router
-#        templating:           templating
-#        serializer:           null
-#        view_handler:         fos_rest.view_handler.default
-#        exception_handler:    fos_rest.view.exception_wrapper_handler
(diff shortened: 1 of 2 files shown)


GuilhemN:
I don't think we should include so many config options, imo only the most important should be there.

20uf:
@GuilhemN What determines the most important? Do you have an example?

GuilhemN:
The one that are likely to be manually configured, so i would keep something minimal like:
```yml
fos_rest:
# Uncomment to activate features
#
#    param_fetcher_listener:  true
#    allowed_methods_listener:  true
#    routing_loader: true
#    view:
#        view_response_listener:  true
#    exception:
#        codes:
#            App\Exception\MyException: 403
#        messages:
#            App\Exception\MyException: Forbidden area.
#    format_listener:
#        rules:
#            - { path: ^/api, prefer_extension: true, fallback_format: json, priorities: [ json, html ] }
```

lsmith77:
puh .. hard to say .. I tend to use quite a few settings in my projects. certainly I always use `body_listener` as well. but I would also say .. better to start small.

GuilhemN:
Well indeed, I think we should add `body_listener`, removed by mistake :)

20uf:
@GuilhemN  @lsmith77 is it ok ?

friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
```diff
@@ -0,0 +1,27 @@
...
+#        view_response_listener:  true
+#    exception:
+#        codes:
+#            App\Exception\MyException: 403
+#        messages:
+#            App\Exception\MyException: Forbidden area.
+#    format_listener:
+#        rules:
+#            - { path: ^/api, prefer_extension: true, fallback_format: json, priorities: [ json, html ] }
+#    body_listener:
```
GuilhemN: Sub options should be removed imo.

GuilhemN:
I wonder if we shouldn't uncomment everything, we would lose a bit in perf (we can probably improve that in the bundle) but dx would be greatly improved.

friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
20uf: just : 

```
#    body_listener:
#        default_format:       null
#        throw_exception_on_unsupported_content_type:  false`
```
?

20uf:
Tell me what you decide :)

friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
GuilhemN: I don't think these are very usual so I would just keep `true`.

GuilhemN:
For me we should uncomment, at worse we'll still be able to revert if i'm the only thinking this way ☺

friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
20uf: Done

GuilhemN:
Can you also do https://github.com/symfony/recipes-contrib/pull/47#issuecomment-306040966 ? :)

20uf:
@GuilhemN  What is the request your link does not work :(

GuilhemN:
It was a link to my previous message:
> For me we should uncomment, at worse we'll still be able to revert if i'm the only thinking this way ☺

I think we should uncomment this default config.

friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
```diff
@@ -0,0 +1,19 @@
+# Read the documentation: https://symfony.com/doc/master/bundles/FOSRestBundle/index.html
+fos_rest: ~
+# Uncomment to activate features
+#
+#    param_fetcher_listener:  true
+#    allowed_methods_listener:  true
+#    routing_loader: true
```
GuilhemN: My mistake, this doesn't exist... should be removed.

friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
GuilhemN: It seems it's enabled by default (https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/Configuration.php#L264) so it should be entirely removed... sorry one more change :stuck_out_tongue:

20uf:
I deleted the listener from the body. Le code is it appropriate?

```
# Read the documentation: https://symfony.com/doc/master/bundles/FOSRestBundle/index.html
fos_rest: ~
# Uncomment to activate features
#
#    param_fetcher_listener:  true
#    allowed_methods_listener:  true
#    routing_loader: true
#    view:
#        view_response_listener:  true
#    exception:
#        codes:
#            App\Exception\MyException: 403
#        messages:
#            App\Exception\MyException: Forbidden area.
#    format_listener:
#        rules:
#            - { path: ^/api, prefer_extension: true, fallback_format: json, priorities: [ json, html ] }

```

20uf:
@GuilhemN @lsmith77 It's OK for you ? If you agree, you can approve this PR?

GuilhemN:
@20uf https://github.com/symfony/recipes-contrib/pull/47/files#r120013178 still need to be adressed and the file uncommented :)

20uf:
@GuilhemN It is done ! Tell me exactly what you want as configuration if it does not suit you, it will be simpler.

friendsofsymfony/rest-bundle/2.2/etc/packages/fos_rest.yaml
```diff
@@ -0,0 +1,17 @@
+# Read the documentation: https://symfony.com/doc/master/bundles/FOSRestBundle/index.html
+fos_rest: ~
+# Uncomment to activate features
```
fabpot: this can be removed

20uf: @fabpot want to tell you everything that is in comment is not worth it to be present?

fabpot: only this line which does not add any value IMHO (and the extra blank line as well)

20uf: It's done.

PR #48, opened 2017-05-29
Add StofDoctrineExtensionsBundle recipe

20uf:
| Q             | A
| ------------- | ---
| License       | MIT

About this bundle:
The most popular one for integration of the gedmo/doctrine-extensions.
6.5 million downloads
https://github.com/stof/StofDoctrineExtensionsBundle
@stof  as the author of the bundle

### DIFF
--- a/stof/doctrine-extensions-bundle/1.2/etc/packages/stof_doctrine_extensions.yaml
+++ b/stof/doctrine-extensions-bundle/1.2/etc/packages/stof_doctrine_extensions.yaml
@@ -1,0 +1,21 @@
+stof_doctrine_extensions:
+    default_locale: en_US
+
+    # Activate the extensions you want
+    orm:
+        default:
+            tree: true
+            timestampable: false # not needed: listeners are not enabled by default
+        other:
+            timestampable: true
+
+#    # Overriding the listeners¶
+#    class:
+#        tree:           MyBundle\TreeListener
+#        timestampable:  MyBundle\TimestampableListener
+#        blameable:      ~
+#        sluggable:      ~
+#        translatable:   ~
+#        loggable:       ~
+#        softdeleteable: ~
+#        uploadable:     ~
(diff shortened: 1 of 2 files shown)


stof/doctrine-extensions-bundle/1.2/etc/packages/stof_doctrine_extensions.yaml
```diff
@@ -0,0 +1,21 @@
+stof_doctrine_extensions:
+    default_locale: en_US
+
+    # Activate the extensions you want
+    orm:
+        default:
+            tree: true
+            timestampable: false # not needed: listeners are not enabled by default
+        other:
```
alOneh: this one will cause an issue because the `other` entity manager does not exist on common projects.
Just keep `default` section.

stof/doctrine-extensions-bundle/1.2/etc/packages/stof_doctrine_extensions.yaml
```diff
@@ -0,0 +1,21 @@
+stof_doctrine_extensions:
+    default_locale: en_US
+
+    # Activate the extensions you want
+    orm:
+        default:
+            tree: true
+            timestampable: false # not needed: listeners are not enabled by default
```
Pierstoval: I'd suggest adding most common extensions enabled and comment them, and add a small comment line explaining that nothing is enabled by default, so users just have to uncomment the lines to activate a specific extension

stof/doctrine-extensions-bundle/1.2/etc/packages/stof_doctrine_extensions.yaml
```diff
@@ -0,0 +1,21 @@
...
+        other:
+            timestampable: true
+
+#    # Overriding the listeners¶
```
Pierstoval: I don't know if it's very common to override listeners, is it? If not, then this is unnecessary because the user will definitely need the docs to override a listener, therefore defaulting the listeners classes is useless IMO

stof/doctrine-extensions-bundle/1.2/etc/packages/stof_doctrine_extensions.yaml
20uf: It is done, I simplified by adding the links of documentation which is too consequent to be expressed here.

PR #53, opened 2017-06-11
Add MobileDetectorBundle recipe

orions:
| Q             | A
| ------------- | ---
| License       | MIT

About this bundle:
The most popular one for detect mobile device.
+900K downloads
https://github.com/suncat2000/MobileDetectBundle
@suncat2000 as the author of the bundle

PR #39, opened 2017-05-17
Add NelmioApiDocBundle recipe

20uf:
| Q             | A
| ------------- | ---
| License       | MIT

About this bundle:
The most popular one for generate a decent documentation for your APIs.
6 million downloads
https://github.com/nelmio/NelmioApiDocBundle
@GuilhemN as the author of the bundle

PR #54, opened 2017-06-13
Add KnpSnappyBundle recipe

orions:
| Q             | A
| ------------- | ---
| License       | MIT

About this bundle:
Snappy is a PHP (5.3+) wrapper for the wkhtmltopdf conversion utility.
+2M downloads
https://github.com/KnpLabs/KnpSnappyBundle
@KnpLabs as the author of the bundle


stof:
@orions mentioning an organization does not send notifications to anyone, so it does not ask the author for a review 😄

knplabs/knp-snappy-bundle/1.4/etc/packages/knp_snappy.yaml
```diff
@@ -0,0 +1,10 @@
+# Read Documentation: https://github.com/KnpLabs/KnpSnappyBundle/blob/v1.4/README.markdown
+knp_snappy: ~
+#    temporary_folder: %kernel.cache_dir%/snappy
+#    pdf:
+#        enabled:    true
+#        options: []
+#    image:
+#        enabled:    true
+#        options: []
```
stof: putting all available options commented does not make any sense:

- recipes are not about providing documentation
- if the bundle adds new config settings, this will get unmaintainable:
    - the recipe will not automatically mention them
    - even if the recipe gets updated, your existing project will not get an updated version of the skeleton file
- everyone using your project will have to wonder all the time about why all the config settings are commented

20uf:
@stof It is nevertheless difficult to know who to tag when not explicitly stated who is / are the major authors.
(And frankly what is the use of the possibility of tag the group if does not receive notification.)

knplabs/knp-snappy-bundle/1.4/etc/packages/knp_snappy.yaml
20uf: @stof I totally agree. 
Some say that you need documentation for a better practice. [e.g](https://github.com/symfony/recipes-contrib/pull/39/files/853054988bd6de2a13835898a0de38019c4ab028#r117183291) 
This topic deserves more clarity in your documentation.

What do you think we should do?
```
# Read Documentation: https://github.com/KnpLabs/KnpSnappyBundle/blob/v1.4/README.markdown
knp_snappy: ~

```
Just that ?

orions: @stof i don't understand what i should provide as conf settings, all settings in this bundle have a default value. 
Should i provide an empty conf like @20uf propose or the most used settings like @fabpot propose in another PR https://github.com/symfony/recipes-contrib/pull/53

stof: Well, if 99% of people using the bundle don't need to configure anything, it is indeed best to rely on default values. But then, my point is that this file is useless in 99% of projects, and so does not need to be there at all (and so the recipe is useless after that, as it is exactly the same than the auto-created recipe if your remove this file).

PR #56, opened 2017-06-15
[WIP] basic recipe for kunstmaan bundles cms

Devolicious:
| Q             | A
| ------------- | ---
| License       | MIT

About this bundle:
- downloads:  55 349 
- https://github.com/Kunstmaan/KunstmaanBundlesCMS/

KunstmaanBundlesCMS is a fully equipped CMS and with this recipe we would like to easily intergrate into any project that needs a good and solid CMS. Now it's just a hassle with copy/pasting configs, routigs, ...

### DIFF
--- a/kunstmaan/bundles-cms/4.0/etc/packages/kunstmaan_liip_imagine.yaml
+++ b/kunstmaan/bundles-cms/4.0/etc/packages/kunstmaan_liip_imagine.yaml
@@ -1,0 +1,75 @@
+liip_imagine:
+    resolvers:
+        default:
+            web_path:
+                cache_prefix: uploads/cache
+    driver: imagick
+    data_loader: default
+    filter_sets:
+        optim:
+            quality: 85
+            format: jpg
+            filters:
+                strip: ~
+        optimjpg:
+            quality: 85
+            format: jpg
+            filters:
+                strip: ~
+        optimpng:
+            quality: 85
+            format: png
+            filters:
+                strip: ~
+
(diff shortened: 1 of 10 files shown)


kunstmaan/bundles-cms/4.0/etc/packages/kunstmaan_liip_imagine.yaml
```diff
@@ -0,0 +1,75 @@
+liip_imagine:
```
fabpot: This recipe defines too many things. It should not configure external libs/bundles like this one.

Devolicious: Hi @fabpot 

This is a configuration that comes from our `StandardEdition`.  It's not critical to be there and your completely right that it's not the responsibility of this recipe, but it's a nice to start with this already configured so you don't have to anymore. My question is, how would you go about `suggested configuration` for a dependency like LiipImagine? Because when I want to use the CMS, I don't want to go to each individual dependency to search how I should configure it. Do you have some kind of solution/suggestion for this?

Pierstoval: @Devolicious As you're relying on your Standard Edition, maybe you should create your own Flex skeleton?

Nyholm:
I see that this PR is in WIP. This is a kind reminder to finish this PR.

Nyholm:

PR #61, opened 2017-06-28
Rss atom bundle

alexdebril:
| Q             | A
| ------------- | ---
| License       | MIT

This bundle offers JSONFeed / Atom / RSS feeds' read and write support.


debril/rss-atom-bundle/4.0/package/rss_atom.yaml
```diff
@@ -0,0 +1,5 @@
+debril_rss_atom:
```
stof: this file must be inside the `etc` folder, otherwise it won't be copied.

stof:
Having no requirement on any Symfony package in the bundle looks weird to me. Are you sure about your composer requirements ?

stof:
Btw, your require-dev uses `<3.3` for all components, meanign you forbid testing with Symfony 3.3+, while Flex is only used for 3.3+. So your recipe is here to allow people to use the bundle in ways that your forbid your CI to test. Not good experience....

alexdebril:
Sorry I forgot about the require-dev issue I had some weeks ago, I'll fix this and submit another pull request later.

stof:
no need to close the PR. The recipe itself is still correct (maybe just mark the PR as WIP until you have tested the bundle with 3.3)

alexdebril:
Thank you for your advice.

OK, I fixed the version issue in require-dev and I added symfony/framework-bundle in the require section. The build is correct with Symfony 3.3, so I reopen the PR.

xabbuh:
I would be more explicit which Symfony components you rely on and not depend on the FrameworkBundle.

alexdebril:
I'm ok with that, however the StreamController depends on `Symfony\Bundle\FrameworkBundle\Controller\Controller`. so the `require` section will look like : 

```json
    "require": {
        "php": ">=5.6",
        "symfony/config": "~2.6|~3.0",
        "symfony/http-foundation": "~2.6|~3.0",
        "symfony/http-kernel": "~2.6|~3.0",
        "symfony/dependency-injection": "~2.6|~3.0",
        "symfony/framework-bundle": "~2.6|~3.0",
        "debril/feed-io": "~3.0"
    },
```
Is that good for you ?

stof:
@alexdebril this looks sensible to me (assuming 2.6 and 3.0 are really the lower bounds for 2.x and 3.x of course)

stof:
Btw, the section `Version 2.5 is a transitional version` in the readme of 4.x is not really useful.

debril/rss-atom-bundle/4.0/packages/rss_atom.yaml
```diff
@@ -0,0 +1,5 @@
+debril_rss_atom:
```
stof: you still haven't moved this file inside the `etc` folder, so it does not get copied

stof:
@fabpot could we have a validation for recipe ensure that all files of the recipe folder are either part of the recipe content (manifest, makefile and maybe others) or matched by one of the `copy-from-recipe` rules ?  this would avoid mistakes where the file is in the wrong place (which are hard to spot in the review. As you can see, I first approved the PR and then saw that the file was in the wrong location)

alexdebril:
@stof I moved `packages` under `etc` and just performed a test using  https://symfony.sh/r/github.com/symfony/recipes-contrib/61, it works. 

You're right about the "transitional version" it's not useful anymore in 4.x, I will remove it.

About the symfony/framework-bundle dependency, could you be more specific about the issue please ? I'm trying to figure out what's wrong with that and I really don't get it. Is it something with the compatibility range from 2.6 to 3.x, could it be too wide ?

alexdebril:
@stof, @fabpot I think it could be a good thing for developers to have a 'recipe-checker' to prevent invalid pull request. It could be working like this : 

```bash 
$ composer require --dev symfony/recipe-checker
$ ./vendor/bin/recipe-checker vendor/package
Checking ...
license [ok]
manifest.json [ok]
routing ending with .yaml [ok]
tabulations [ok]
...
```

Also it could make your validation process easier for you, with less invalid PRs to handle (like this one). And to force developers to use it, symfony/recipes-contrib could have a pre-push hook that runs the checker and refuses the push if the recipe is invalid.

fabpot:
@stof done

fabpot:
@alexdebril The validator runs as soon as you submit the pull request. It does exactly what you describe. Having it as a tool used by developers would be much more cumbersome as people would need to do an extra manual step, think about updating to the latest version of the tool, and I would have to support it, and probably more I don't think of right now.

The current process is much easier for everyone as the hook runs automatically and always uses the latest version. Just push your code and validation is triggered.  Errors are displayed in the diff and as a nice summary.

fabpot:
@alexdebril If you have finished here, you should remove the 'WIP' prefix on the title (which prevents merging the PR).

alexdebril:
@fabpot yes I made the requested changes, I remove the [WIP] prefix. Thank you for your explanations about the validator.

PR #65, opened 2017-07-30
Add GeoSocio Http Serializer Bundle

davidbarratt:
| Q             | A
| ------------- | ---
| License       | MIT


davidbarratt:
I updated the license in `1.0.5` but now it says the version is not valid. :(
https://packagist.org/packages/geosocio/http-serializer-bundle

javiereguiluz:
Version 1.0.5 is valid according to Packagist: https://packagist.org/packages/geosocio/http-serializer-bundle  Maybe the problem is that it has just been released and we need to wait a bit?

fabpot:
Version 1.0.5 is indeed not valid. It should be 1.0 I suppose.

fabpot:
I've also just updated the validator to make the error message more explicit (you can see the new error message on this PR).

PR #70, opened 2017-08-06
Add recipes for endroid libraries

endroid:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/endroid/asset/1.0/config/packages/endroid_asset.yaml
+++ b/endroid/asset/1.0/config/packages/endroid_asset.yaml
@@ -1,0 +1,2 @@
+endroid_asset:
+    public_dir: '%kernel.project_dir%/public'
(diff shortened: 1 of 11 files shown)


stof:
@endroid to get a better experience when using flex, you should split bundles into separate packages than the library (with a dependency on the library of course). This gets you several benefits:

- the bundle can be autoregistered in the kernel by flex instead of forcing to use manual registration (not that handy in flex projects), as the package can then have the type `symfony-bundle`.
- composer requirements can be used to tell composer about which versions of Symfony are supported, allowing you to drop support for older versions of Symfony without breaking everything (if you don't have requirements, Composer will consider that any version of Symfony is fine with your bundle, instead of using the older version of the bundle supporting the old Symfony version being used)

endroid:
Hi @stof thank you for you input. I am aware that separating the library from the bundles would be better. The only reason I merged them at a certain point is that I did not have enough time to maintain too many repositories and dependencies so at a moment I decided to keep them together to keep things easier for myself though I really rather kept them separated from an architectural point of view. Personally I experience a lot less mentions about people having trouble installing the library so I can focus more on what I want to build.

Last week I set the package type to "symfony-bundle" to work around the first issue, which is not 100% correct but I am now able to auto-register the bundles inside the library and register an optional demo bundle included in the library for the dev environment.

PR #71, opened 2017-08-22
Add recipe for league/tactician-bundle

kejwmen:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/thephpleague/tactician-bundle

Integration with command bus implementation.


league/tactician-bundle/0.4/config/packages/league_tactician.yaml
```diff
@@ -0,0 +1,17 @@
+# Library documentation: http://tactician.thephpleague.com/
+# Bundle documentation: https://github.com/thephpleague/tactician-bundle/blob/v0.4.1/README.md
+tactician:
+    default_bus: default
+    commandbus:
+        default:
+            middleware:
+                - tactician.middleware.command_handler
+# Example of separate command/query buses with transactional command bus
+#    query:
```
ogizanagi: Indentation looks weird. Should be 4 more sapces to me.

kejwmen: thanks, good catch!

rosstuck:
Hi @kejwmen! Thanks for putting this together. :)

I would recommend leaving out the query bus example, it's not really a use case we do much with and the bulk of users never touch it.

It might be nice to have commented out versions of common middleware here, such as the doctrine or validator ones. I'm on mobile right now but can send a PR later this week if you like.

Also, you can leave out the default_bus line since you're using the default bus by, well, default :)

I'm going to tag a big new release of Tactician bundle this week, maybe tomorrow. Shouldn't break anything here but just FYI.

Thanks again!

chalasr:
I suggest to not add a recipe for `0.4` and start with `1.0` instead, it is more suited for symfony/dependency-injection 3.3+.

kejwmen:
> I would recommend leaving out the query bus example, it's not really a use case we do much with and the bulk of users never touch it.

ok, I agree. Recipe should help end users with integration, should not present patterns.

> It might be nice to have commented out versions of common middleware here, such as the doctrine or validator ones.

:+1: 

> Also, you can leave out the default_bus line since you're using the default bus by, well, default :)

Ok, this is too verbose :)

> I'm going to tag a big new release of Tactician bundle this week, maybe tomorrow. Shouldn't break anything here but just FYI.

As @chalasr suggested, I will make some changes and introduce it as a recipe for `1.0`.

league/tactician-bundle/1.0/config/packages/league_tactician.yaml
```diff
@@ -0,0 +1,15 @@
...
+    commandbus:
+        default:
+            middleware:
+# Security middleware - https://github.com/thephpleague/tactician-bundle#security-middleware-tacticianmiddlewaresecurity
+#                - tactician.middleware.security
+# Validator middleware - https://github.com/thephpleague/tactician-bundle/tree/v1.0#validator-middleware-tacticianmiddlewarevalidator
+#                - tactician.middleware.validator
+# Locking middleware - http://tactician.thephpleague.com/plugins/locking-middleware/
+#                - tactician.middleware.locking
+# Doctrine transactional middleware, requires additional package - https://github.com/thephpleague/tactician-doctrine
```
rosstuck: Nice touch linking to the docs with the example!

rosstuck:
Any update here? This looks okay to me. :)

fabpot:
As long as there is `WIP` in the title, nothing will happen.

kejwmen:
sorry @rosstuck, I forgot about WIP in the title

kejwmen:
@rosstuck looks like you have to [accept PR](https://github.com/symfony/recipes-contrib#contributing) to get it automatically merged

PR #74, opened 2017-09-03
Remove asset and set required parameter

endroid:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/endroid/asset/1.0/manifest.json
+++ b/endroid/asset/1.0/manifest.json
@@ -1,8 +1,0 @@
-{
-    "bundles": {
-        "Endroid\\Asset\\Bundle\\AssetBundle\\EndroidAssetBundle": ["all"]
-    },
-    "copy-from-recipe": {
-        "config/": "%CONFIG_DIR%/"
-    }
-}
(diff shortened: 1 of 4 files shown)


endroid/asset/1.0/manifest.json
```diff
@@ -1,8 +0,0 @@
-{
-    "bundles": {
-        "Endroid\\Asset\\Bundle\\AssetBundle\\EndroidAssetBundle": ["all"]
-    },
```
yceruto: At least this section is required to register the bundle automatically.

endroid: Hi @yceruto endroid/asset does not exist anymore because it provided functionality that is now provided by Symfony itself. That means the manifest can also be removed completely. Otherwise it would only refer to a non-existent package.

yceruto: Means the recipe can be completely eliminated?

endroid: Yes, correctly, no need for it anymore. Thank you.

PR #76, opened 2017-09-07
Add PredictionIO library

endroid:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/endroid/prediction-io/2.3/manifest.json
+++ b/endroid/prediction-io/2.3/manifest.json
@@ -1,7 +1,7 @@
 {
     "bundles": {
         "Endroid\\PredictionIO\\Bundle\\PredictionIOBundle\\EndroidPredictionIOBundle": ["all"],
-        "Endroid\\PredictionIO\\Bundle\\PredictionIODemoBundle\\EndroidPredictionIODemoBundle": ["all"]
+        "Endroid\\PredictionIO\\Bundle\\PredictionIODemoBundle\\EndroidPredictionIODemoBundle": ["dev"]
     },
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/"


endroid/prediction-io/2.3/manifest.json
```diff
@@ -0,0 +1,12 @@
...
+  "bundles": {
+    "Endroid\\PredictionIO\\Bundle\\PredictionIOBundle\\EndroidPredictionIOBundle": ["all"],
+    "Endroid\\PredictionIO\\Bundle\\PredictionIODemoBundle\\EndroidPredictionIODemoBundle": ["all"]
+  },
+  "copy-from-recipe": {
+    "config/": "%CONFIG_DIR%/"
+  },
+  "env": {
+    "PREDICTION_IO_ACCESS_KEY": ""
+  }
```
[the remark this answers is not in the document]
chalasr: @symfony-bot Once per file should be enough for this one ;)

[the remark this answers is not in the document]
fabpot: Not sure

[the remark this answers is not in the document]
stof: @fabpot the bot may check whether the indentation is consistent in the file (using 2 spaces per level everywhere), and use a single comment in this case.

endroid/data-sanitize/1.3/config/packages/dev/endroid_data_sanitize.yaml
```diff
@@ -0,0 +1,14 @@
+endroid_data_sanitize:
+    entities:
+        project:
+            class: Endroid\DataSanitize\Bundle\DataSanitizeDemoBundle\Entity\Project
```
stof: I don't think that the demo bundle should be configured when creating a new project

endroid/prediction-io/2.3/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Endroid\\PredictionIO\\Bundle\\PredictionIOBundle\\EndroidPredictionIOBundle": ["all"],
+        "Endroid\\PredictionIO\\Bundle\\PredictionIODemoBundle\\EndroidPredictionIODemoBundle": ["all"]
```
stof: the demo bundle should not be registered when installing the package in a project. It does not make sense to have it in actual projects.

stof: btw, I would even suggest splitting the demo into a totally separate package, so that it does not even need to be installed.

endroid/data-sanitize/1.3/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "bundles": {
+        "Endroid\\DataSanitize\\Bundle\\DataSanitizeBundle\\EndroidDataSanitizeBundle": ["all"],
+        "Endroid\\DataSanitize\\Bundle\\DataSanitizeDemoBundle\\EndroidDataSanitizeBundle": ["dev"]
```
stof: a demo bundle should not be registered automatically in projects wanting to use the feature.

endroid/data-sanitize/1.3/config/packages/dev/endroid_data_sanitize.yaml
endroid: That's true, ideally not

fabpot:
@endroid You should rebase to get rid of the merge commit.

PR #82, opened 2017-09-15
Add ramsey/uuid-doctrine

DerStoffel:
| Q             | A
| ------------- | ---
| License       | MIT


fabpot:
Version 3.3 does not exist, do you mean version 1.3 or 1.4 perhaps?

DerStoffel:
latest on packagist is 1.4.1
I assume that's the one to chose then.

DerStoffel:
1.4 it is apparently ;)
How is this version scheme working?

fabpot:
You should use the oldest version for which the recipe works. It will then apply the recipe to this version and up.

fabpot:
Have you tried with several types defined in different files. I fear that they will override each other, which is not what we want. If I recall correctly, there is no merge between config files. Can you check?

DerStoffel:
I will check, that would be a shame though, since this library is created for this purpose especially

DerStoffel:
Behaviour is as follows:

A.yaml containing: doctrine.dbal.types.uuid: XXX
B.yaml containing: doctrine.dbal.types.uuid: YYY
=> doctrine.dbal.types.uuid: YYY

A.yaml containing: doctrine.dbal.types.uuid: XXX
B.yaml containing: doctrine.dbal.types.uuid2: YYY
=> doctrine.dbal.types.uuid: XXX
=> doctrine.dbal.types.uuid: YYY

So it only overrides if specified this exact type being specified before.
Imho this is okay, but that's up to you.

What do you propose?
That would possibly end in a bunch of things beeing configured as <libname>_varname
I doubt, thats a goal to aim for.

fabpot:
@DerStoffel Great! Exactly the behavior we need.

ramsey/uuid-doctrine/1.3/config/packages/ramsey_uuid_doctrine.yaml
```diff
@@ -0,0 +1,4 @@
+doctrine:
```
xabbuh: I see one issue here: `ramsey/uuid-doctrine` does not require `doctrine/doctrine-bundle`. So if one installs the UUID package without requiring DoctrineBundle too, this part will lead to errors. Are we okay with that?

DerStoffel: but why else would I use it in a symfony project?

Issue #75, opened 2017-09-04
[doctrine/mongodb-odm-bundle] Impossible to set db name via MONGODB_URL


BoShurik:
MONGODB_URL=mongodb://localhost:27017/symfony
but uses `default` db

stof:
Well, please show us the configuration you have in your project for the doctrine_odm bundle

BoShurik:
Same as in recipe:
```
doctrine_mongodb:
    auto_generate_proxy_classes: '%kernel.debug%'
    auto_generate_hydrator_classes: '%kernel.debug%'
    connections:
        default:
            server: '%env(MONGODB_URL)%'
            options: {}
    document_managers:
        default:
            auto_mapping: true
            mappings:
                App:
                    is_bundle: false
                    type: annotation
                    dir: '%kernel.project_dir%/src/Document'
                    prefix: App\Document\
                    alias: App
```

PR #86, opened 2017-09-22
Set mongodb-odm-bundle default database

BoShurik:
| Q             | A
| ------------- | ---
| License       | MIT

Fix https://github.com/symfony/recipes-contrib/issues/75

### DIFF
--- a/doctrine/mongodb-odm-bundle/3.3/manifest.json
+++ b/doctrine/mongodb-odm-bundle/3.3/manifest.json
@@ -7,6 +7,7 @@
         "src/": "%SRC_DIR%/"
     },
     "env": {
-        "MONGODB_URL": "mongodb://localhost:27017/symfony"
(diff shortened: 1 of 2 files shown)


doctrine/mongodb-odm-bundle/3.3/manifest.json
```diff
@@ -7,6 +7,7 @@
         "src/": "%SRC_DIR%/"
     },
     "env": {
-        "MONGODB_URL": "mongodb://localhost:27017/symfony"
+        "MONGODB_URL": "mongodb://localhost:27017",
+        "MONGODB_DB": "symfony"
```
fabpot: So, it means that the bundle does not parse the DSN for database name? Wouldn't it be better to "fix" that in the bundle? Or perhaps that does not make sense for MongoDB?

fabpot:

PR #88, opened 2017-09-22
Add Bukashk0zzz FilterBundle

Bukashk0zzz:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/bukashk0zzz/filter-bundle/2.0/config/packages/bukashk0zzz_filter.yaml
+++ b/bukashk0zzz/filter-bundle/2.0/config/packages/bukashk0zzz_filter.yaml
@@ -1,0 +1,3 @@
+bukashk0zzz_filter:
+    # Enable if you need auto filtering form data before constraint(Validation) check
+    auto_filter_forms: false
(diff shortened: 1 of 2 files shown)


bukashk0zzz/filter-bundle/2.0/config/packages/bukashk0zzz_filter.yaml
```diff
@@ -0,0 +1,3 @@
+bukashk0zzz_filter:
+    # Enable if you need auto filtering form data before constrain(Validation) check
```
stof: isn't there a typo here ?

Bukashk0zzz: Checked twice. I think no. All correct.

fabpot: Shouldn't it be `constraint` instead of `constrain`?

Bukashk0zzz: My bad. Updated.

PR #96, opened 2017-09-30
Added leogout-seo bundle

leogout:
| Q             | A
| ------------- | ---
| License       | MIT

As the config is essentially keywords and titles, I didn't added any default values or ENV parameters. I've also referenced the bundle's config to make it simpler for the user.


fabpot:

PR #98, opened 2017-10-03
Add Ad3n Ratchet Bundle

ad3n:
| Q             | A
| ------------- | ---
| License       | MIT

Add ad3n ratchet bundle


ad3n/ratchet-bundle/0.4/config/packages/ihsan_ratchet.yaml
```diff
@@ -0,0 +1,2 @@
+ihsan_ratchet:
+    web_socket_port: "%env(WEB_SOCKET_PORT)%"
```
stof: the manifest should also declare the env variable so that it is added in `.env.dist`

ad3n: done?

ad3n: i just see from SNC Redis, it's ok for you?

PR #102, opened 2017-10-17
Add secit-pl/imap-bundle recipe

secit-pl:
| Q             | A
| ------------- | ---
| License       | MIT


secit-pl/imap-bundle/1.0/config/packages/imap.yaml
```diff
@@ -0,0 +1,2 @@
+imap:
+    connections: []
```
stof: does this empty array actually make sense ?

secit-pl: As you can reed here https://github.com/secit-pl/imap-bundle (Configuration section) this bundle require from user to setup the connections in the configuration file.

If the `imap: connections:` is not set in the configuration Symfony will throw an error that this value is required. In normal situation it's not a problem and user will know what to do. While using flex throwing the exceptions while doing `composer req` will break the installation process. 

```
!!    [Symfony\Component\Config\Definition\Exception\InvalidConfigurationException]
!!    The child node "connections" at path "imap" must be configured.
!!

Installation failed, reverting ./composer.json to its original content.
```

That's why the default (empty array) value is required to finnish the installation process with success.

fabpot:

PR #104, opened 2017-10-19
Adding php-cache's adapter-bundle and cache-bundle

aequasi:
| Q             | A
| ------------- | ---
| License       | MIT

@Nyholm

### DIFF
--- a/cache/cache-bundle/1.0/config/packages/cache.yaml
+++ b/cache/cache-bundle/1.0/config/packages/cache.yaml
@@ -1,0 +1,33 @@
+cache:
+    session:
+        enabled: "%env(bool:SESSION_CACHE_ENABLED)%"
+        service_id: "cache.provider.default"
+        use_tagging: true
+        ttl: 7200
+    router:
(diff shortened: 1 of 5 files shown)


cache/cache-bundle/1.0/config/packages/cache.yaml
```diff
@@ -0,0 +1,33 @@
...
+    validation:
+        enabled: "%env(bool:VALIDATION_CACHE_ENABLED)%"
+        service_id: "cache.provider.default"
+        use_tagging: true
+
+framework:
+    annotations:
+        cache: 'cache.service.annotation'
+    validation:
+        cache: 'cache.service.validation'
```
stof: I'm not sure silently replacing the Symfony builtin cache is good.

aequasi: How would you suggest we replace it?

aequasi:
@stof had a talk with @nicolas-grekas about cache-bundle. We are going to actually remove the validator, annotation, and serializer caches from `cache-bundle`

fabpot:

PR #105, opened 2017-10-20
Add mediamonks/rest-api-bundle

slootjes:
| Q             | A
| ------------- | ---
| License       | MIT

I'm the creator of both the bundle and the package it uses.

### DIFF
--- a/mediamonks/rest-api-bundle/3.0/config/packages/mediamonks_rest_api.yaml
+++ b/mediamonks/rest-api-bundle/3.0/config/packages/mediamonks_rest_api.yaml
@@ -1,15 +1,20 @@
 mediamonks_rest_api:
     request_matcher:
         path: /api
+
     # manually force debug to be enabled or disabled
     #debug: %kernel.debug%
+
     # set the post message origin when needed
     #post_message_origin:
+
     # use regular expressions instead of a path to enable this bundle
     #request_matcher:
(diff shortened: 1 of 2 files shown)


mediamonks/rest-api-bundle/3.0/config/packages/mediamonks_rest_api.yaml
```diff
@@ -0,0 +1,20 @@
+mediamonks_rest_api:
+    request_matcher:
+        path: /api
+
+    # manually force debug to be enabled or disabled
+    #debug: %kernel.debug%
```
stof: you should not put all options commented. The recipe is not meant to replace your documentation, as this file won't be updated with new options when the bundle gets updated (showing the wrong options)

mediamonks/rest-api-bundle/3.0/config/packages/mediamonks_rest_api.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+    # set the post message origin when needed
+    #post_message_origin:
+
+    # use regular expressions instead of a path to enable this bundle
+    #request_matcher:
```
stof: this gets even worse here. Uncommenting these lines would make the Yaml invalid due to duplicate keys.

slootjes:
Makes sense, thanks for the feedback.

PR #112, opened 2017-10-21
Adding PHP translation bundle

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

I have two questions:
* Check my paths for `dir` and `output_dir`, are they correct according to sf4 directory structure? 
* Am I correct by putting "dev only" config in a subfolder called "dev" next to the config shared on all environments?


php-translation/symfony-bundle/0.4/config/packages/dev/php_translation.yaml
```diff
@@ -0,0 +1,3 @@
+translation:
+    symfony_profiler: # must be placed in config_dev.yml
```
fabpot: Using `packages/dev` is correct 👍

php-translation/symfony-bundle/0.4/config/packages/php_translation.yaml
```diff
@@ -0,0 +1,14 @@
+translation:
+    locales: ["en", "sv", "fr"]
```
fabpot: I would only use `["en"]` here.

php-translation/symfony-bundle/0.4/config/packages/php_translation.yaml
```diff
@@ -0,0 +1,14 @@
...
+    webui:
+        enabled: true
+    edit_in_place:
+        enabled: false
+        config_name: app
+        activator: php_translation.edit_in_place.activator
+    configs:
+        app:
+            dirs: ["%kernel.project_dir%/templates", "%kernel.root_dir%/src"]
+            output_dir: "%kernel.project_dir%/app/Resources/translations"
```
fabpot: This one is not correct, should be `%kernel.project_dir%/translations`

php-translation/symfony-bundle/0.4/config/packages/php_translation.yaml
```diff
@@ -0,0 +1,14 @@
...
+            excluded_names: ["*TestCase.php", "*Test.php"]
+            excluded_dirs: [cache, data, logs]
```
fabpot: What does it refer to? `var/`? Why?

php-translation/symfony-bundle/0.4/config/packages/php_translation.yaml
```diff
@@ -0,0 +1,14 @@
+translation:
+    locales: ["en", "sv", "fr"]
+    webui:
+        enabled: true
+    edit_in_place:
+        enabled: false
+        config_name: app
+        activator: php_translation.edit_in_place.activator
```
javiereguiluz: Could we remove this `activator` option and rely on a default value? It's super redundant that the value of the `activator` option for `edit_in_place` option of PGP translation is `php_translation.edit_in_place.activator`.

fabpot:
@Nyholm Any news on this one?

php-translation/symfony-bundle/0.4/config/packages/php_translation.yaml
Nyholm: These values will be given to `Finder::notPath()`. Ie, when extracting the `src` or `vendor` for translations we should never go in to folders named "cache", "data" or "logs".

php-translation/symfony-bundle/0.4/config/packages/php_translation.yaml
```diff
@@ -0,0 +1,13 @@
+translation:
+    locales: ["en"]
+    webui:
+        enabled: true
```
stof: shouldn't this be enabled only in dev, as you import the routing only in dev ?

php-translation/symfony-bundle/0.4/config/packages/dev/php_translation.yaml
stof: I suggest removing this `must be placed in config_dev.yml` comment. It is more confusing than anything else here

php-translation/symfony-bundle/0.4/config/packages/php_translation.yaml
Nyholm: You are correct. Thank you

Nyholm:
Thank you for the reviews

PR #114, opened 2017-10-22
Add PHP-CS-Fixer recipe

jdecool:
| Q             | A
| ------------- | ---
| License       | MIT

Add recipe for the [PHP-CS-Fixer](http://github.com/friendsofphp/php-cs-fixer) lib.

### DIFF
--- a/friendsofphp/php-cs-fixer/2.7/.php_cs.dist
+++ b/friendsofphp/php-cs-fixer/2.7/.php_cs.dist
@@ -1,0 +1,71 @@
+<?php
+
+$header = <<<'EOF'
+This file is part of PHP CS Fixer.
+
+(c) Fabien Potencier <fabien@symfony.com>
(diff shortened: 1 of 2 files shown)


fabpot:
Do we really need to have such a long config? /cc @keradus

jdecool:
It's the default file that use in the project.

But maybe, we can reduce it to a minimal version that check PSR-2 convention and allow user to add more configuration.

keradus:
@fabpot not, for sure not!

> It's the default file that use in the project.

@jdecool this is not good approach to follow ;)
- we have very tweaked ruleset, yet we shall ask ppl to follow well-established one instead
- risky by default is not good idea
- workaround for fabbot.io is not needed at all for all projects except dev line of PHP CS Fixer itself

IMO, how it shall look like: https://github.com/symfony/recipes/pull/220

fabpot:
closing in favor of symfony/recipes#220

PR #115, opened 2017-10-23
Added Neo4j default configuration

### LINKED ISSUE #109: Add Neo4jBundle
Ping @xavismeh, Do you want to do this or should I?

xavismeh:
| Q             | A
| ------------- | ---
| License       | MIT

Closes #109


fabpot:
Is it possible to pass a URL/DSN instead of individual en vars? That's what we've done with Doctrine/Switfmailer. It would be more consistent if possible.

xavismeh:
Hello @fabpot. We're going to have a look to use a dsn for the reason you've mentioned. Still, we need to investigate on this as it requires to perform some changes into the bundle itself. Thx for your feedback!

neo4j/neo4j-bundle/0.3/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+    "bundles": {
+        "Neo4j\\Neo4jBundle\\": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "NEO4J_DEFAULT_CONNECTION_SCHEME": "bolt",
```
javiereguiluz: Do we need `_DEFAULT_` to be part of env var names? We usually don't do that in other recipes.

neo4j/neo4j-bundle/0.3/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+    "bundles": {
+        "Neo4j\\Neo4jBundle\\": ["all"]
```
gregurco: is namespace correct here? Probably it should be `Neo4j\\Neo4jBundle\\Neo4jBundle`

fabpot:
@xavismeh any news on this PR?

neo4j/neo4j-bundle/0.3/manifest.json
xavismeh: Hello @javiereguiluz, thanks for your feedback. If the `_DEFAULT_CONNECTION_` prefix is not something common, I think that providing a such information would be helpful, don't you think?

neo4j/neo4j-bundle/0.3/manifest.json
xavismeh: It is :)

neo4j/neo4j-bundle/0.3/manifest.json
Nyholm: I think it is needed. The bundle supports multiple connections.

Nyholm:
The bundle nor the underlying library supports DSN if I recall correctly. 

I think we should focus on this recipe with a bunch of parameters and the @xavismeh and I need to make updates to the Bundle so it supports DSN. Does that sound fair?

xavismeh:
It took me a while to check but yeah, @Nyholm is right: there are some modifications to work on in order to allow a DSN

Nyholm:
DSN is not supported in the bundle. @xavismeh, could you update this PR?

Nyholm:
Any updates on this PR?

TSchuermans:
Are there any updates on this? :)

PR #122, opened 2017-10-27
 Add Orbitale EasyImpressBundle

Pierstoval:
| Q             | A
| ------------- | ---
| License       | MIT

=> https://github.com/Orbitale/EasyImpressBundle

### DIFF
--- a/orbitale/easyimpress-bundle/1.0/config/presentations/home.yaml
+++ b/orbitale/easyimpress-bundle/1.0/config/presentations/home.yaml
@@ -1,0 +1,15 @@
+# Read the documentation https://github.com/Orbitale/EasyImpressBundle#yml-reference for more details about yaml options
+data:
+    # In seconds
+    transition-duration: 1000
+
+slides:
+    -
+        content: 'Hello world!'
+        data:
+            x: 0
+            y: 0
+            z: 1000
+            rotate-x: 0
+            rotate-y: 0
+            rotate-z: 0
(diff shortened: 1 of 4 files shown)


orbitale/easyimpress-bundle/1.0/config/presentations/home.yaml
```diff
@@ -0,0 +1,16 @@
+# Documentation for yml reference options:
+# https://github.com/Orbitale/EasyImpressBundle#yml-reference
```
fabpot: Can you fix this comment so that it looks like similar ones for other packages?

orbitale/easyimpress-bundle/1.0/post-install.txt
```diff
@@ -0,0 +1,4 @@
+<bg=blue;fg=white> Next for EasyImpressBundle </>
+    Create your presentations in yaml in the <comment>config/presentations/</comment> directory.
+
+    Check documentation for yaml reference options: <comment>https://github.com/Orbitale/EasyImpressBundle#yml-reference</comment>.
```
fabpot: This file should be removed. People probably come from the docs anyway. We try to limit the usage of the post install messages to when it's really important, like having to run a command before getting started.

orbitale/easyimpress-bundle/1.0/config/presentations/home.yaml
Pierstoval: Fixed

orbitale/easyimpress-bundle/1.0/post-install.txt
Pierstoval: Removed

Issue #125, opened 2017-10-29
[bug] Flex Server: Unable to parse git ls-tree output


sroze:
I'm not sure where to report this. While working on a contrib pull-request (#124) I am using the `SYMFONY_ENDPOINT` environment variable to use a given "pull-request environment". It was working impressively well yesterday, but this morning, Composer complained:

```
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
  - Installing sroze/openwhisk-bundle (0.2.0): Loading from cache
Writing lock file
Generating autoload files

Installation failed, reverting ./composer.json to its original content.


  [Composer\Downloader\TransportException]
  The "https://symfony.sh/r/github.com/symfony/recipes-contrib/124/p/sroze,openwhisk-bundle,i0.2.0,1509194154" file could not be downloaded (HTTP/1.1 500 Internal Server Error)
```

The `symfony.sh` URL returns the following:
```
< HTTP/1.1 500 Internal Server Error
< Content-Length: 34
< Content-Type: text/plain; charset=utf-8
< Date: Sun, 29 Oct 2017 09:33:01 GMT
< Vary: Origin
< X-Correlation-Id: khhctnfqwrbagxbu4budujoc
< X-Sensiocloud-Cache: MISS
< X-Sensiocloud-Cluster: egjbl6qhu5lse-master-7rqtwti
< X-Sensiocloud-Processor: 7zwai3z77ggmnxkao6dhwzsjxq
< X-Sensiocloud-Router: 3hjnjz4dfxtohh4gibk6eanr7y

unable to parse git ls-tree output
```

sroze:
I was hoping amend+push on the branch would "refresh" the build and make this error disappear but it did not work.

One for you @tucksaun? Or @fabpot?

Pierstoval:
In your composer cache directory (most of the time, `$HOME/.cache/Composer`) you can find a `repo` dir that contains a directory corresponding to your `SYMFONY_ENDPOINT`.

Have you tried clearing it?

sroze:
@Pierstoval just to clarify: the "unable to parse git ls-tree output" error comes from the Flex Server (see the HTTP request).

fabpot:
Whoops, my bad. Fixed now.

PR #124, opened 2017-10-28
[WIP] Add the openwhisk-bundle recipe

sroze:
| Q             | A
| ------------- | ---
| License       | MIT

This recipe will automatically create the Openwhisk entry point for the action (or "cloud function").

PR #126, opened 2017-10-29
Add the sroze/openwhisk-bundle recipe

sroze:
| Q             | A
| ------------- | ---
| License       | MIT

This recipe will automatically create the Openwhisk entry point for the action (or "cloud function").
This makes Symfony deployable to Openwhisk in almost one `composer req`.

Replaces #124 because of the issue #125.

### DIFF
--- a/sroze/openwhisk-bundle/0.2/manifest.json
+++ b/sroze/openwhisk-bundle/0.2/manifest.json
@@ -1,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "index.php": "index.php"
+    }
+}
--- a/sroze/openwhisk-bundle/0.2/manifest.json
+++ b/sroze/openwhisk-bundle/0.2/manifest.json
@@ -1,5 +1,9 @@
 {
     "copy-from-recipe": {
         "index.php": "index.php"
-    }
+    },
+    "gitignore": [
+        "/.serverless",
+        "/node_modules"
+    ]
 }
--- a/sroze/openwhisk-bundle/0.2/manifest.json
+++ b/sroze/openwhisk-bundle/0.2/manifest.json
@@ -1,9 +1,5 @@
 {
     "copy-from-recipe": {
         "index.php": "index.php"
-    },
-    "gitignore": [
-        "/.serverless",
-        "/node_modules"
-    ]
+    }
 }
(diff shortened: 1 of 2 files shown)


sroze:
It actually has the same #125 issue 😅

fabpot:
@sroze You can test it now (it works for me after my hotfix :)). Tell me if everything works fine for you.

fabpot:
Not sure if this is relevant to add this here. It looks more like an example than something that can be used as is. A recipe should not replace documentation. Can you explain a bit more about how one would us it in the context of an app managed by Flex?

sroze:
@fabpot super, it worked, thanks 👍 

This is the front-controller of your application if running it within an openwhisk function. (it replaces the `public/index.php` file). Therefore, when you `req sroze/openwhisk-bundle` one your Symfony application, it makes it directly "deployable" on such platform.

fabpot:
I understand that, but then, the `public` directory is not needed anymore, right?

fabpot:
@sroze For fun: https://coinhive.com/ :)

sroze:
Haha, I love it. Even easier than I thought to integrate with the Debug Toolbar 😜 The "proof of work" captcha is a great idea, if we make it a "proof of issue" ... 💸 

You're right. But the `public` directory is still useful when working on the application from the developer's laptop, so I'd keep it. By the way, this `index.php` is at the root folder because it's a limitation of the Openwhisk PHP integration.

sroze/openwhisk-bundle/0.2/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "index.php": "index.php"
+    },
+    "gitignore": [
+        "/.serverless",
+        "/node_modules"
```
fabpot: This one is not related to the bundle, but more generic. Is it something we want by default in the `symfony/framework-bundle` recipe?

sroze: Actually.. none of these are related to openwhisk directly either 🤔 I'll remove them from this PR. 

`/node_modules/` is already in the [`webpack-encore-pack` manifest](https://github.com/symfony/recipes/blob/master/symfony/webpack-encore-pack/1.0/manifest.json#L9) so I guess it's alright for us.

sroze:
I really like this symfony-flex-server 👍

PR #127, opened 2017-10-30
OverblogGraphiQLBundle: Added recipe for ^0.1.0

renatomefi:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/overblog/graphiql-bundle/0.1/post-install.txt
+++ b/overblog/graphiql-bundle/0.1/post-install.txt
@@ -1,0 +1,2 @@
+<bg=blue;fg=white> Next for OverblogGraphiQLBundle </>
+    1. Browse to <comment>/graphiql</>
--- a/overblog/graphiql-bundle/0.1/config/routes/dev/graphiql.yaml
+++ b/overblog/graphiql-bundle/0.1/config/routes/dev/graphiql.yaml
@@ -1,0 +1,2 @@
+overblog_graphiql:
+    resource: "@OverblogGraphiQLBundle/Resources/config/routing.xml"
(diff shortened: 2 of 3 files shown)


renatomefi:
@symfony-flex-server review again please

renatomefi:
@symfony-flex-server review again please

overblog/graphiql-bundle/0.1/post-install.txt
```diff
@@ -0,0 +1,3 @@
+<bg=blue;fg=white> Next for OverblogGraphiQLBundle </>
+    1. The bundle is auto configured now, if you need special/custom configuration read the documentation <comment>https://github.com/overblog/GraphiQLBundle/blob/master/Resources/doc/definitions/index.md</>
```
fabpot: This first comment can be removed as we don't link to the docs in post install information. What you can do instead is add the link as a comment in the configuration file (see other recipes to use the same semantic).

stof: but there is no such config file (as the default settings are meaningful).

anyway, I don't think we need to link to the doc at all from Flex here (as you only need for more advanced cases)

renatomefi: Exactly, if we put an empty file with:
`overblog_graphiql: ~`
the bot won't allow it

I've then removed the link, would still be good to point users to documentation somehow

overblog/graphiql-bundle/0.1/post-install.txt
```diff
@@ -0,0 +1,3 @@
+<bg=blue;fg=white> Next for OverblogGraphiQLBundle </>
+    1. The bundle is auto configured now for <comment>dev</comment> environment only
```
fabpot: I would still remove this line which does not say anything useful.

renatomefi: Sure!

renatomefi: @fabpot done

PR #128, opened 2017-10-31
Add Sentry 1.0

ruudk:
| Q             | A
| ------------- | ---
| License       | MIT

This adds Sentry ~0.8~ 1.0 support. 

The difference with #40 is:
* It uses ~0.8~ 1.0 instead of 0.7
* It is installed always, so not only in `prod`. This is handy, because it allows you to depend on `SentrySymfonyClient` in all environments. Only in prod it's filled with the `SENTRY_DSN` so that it will send reports to Sentry.
* It uses `curl_method: async` by default, which is better if you ask me. It doesn't block the requests for the end-user when an exception occurs, but moves it to a shutdown listener.


ruudk:
Rebased this PR to use 1.0 version

PR #120, opened 2017-10-27
Added Manifest file for AWS SDK Bundle

trandangtri:
| Q             | A
| ------------- | ---
| License       | MIT

Manifest for AWS SDK. This one closes for an issue: https://github.com/symfony/flex/issues/193

PR #130, opened 2017-11-02
[WIP] Added Manifest for TriTranSqsQueueBundle

trandangtri:
| Q             | A
| ------------- | ---
| License       | MIT

This bundle provides an easy way to work with AWS SQS.

https://github.com/trandangtri/sqs-queue-bundle

### DIFF
--- a/tritran/sqs-queue-bundle/2.0/config/packages/sqs_queue.yaml
+++ b/tritran/sqs-queue-bundle/2.0/config/packages/sqs_queue.yaml
@@ -1,0 +1,3 @@
+tritran_sqs_queue:
+    sqs_queue:
+        queues: ~
--- a/tritran/sqs-queue-bundle/2.0/manifest.json
+++ b/tritran/sqs-queue-bundle/2.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "TriTran\\SqsQueueBundle\\TriTranSqsQueueBundle": ["all"]
+    },
+    "copy-from-recipe": {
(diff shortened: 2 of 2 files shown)


trandangtri:
Blocked by https://github.com/symfony/recipes-contrib/pull/120

tritran/sqs-queue-bundle/2.0/config/packages/sqs_queue.yaml
```diff
@@ -0,0 +1,3 @@
+tritran_sqs_queue:
+    sqs_queue:
+        queues: ~
```
Nyholm: What would happen if this config was removed?

trandangtri: Hi, It's just a sample structure of configuration. Basically, there are not any error which belongs to.
BTW, this PR is still WIP, sorry about that. I need to update something else.  Thanks

Nyholm: Okey. So if I understand you correctly, you could remove this line and still get the same results?

If so, (Looking at the current state of the PR) I would suggest closing this PR because Flex is smart enough to install this bundle without a recipe. 

However, I would love if you added some good default config that will make sense for 90% of the users. If so, I'll be happy to merge this PR.

trandangtri: Thanks for your comments. Seems I should close this PR.

PR #132, opened 2017-11-07
Add ElasticsearchBundle recipe

saimaz:
| Q             | A
| ------------- | ---
| License       | MIT

This bundle provides extensive integration with the official elasticsearch client library.

### DIFF
--- a/ongr/elasticsearch-bundle/5.0/config/packages/ongr_elasticsearch.yaml
+++ b/ongr/elasticsearch-bundle/5.0/config/packages/ongr_elasticsearch.yaml
@@ -1,0 +1,10 @@
+# Read the documentation: http://docs.ongr.io/ElasticsearchBundle
+ongr_elasticsearch:
+    managers:
+        default:
+            index:
+                hosts:
+                    - 127.0.0.1:9200
+                index_name: my-app-index
+            mappings:
+                - AppBundle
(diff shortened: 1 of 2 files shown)


ongr/elasticsearch-bundle/5.0/config/packages/ongr_elasticsearch.yaml
```diff
@@ -0,0 +1,10 @@
+# Read the documentation: http://docs.ongr.io/ElasticsearchBundle
+ongr_elasticsearch:
+    managers:
+        default:
+            index:
+                hosts:
+                    - 127.0.0.1:9200
+                index_name: my-app-index
+            mappings:
+                - AppBundle
```
fabpot: We don"t have `AppBundle` in Flex's project, so that looks wrong to me.

saimaz: I could rename it to the `YourBundleWithDocuments` or smth like this. It's an example of mapping documents from the bundle.

saimaz: Or maybe `AcmeBundle` ?

fabpot: What I mean is that we don't have bundles anymore in a Flex application for project's code, but a simple `App` namespace under `src/`.

PR #133, opened 2017-11-13
:heavy_plus_sign: Add Bukashk0zzz LiipImagineSerializationBundle

Bukashk0zzz:
| Q             | A
| ------------- | ---
| License       | MIT


stof:
Do users need to change the config often, or are the default values suitable for most cases ?

Options that are only an advanced setting should be removed from the recipe. The recipe is not a replacement for the documentation, and it is better for projects to rely on default values for them instead of configuring the advanced parameter explicitly (and then having developers reading the project config to understand these settings).

Bukashk0zzz:
@stof All options are optional. I updated PR and leave only often changed options.

PR #134, opened 2017-11-16
Add recipe for eightpoints/guzzle-bundle

gregurco:
| Q             | A
| ------------- | ---
| License       | MIT

This bundle integrates Guzzle 6.x into Symfony. Guzzle is a PHP framework for building RESTful web service clients.

### DIFF
--- a/eightpoints/guzzle-bundle/7.0/config/packages/dev/eight_points_guzzle.yaml
+++ b/eightpoints/guzzle-bundle/7.0/config/packages/dev/eight_points_guzzle.yaml
@@ -1,0 +1,2 @@
+eight_points_guzzle:
+    logging: true
--- a/eightpoints/guzzle-bundle/7.0/config/packages/dev/eight_points_guzzle.yaml
+++ b/eightpoints/guzzle-bundle/7.0/config/packages/dev/eight_points_guzzle.yaml
@@ -1,2 +1,0 @@
-eight_points_guzzle:
-    logging: true
--- a/eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
+++ b/eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
@@ -1,0 +1,2 @@
+eight_points_guzzle:
+    clients: ~
--- a/eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
+++ b/eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
@@ -1,11 +1,13 @@
 eight_points_guzzle:
     clients:
-        api_payment:
-            base_url: "http://api.domain.tld"
+        my_client:
+            base_url: "http://target.url"
 
             options:
                 timeout: 30
 
+                # Configure headers. Ex: what do you accept as response (application/json, text/html, etc..)
+                # More info: http://docs.guzzlephp.org/en/stable/request-options.html#headers
                 headers:
                     Accept: "application/json"
 
--- a/eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
+++ b/eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
@@ -1,15 +1,16 @@
 eight_points_guzzle:
     clients:
         my_client:
+            # Write here the host where to do requests
             base_url: "http://target.url"
 
             options:
                 timeout: 30
 
-                # Configure headers. Ex: what do you accept as response (application/json, text/html, etc..)
+                # Configure headers.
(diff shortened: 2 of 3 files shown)


eightpoints/guzzle-bundle/7.0/config/packages/dev/eight_points_guzzle.yaml
```diff
@@ -0,0 +1,2 @@
+eight_points_guzzle:
+    logging: true
```
Nyholm: The default value for this parameter is KERNEL_DEBUG, right?

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
```diff
@@ -0,0 +1,2 @@
+eight_points_guzzle:
+    clients: ~
```
Nyholm: Is this really needed? I just looked at the code quickly, it seams like the bundle will have the exact same behaviour without these two lines, is that correct?

eightpoints/guzzle-bundle/7.0/config/packages/dev/eight_points_guzzle.yaml
gregurco: yep, `%kernel.debug%`

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
gregurco: yep, behaviour the same, but user will know where to add clients...

gregurco:
@Nyholm I added client in config file, good idea :+1:

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
```diff
@@ -0,0 +1,14 @@
+eight_points_guzzle:
+    clients:
+        api_payment:
```
Nyholm: Rename to ‘my_client’

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
```diff
@@ -0,0 +1,14 @@
...
+            base_url: "http://api.domain.tld"
```
Nyholm: Remove

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+            options:
+                timeout: 30
+
+                headers:
+                    Accept: "application/json"
```
Nyholm: Add a comment before this line to explain what we are doing. 
Maybe aleo change this to an less important header. Like the User-Agent header.

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
gregurco: Better to keep here some URL. Probably one 100% not active or to write `http://target.url`. It's better to enforce users to write host here, not in controllers and it's better to find proper way to do that

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
gregurco: wrote comment

Nyholm:
The point of the recipes is to bring the users good default configuration that will work for 90% of the users. When we have a base_url, the user cannot start using the new services that the bundle provides without editing the configuration. 
Same with the Accept header. I think it is important to show how to configure headers by example like you are doing. But specifying "Accept: application/json" it will not be working for 90% of the cases. I still suggest to use the "User-Agent" header.

That is just my 50 cent.

gregurco:
@Nyholm ok, but which one to use? `test/1.0`, `EightpointsGuzzleBundle/1.0` or another one?

Nyholm:
I think `EightpointsGuzzleBundle/7.0` would be a good User-Agent.

eightpoints/guzzle-bundle/7.0/config/packages/eight_points_guzzle.yaml
```diff
@@ -0,0 +1,17 @@
+eight_points_guzzle:
+    clients:
+        my_client:
+            # Write here the host where to do requests
+            base_url: "http://target.url"
```
Nyholm: You like this line and I agree it is important. But lets make it a comment and this service will be useable without the user has to change the config.

gregurco: A leave `base_url` here but commented it. Is it ok?

PR #135, opened 2017-11-16
Add spiechu symfony commons bundle

spiechu:
| Q             | A
| ------------- | ---
| License       | MIT

Bundle allows to validate JSON / XML responses against given JSON Schema / XSD files using simple controller annotations.
Also allows API versioning and provides compatibility with JMSSerializerBundle `since-version` & `until-version` attributes.

Notice that JMSSerializerBundle and FOSRestBundle are not yet fully compatible with Symfony 4.

### DIFF
--- a/spiechu/symfony-commons-bundle/0.1/config/packages/spiechu_symfony_commons.yaml
+++ b/spiechu/symfony-commons-bundle/0.1/config/packages/spiechu_symfony_commons.yaml
@@ -1,0 +1,24 @@
+spiechu_symfony_commons:
+    get_method_override:
+        enabled:              false
+        #listener_service_id:  spiechu_symfony_commons.event_listener.get_method_override_listener
+        #query_param_name:     _method
+        #allow_methods_override:
+            # Defaults:
+            #- DELETE
+            #- POST
+            #- PUT
+    response_schema_validation:
+        enabled:              false
+        #throw_exception_when_format_not_found: true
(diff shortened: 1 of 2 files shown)


spiechu/symfony-commons-bundle/0.1/config/packages/spiechu_symfony_commons.yaml
```diff
@@ -0,0 +1,24 @@
+spiechu_symfony_commons:
+    get_method_override:
+        enabled:              false
+        #listener_service_id:  spiechu_symfony_commons.event_listener.get_method_override_listener
```
Nyholm: Do we really need these comments?
Will they help the user or would it be better to refer them to the readme?

spiechu: To be honest I was suggested by other already accepted recipes.
I saw configuration dumps fully commented out. I'm not sure what is better for DX.

Of course I can remove any commented out configuration and point user to readme.

Nyholm: Hm, I see. 

The rule of thumb is: Do you think most (80%) of the users will change these values? If not, we should refer them to your docs instead.

spiechu:
I've removed commented out configuration. I think PR is ready for final CR.

PR #137, opened 2017-11-17
JWT Framework Suite Recipes

Spomky:
| Q             | A
| ------------- | ---
| License       | MIT

This PR adds recipes for the JWT Framework Suites.


Nyholm:
Thank you for this PR. 

I see that you dumped all configuration of the bundle as comments. As a general rule of thumb we should add configuration that fits for 90% of the users in a recipe. For more "edge case stuff" we should link them to the bundle's readme.

Spomky:
Hi @Nyholm 

Thanks for the review. I will completely remove the configuration file.

Spomky:
Done. Ready to merge if you agree with the changes

PR #139, opened 2017-11-18
Recipe for knp-snappy-bundle

jzawadzki:
| Q             | A
| ------------- | ---
| License       | MIT

Small recipe for https://github.com/KnpLabs/KnpSnappyBundle

#SymfonyConHackday2017


jzawadzki:
not sure which version should be here as minimum - need to wait for symfony4 support: https://github.com/KnpLabs/KnpSnappyBundle/pull/190

weaverryan:
Flex also works with Symfony 3.3, so using whatever the current latest version of snappy is great :). If there is a new version released later, it will still use this recipe

PR #140, opened 2017-11-18
mailxpert apibundle

scuben:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/mailxpert/apibundle/0.1/manifest.json
+++ b/mailxpert/apibundle/0.1/manifest.json
@@ -1,0 +1,9 @@
+{
+    "bundles": {
+        "Mailxpert\\APIBundle\\MailxpertAPIBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "src/Entity/AccessToken.php": "%SRC_DIR%/Entity/AccessToken.php"
+    }
+}
--- a/mailxpert/apibundle/0.1/manifest.json
+++ b/mailxpert/apibundle/0.1/manifest.json
@@ -5,5 +5,9 @@
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/",
         "src/Entity/AccessToken.php": "%SRC_DIR%/Entity/AccessToken.php"
+    },
+    "env": {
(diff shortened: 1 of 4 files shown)


mailxpert/apibundle/0.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "bundles": {
+        "Mailxpert\\APIBundle\\MailxpertAPIBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "src/Entity/AccessToken.php": "%SRC_DIR%/Entity/AccessToken.php"
```
fabpot: I would use `src/` here instead.

mailxpert/apibundle/0.1/src/Entity/AccessToken.php
```diff
@@ -0,0 +1,26 @@
...
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    protected $id;
+
+    public function __construct()
+    {
+        parent::__construct();
+        // your own logic
+    }
```
fabpot: I would remove the constructor altogether as the tip is just plain PHP.

mailxpert/apibundle/0.1/src/Entity/AccessToken.php
```diff
@@ -0,0 +1,21 @@
...
+ */
+class AccessToken extends BaseAccessToken
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue(strategy="AUTO")
+     */
+    protected $id;
+
```
fabpot: empty line could be removed.

mailxpert/apibundle/0.1/manifest.json
scuben: I had it like that but it got interfered with the orm repo recipe as it would have removed the empty `Entity` folder. But that gave an exception due to doctrine then did not found the `Entity` folder. should I switch it back to `"src/": "%SRC_DIR%"` anyway?

fabpot: I've just tried with `src/` and everything works fine for me.

PR #143, opened 2017-11-18
Added SgDatatablesBundle recipe

samuel4x4:
| Q             | A
| ------------- | ---
| License       | MIT

This PR adds recipes for the SgDatatablesBundle (PR https://github.com/stwe/DatatablesBundle/pull/716)

#SymfonyConHackday2017

### DIFF
--- a/sg/datatablesbundle/1.0/config/routes/sg_datatables_bundle.yaml
+++ b/sg/datatablesbundle/1.0/config/routes/sg_datatables_bundle.yaml
@@ -1,3 +1,4 @@
 sg_datatables_bundle:
     resource: "@SgDatatablesBundle/Controller/"
     type:     annotation
+    prefix:   /sg
--- a/sg/datatablesbundle/1.0/config/routes/sg_datatables_bundle.yaml
+++ b/sg/datatablesbundle/1.0/config/routes/sg_datatables_bundle.yaml
@@ -1,0 +1,3 @@
+sg_datatables_bundle:
+    resource: "@SgDatatablesBundle/Controller/"
+    type:     annotation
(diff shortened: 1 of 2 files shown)


sg/datatablesbundle/1.0/manifest.json
```diff
@@ -0,0 +1,12 @@
...
+    "bundles": {
+        "Sg\\DatatablesBundle\\SgDatatablesBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "composer-scripts": {
+        "make cache-warmup": "script",
+        "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd"
+    }
```
fabpot: Looks like this is a copy/paste remaining from framework-bundle, which should be removed.

sg/datatablesbundle/1.0/config/routes/sg_datatables_bundle.yaml
```diff
@@ -0,0 +1,3 @@
+sg_datatables_bundle:
+    resource: "@SgDatatablesBundle/Controller/"
+    type:     annotation
```
fabpot: I haven't had a look at the bundle, but don't we need a `prefix` here?

samuel4x4: There wasn't a global bundle prefix, I just added one moving it from the controller annotation on https://github.com/stwe/DatatablesBundle/pull/716. I updated their documentation and now I'll proceed to update the flex recipe also. I think it's a good practice to have a global bundle routing prefix.

sg/datatablesbundle/1.0/manifest.json
samuel4x4: done

samuel4x4:
Thank you @fabpot ;)

PR #147, opened 2017-11-21
Adding BCCCronManagerBundle 3.1 recipe

samuel4x4:
| Q             | A
| ------------- | ---
| License       | MIT

This PR adds recipes for the BCCCronManagerBundle 3.1 (https://github.com/michelsalib/BCCCronManagerBundle/pull/82 and https://github.com/michelsalib/BCCCronManagerBundle/pull/84)

### DIFF
--- a/bcc/cron-manager-bundle/3.1/config/packages/bcc_cron_manager_bundle.yaml
+++ b/bcc/cron-manager-bundle/3.1/config/packages/bcc_cron_manager_bundle.yaml
@@ -1,3 +1,0 @@
-security:
-    access_control:
-        - { path: ^/admin, roles: ROLE_USER }
--- a/bcc/cron-manager-bundle/3.1/config/packages/bcc_cron_manager_bundle.yaml
+++ b/bcc/cron-manager-bundle/3.1/config/packages/bcc_cron_manager_bundle.yaml
@@ -1,0 +1,3 @@
+security:
+    access_control:
+        - { path: ^/admin, roles: ROLE_USER }
--- a/bcc/cron-manager-bundle/3.1/config/routes/bcc_cron_manager_bundle.yaml
+++ b/bcc/cron-manager-bundle/3.1/config/routes/bcc_cron_manager_bundle.yaml
@@ -1,0 +1,3 @@
+BCCCronManagerBundle:
+    resource: "@BCCCronManagerBundle/Resources/config/routing.xml"
+    prefix:   admin/cron-manager
--- a/bcc/cron-manager-bundle/3.1/manifest.json
+++ b/bcc/cron-manager-bundle/3.1/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "BCC\\CronManagerBundle\\BCCCronManagerBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}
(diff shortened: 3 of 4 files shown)


bcc/cron-manager-bundle/3.1/config/packages/bcc_cron_manager_bundle.yaml
```diff
@@ -0,0 +1,3 @@
+security:
```
Nyholm: When installing this recipe on a new installation I've got the error that "security" is not found.

samuel4x4: I'll check, thanks.

samuel4x4: Hi @Nyholm, since symfony/security-bundle is not installed by default in Symfony 4, it can be added and there's no an issue with the "security" configuration. I can add it as a dependency in the composer.json file of bcc/cron-manager-bundle, but I'm thinking if this should not be a decision of the user. Instead of adding the security dependency, I'm thinking just to display in the post-install-output configuration the suggestion for the user to add him self the security bundle, if he wants to secure the /admin/cron-manager route. Which approach do you think is better?

Nyholm: The idea is that the bundle should work out of the box. If you have a real dependency on symfony/security-bundle you should put that in your composer.json. 

If you want to let the user decide if the security bundle should be used or not, then make sure to change the logic in the bundle so it works (out of the box) without the security bundle.

samuel4x4: I added symfony/security-bundle in composer.json.

bcc/cron-manager-bundle/3.1/config/packages/bcc_cron_manager_bundle.yaml
```diff
@@ -0,0 +1,3 @@
...
+    access_control:
+        - { path: ^/admin, roles: ROLE_USER }
```
Nyholm: Learning from other packages and discussions: We should not modify other bundles config.

Instead you should put a post-install note that the user need to make this change.

samuel4x4: I replaced it from the config, post install output will recommendations the user to do it

samuel4x4:
@Nyholm, @fabpot I see here interesting situation and I want to share my thoughts with you. 

I created the recipe for the last stable bundle version (3.1), but it was already released some time ago and it doesn't support Symfony 4. The Symfony 4 support was merged in the master branch, it will be released in 3.2, maybe, it depends of the bundle owner. So, if 3.2 will include it, I'll have to update the recipe also, to apply starting from 3.2. It's like a ping-pong between Bundle and Flex recipe :) Normally, I should wait for the bundle, to release a new version with Symfony 4 support included, then I know the version I should use in the flex recipe creation.

Flex is great and helps a lot the developers when 3rd-party bundles are installed. But I see this part should be closer to the bundle it self, not outside of it. It can come maybe with the bundle directly. What about having in composer.json something like this:
```json
{
    "scripts": {
        "flex-scripts": [
            "RecipeNamespace\\ScriptHandler::applyRecipe",
        ],
        "post-install-cmd": [
            "@symfony-scripts",
            "@flex-scripts"
        ]
    }
}
```
so the bundle owner can decide and provide directly with the bundle what should happens in order to configure its bundle automatically. 

Having the community helping on the recipes creation is great, and I think this can be kept as well having this part in the bundle side.

What do you think? 

Thanks!

Nyholm:
Im not sure. If you have the library decoupled from the bundle you want them to stay decoupled. Adding more logic and dependencies between them seams weird. 

------

Im happy with this PR now. Is it ready for be merged?

samuel4x4:
I was thinking this is the bundle purpose, to assure everything needed for library integration in a Symfony application. Anyway, the auto-configuration provided by Flex is really great. 

The PR is ready, you can merge it.

PR #148, opened 2017-11-21
Adding BCCCronManagerBundle 2.6 recipe

samuel4x4:
| Q             | A
| ------------- | ---
| License       | MIT

This PR adds recipes for the BCCCronManagerBundle 2.6 (https://github.com/michelsalib/BCCCronManagerBundle/pull/82 and https://github.com/michelsalib/BCCCronManagerBundle/pull/84)

### DIFF
--- a/bcc/cron-manager-bundle/2.6/config/packages/bcc_cron_manager_bundle.yaml
+++ b/bcc/cron-manager-bundle/2.6/config/packages/bcc_cron_manager_bundle.yaml
@@ -1,0 +1,3 @@
+security:
+    access_control:
+        - { path: ^/admin, roles: ROLE_USER }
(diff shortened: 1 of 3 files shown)


Nyholm:
It seams like this recipe is that same as #147, right? 

The recipe for 2.6 will be used when installing 3.1 if no other recipes exists.

samuel4x4:
Hi @Nyholm, thank you for review. I started both 3.1 and 2.6 with the same content but I'm not sure they will stay the same. Maybe some changed should be done on the bundle 2.6 in order to can be used with flex. The validation says "Package must not depend on symfony/symfony", so maybe their 2.6 composer.json can contains explicitly the Symfony components, instead of symfony/symfony. What do you think?

Nyholm:
All (maintained) versions of all bundles should specify symfony components specifically. 

Im just saying, if the recipe for 2.6 and 3.1 is the same, we can remove the recipe for 3.1 and the user will not notice.

Nyholm:
How do you want to do with this PR now hen #147 is merged?

samuel4x4:
I think we can close this one for now.

PR #151, opened 2017-11-21
Add sonata/core-bundle

rande:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sonata-project/core-bundle/3.7/config/packages/sonata_core.yaml
+++ b/sonata-project/core-bundle/3.7/config/packages/sonata_core.yaml
@@ -1,0 +1,9 @@
+sonata_core:
+    flashmessage:
+        error:
+            css_class: danger
+            types:
+                - {type: sonata_product_error, domain: "SonataProductBundle" }
+        success:
+            types:
+                - {type: sonata_customer_success, domain: "SonataCustomerBundle" }
--- a/sonata-project/core-bundle/3.7/manifest.json
+++ b/sonata-project/core-bundle/3.7/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Sonata\\CoreBundle\\SonataCoreBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


rande:
@fabpot  not sure how to enable the session service from my definition ?

```
 [KO]
Script cache:clear returned with error code 1
!!
!!
!!    [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]
!!    The service "sonata.core.flashmessage.manager" has a dependency on a non-ex
!!    istent service "session".
!!
!!
!!
```

rande:
Or maybe the bundle should work without the service ?

Nyholm:
Yeah, do you **really** need the session service in the bundle?

jordisala1991:
This config is only for Symfony >= 3.4?

Nyholm:
Na, you can use flex with sf3.3.

sonata-project/core-bundle/3.7/config/packages/sonata_core.yaml
```diff
@@ -0,0 +1,9 @@
+sonata_core:
+    flashmessage:
+        error:
+            css_class: danger
+            types:
+                - {type: sonata_product_error, domain: "SonataProductBundle" }
+        success:
+            types:
+                - {type: sonata_customer_success, domain: "SonataCustomerBundle" }
```
greg0ire: Form mapping should be disabled by default, can you please add this? :
```yaml
sonata_core:
    form:
        mapping:
            enabled: false
```

covex-nn: To fix `The service "sonata.core.flashmessage.manager" has a dependency on a non-existent service "session".` error add this:

```yaml
framework:
    session: ~
```

covex-nn: To fix `The service "sonata.core.flashmessage.manager" has a dependency on a non-existent service "session".` error add this:

```yaml
framework:
    session: ~
```

sonata-project/core-bundle/3.7/config/packages/sonata_core.yaml
```diff
@@ -0,0 +1,9 @@
+sonata_core:
+    flashmessage:
```
jordisala1991: Do we need this config as a default, I never used it on my projects

aLkRicha:
@greg0ire will we merge this?

OskarStark:
@aLkRicha can you please create a new PR with the minimal config?

aLkRicha:
@OskarStark this pr is wrong?

OskarStark:
> @OskarStark this pr is wrong?

I guess not, but I don't know when @rande will find some time to give feedback on the comment of @jordisala1991, so getting this merged will be quicker if you create a new one IMO

But I could be wrong 😄

aLkRicha:
@OskarStark ok, what is the best way to not use sessions in this bundle?

Nyholm:
Could this be closed?

rande:
This PR is a way to understand how works flex recipe. My understanding is:
- flex does not provide any mechanism to change configuration (I can understand why)
- this required to make bundles works as it with the minimal configuration possible - so any changes in the global base configuration might break a flex recipe.
- we (sonata's contributor) need to review all bundles to make them work properly with missing services.
- update the documentation to explain changes to apply in the global configuration (for instance, enable the session service)

Not sure why you want to close this PR, this work is still valid, not the related bundle.

jordisala1991:
Aparently it was merged here @rande ... https://github.com/symfony/recipes-contrib/pull/240

Anyway it would be nice to be able to review it before merge, because I am still not sure about that flashmessage configuration...

rande:
@jordisala1991 yep, maybe the @sonata-project/contributors should check the configuration. Some settings was there to work with the sonata's sandbox demo.

I guess this can be remove as the PR #240 did.

PR #159, opened 2017-11-27
New Recipe for common-js-bundle

silverbackdan:
| Q             | A
| ------------- | ---
| License       | MIT

A new bundle recipe for js-sdk-bundle


Nyholm:
Hey. 
Thank you for this PR. I like that you created `post-install.txt`. 

It seams that your config is all comments. Could we try to provide a good default configuration that would work for 90% of your users? 

Because as it is now, a recipe is not needed. Flex is so awesome that it will install and activate your bundle without a recipe =)

silverbackdan:
Thanks for taking a look at this PR Nyholm.

Sure, it's commented out just because it's an example template config and will make it easier for a user to quickly configure. The ID would have to be a Google tracking ID which will be different for every user but I think 90% of users would simply uncomment the lines and change the id to their own tracking ID which will mean they can easily add page view tracking.

You are right though, right now I'm barely touching the requirements to need a recipe. I wanted to get started with it and check out how it works (I found it really cool!) - I really like that Flex just adds the bundle into the bundles.php file too, really enjoying the changes in SF3.4/4

I'll make some more modifications to this PR as I add more SDKs.

Is it worth having a recipe though to make it quicker and easier for a user to configure the bundle just by commenting out lines for the 90% use case?

Nyholm:
Will **all** the users remove that comment? If so why not write something like (with no comments): 

```yaml
js_sdk:
    google_analytics:
        id: UA-12345678
```

silverbackdan:
I think to give the full picture I'll finish the implementation of other SDKs and have those configs in there commented too because perhaps not.

If I were using it I'd always uncomment google_analytics, but others may end up using the finished bundle just for the Facebook Pixel SDK or Woopra Analytics for example and when I've done those I'll update these config templates to make it easy for someone using the bundle to quickly enable any one (or more) of the SDKs.

Thinking aloud: I could perhaps make the id an env parameter but it seems a bit convoluted if it'll only ever be used in the config. It may be different in dev and prod, but that can be configured straight in config files too and the value doesn't need to be secret because it is output into the final HTML anyway.

fabpot:
I would definitely use an env var for something like the Google Analytics ID.

silverbackdan:
OK cool, I'll do that then and follow suggestions of the code being uncommented but with a parameter for the ID which I'll add to the env configuration. Thank you both for your time giving feedback.

silverbackis/common-js-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "JsSdkBundle\\JsSdkBundle": ["all"]
```
gregurco: My suggestion is to normalize path.. first of all, it's your realization of JsSdkBundle, probably somebody would like to do bundle with same name, and what to do in this case? conflict will be caused. My suggestion is to add `Silverbackis` before path. Finally you will have `Silverbackis\\JsSdkBundle\\JsSdkBundle`. You can inspire by top bundles, see examples here: https://packagist.org/profile/?q=bundle&p=0
You need to do changes in your bundle first.

silverbackis/common-js-bundle/1.0/config/packages/common_js.yaml
```diff
@@ -0,0 +1,17 @@
+# Enable and configure the scripts you'd like
+common_js:
+    google_analytics:
+        enabled: false
+        debug: '%kernel.debug'
```
gregurco: I see that your bundle is not released yet. And by default `debug` is `false`, but there you set another default value. It's not late to do this functionality by default in bundle config. My suggestion is to do that in: https://github.com/silverbackis/common-js-bundle/blob/master/DependencyInjection/Configuration.php
See examples here:
https://github.com/8p/EightPointsGuzzleBundle/blob/master/src/DependencyInjection/Configuration.php#L52
https://github.com/schmittjoh/JMSSerializerBundle/blob/master/DependencyInjection/Configuration.php#L195

silverbackis/common-js-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "GOOGLE_ANALYTICS_ID": "UA-XXXXXXXX",
```
gregurco: will all users need these 3 env variables?

silverbackis/common-js-bundle/1.0/manifest.json
silverbackdan: Cool, understood - I'll do that thanks.

silverbackis/common-js-bundle/1.0/config/packages/common_js.yaml
silverbackdan: As I switched off yesterday though that too - I'll set the default as suggested

silverbackis/common-js-bundle/1.0/manifest.json
silverbackdan: Not all users would, no, I'll comment them out to be included as wanted and adjust the post-install - thanks again.

silverbackdan:
@gregurco I have made the environment variables into comments as they will not all be needed by all users. Is the latest PR how you'd recommend doing this and what you were referring to doing when you asked if all the environment vars would be needed by all users?

Nyholm:
Im not too sure I like this. 

- We add 3 parameters in .env that are commented out
- We add some config which is same as default
- We add more config which are commented out, and Im not sure I understand it without looking at the docs. 

My point is (same as before) Either we add some default config that are good for 90% of the users.  Or we do not have a recipe since it does not bring any additional value. 

### An idea: 
What if you wrote your bundle to automatically enable GoogleAnalytics if the environment variable exists and has a value? (And of course if no config says it should be disabled) 

Then the recipe could include adding some environment variables and nothing more. Would that work?

silverbackdan:
I see, that could work yes. When I first started making this recipe I thought they could be used to create commented out configs for a user to easily uncomment what was wanted but I didn't realise that was not the point at all.

With the input I've received the bundle which be much better so thank you for the comments. I think I will be able to change it to just be env vars and the post-install.txt and it'll work well, I'll update the PR later today.

silverbackis/common-js-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "Silverback\\CommonJsBundle\\SilverbackCommonJsBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
```
Nyholm: This is not needed anymore

silverbackis/common-js-bundle/1.0/post-install.txt
```diff
@@ -0,0 +1,7 @@
+<bg=blue;fg=white>                    </>
+<bg=blue;fg=white>    What's next?    </>
+<bg=blue;fg=white>                    </>
+
+Now edit your .env file and set the IDs for scripts you want to enable
+
+* <fg=blue>Read</> the documentation at <comment>https://github.com/silverbackis/common-js-bundle</> to find out the blocks available and how to use the bundle.
```
Nyholm: Lets not use the colors.

silverbackis/common-js-bundle/1.0/manifest.json
silverbackdan: Both the "bundles" and "copy-from-recipe" keys?

silverbackdan: Sorry, stupid question. Removed both - the config dir was not required.

silverbackis/common-js-bundle/1.0/post-install.txt
silverbackdan: Sure, no problem - the example in docs used colours just like this but I've removed them.

silverbackdan:
Sorry for being the best example of how not to write a recipe. I've added 1 simple manifest in for another bundle here. The CI says I am only registering a bundle for common-js-bundle but I'm setting env vars too (tried with empty vars and not) I've got a post-install.txt as well.

Is this me being stupid again (likely) or a CI check that hasn't picked up on the env vars?

silverbackis/api-component-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "bundles": {
+        "Silverback\\ApiComponentBundle\\SilverbackApiComponentBundle": ["all"],
+        "Cocur\\Slugify\\Bridge\\Symfony\\CocurSlugifyBundle": ["all"]
```
Nyholm: This bundle does not belong in the API-component package, right?

silverbackis/api-component-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
```
Nyholm: This file has nothing to do with `common-js-bundle`, right?

silverbackis/api-component-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,10 @@
...
+    "bundles": {
+        "Silverback\\ApiComponentBundle\\SilverbackApiComponentBundle": ["all"],
+        "Cocur\\Slugify\\Bridge\\Symfony\\CocurSlugifyBundle": ["all"],
+        "Liip\\ImagineBundle\\LiipImagineBundle": ["all"]
```
Nyholm: You should remove everything that is not `common-js-bundle`. Not add more bundles here =)

silverbackdan: Ah! I had not seen the previous comment before adding this

I'll do that :)

silverbackdan:
If the reason this does not pass is because you'd rather not have recipes with empty env vars and a post-install text file alone, that's OK I'm happy enough to close this and just submit a PR for api-component-bundle as I feel I've wasted too much of your time. I'm pretty sure the api-component-bundle one would be OK as it passes validation. I really appreciate your time on this PR though!

Nyholm:
Yeah. Doing one bundle at the time would make things easier. Im looking forward to that PR.

PR #163, opened 2017-11-30
Add recipe for LongRunning

ruudk:
| Q             | A
| ------------- | ---
| License       | MIT

Adds support for https://github.com/LongRunning/LongRunning. It uses a different path for the bundle so this will help others by auto installing it :)

### DIFF
--- a/long-running/long-running/0.1/manifest.json
+++ b/long-running/long-running/0.1/manifest.json
@@ -1,0 +1,5 @@
+{
+    "bundles": {
+        "LongRunning\\Bundle\\LongRunningBundle\\LongRunningBundle": ["all"]
+    }
+}


ruudk:
```
Package long-running/long-running (0.1)
Recipe is not needed as it only registers bundle "LongRunning\Bundle\LongRunningBundle\LongRunningBundle" for all environments
Type must be "symfony-bundle" as the manifest registers a bundle, "library" detected
```

@fabpot I guess that this is not supported right? The package that I want to auto-install is a library that contains also a Bundle, see https://github.com/LongRunning/LongRunning 

Not sure how to proceed.

fabpot:
It's not supported by the validator, but we can bypass the validator for such cases.

stof:
There is a big drawback about putting the bundle in the same package than the repository: you cannot require any of the bundle-specific requirements, and so you cannot specify compatibility with Symfony for instance (btw, this means that any change in the supported Symfony versions, including dropping old versions, or the components being used is a BC break for your users and require a semver major release, because you cannot rely on composer to install compatible versions)

ruudk:
@stof It's totally not something to recommend, but it exists in the open. I can imagine that you don't want to support this case, in that case I'm fine with it as well. But if it happens more, it should be possible I think.

stof:
@ruudk well, I know it exists. But I would recommend you to consider switching to separate packages instead (requires a major versions as the bundle would be removed from the existing library package).
and in the meantime, @fabpot said he could bypass the validator

ruudk:
Unfortunately I'm not the maintainer of LongRunning, but I saw this (abandoned) PR to split it up into single packages: https://github.com/LongRunning/LongRunning/pull/16

ruudk:
How to proceed with this? Is it possible to merge this? Or should we close the PR?

PR #164, opened 2017-11-30
Created the Apache recipe.

ricknox:
| Q             | A
| ------------- | ---
| License       | MIT

This is related to symfony/symfony-docs/issues/8724 and symfony/recipes/issues/190.


ricknox:
This has already been submitted in symfony/recipes/pull/289 but belongs in `recipes-contrib`. Updated this PR in #da5a9bb. And fixed the issue https://github.com/symfony/recipes-contrib/pull/164#discussion_r154106229 in ede139a.

symfony/apache-pack/1.0/public/.htaccess
```diff
@@ -0,0 +1,68 @@
...
+    # Rewrite all other queries to the front controller.
+    RewriteRule ^ %{ENV:BASE}/index.php [L]
+</IfModule>
+
+<IfModule !mod_rewrite.c>
+    <IfModule mod_alias.c>
+        # When mod_rewrite is not available, we instruct a temporary redirect of
+        # the start page to the front controller explicitly so that the website
+        # and the generated links can still be used.
+        RedirectMatch 302 ^/$ /index.php/
```
greg0ire: Wouldn't 307 be better here?

greg0ire: https://httpstatuses.com/307

ricknox: 307 came about because user agents adopted as a de facto behaviour to take POST requests that receive a 302 response and send a GET request to the Location response header. 307 was introduced to allow servers to make it clear to the user agent that a method change should not be made by the client when following the Location response header.

Not sure if we need to change it to 307.

weaverryan: not sure - make a new PR for it?

stof: this would have to be tested to see what happens for a Apache setup *not* having mod_rewrite. I'm not sure this is that common

Pierstoval: It's just a copy/paste from the standard-edition

PR #167, opened 2017-11-30
Add fresh/doctrine-enum-bundle

fre5h:
| Q             | A
| ------------- | ---
| License       | MIT

I put to the recipe a default directory for custom Doctrine types which are handled by this bundle


Nyholm:
Hey. 
Thank you for this PR. I've not seen this bundle before. It looks great. 

I see that the recipe does not contain much instructions ie no default config etc. Flex is smart enough to install fresh/doctrine-enum-bundle without this recipe. 

Try: 
```
composer create-project "symfony/skeleton:^4.0" sf
cd sf
composer req fresh/doctrine-enum-bundle
```

fre5h:
This recipe adds only default directory for Doctrine types under `DBAL\Types`. I agree that it is not a lot of work to be included into recipes, but at least a very small optimization :)

PR #168, opened 2017-12-01
Prepare recipe for version 3.x and drop 2.x

endroid:
| Q             | A
| ------------- | ---
| License       | MIT


endroid:
Hi @Nyholm normally I would leave it, but I think it is cleaner to remove it and start supporting recipes from 3.x because in 2.x the bundle was not separated from the library yet and the recipe was tied to the library, which means it is executed even if you don't use the bundle. I now created a endroid/qrcode-bundle package that only contains the bundle and want all recipes to be connected to bundles, not to libraries to avoid confusion and because it is cleaner.

The same goes for the endroid/cm-sms-bundle where I did the same. I am doing refactorings in all my libraries at the moment so bundles have a separate package, are built for PHP 7.x and support SF 4.0. So endroid/import and endroid/flusher will undergo the same process.

curry684:
This seems to be breaking stuff in really weird ways right now?
```sh
curry@ubuntu:$ composer search endroid/qrcode-bundle
endroid/qrcode-bundle Endroid QR Code Bundle
curry@ubuntu:$ composer req endroid/qrcode-bundle

                                                 
  [InvalidArgumentException]                     
  Could not find package endroid/qrcode-bundle.  
                                                 
  Did you mean this?                             
      endroid/qrcode-bundle                      
```
Not sure where to report this hehe.

PR #170, opened 2017-12-02
add recipe for BeelabTestBundle

garak:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/beelab/test-bundle/2.0/manifest.json
+++ b/beelab/test-bundle/2.0/manifest.json
@@ -1,0 +1,5 @@
+{
+    "bundles": {
+        "Beelab\\TestBundle\\BeelabTestBundle": ["test"]
+    }
+}    


Nyholm:
Thank you for this PR. 

We do only accept apache and MIT licences at this point. Do you want to consider changing to a more generous license?

garak:
Instead, what about being less strict about accepted licenses? I also opened an [issue on Flex](https://github.com/symfony/flex/issues/221)

alcaeus:
@Nyholm while I agree that you want to whitelist acceptable licenses, there shouldn't be a problem adding the LGPL to the list. I get why there may be issues with GPL, but the LGPL should be fine.

Nyholm:
Im happy with this recipe. Lets see how the discussion in https://github.com/symfony/flex/issues/221 goes.

fabpot:
@symfony-flex-server review

PR #172, opened 2017-12-03
Add datatables bundle

curry684:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/omines/datatables-bundle/0.1/config/packages/datatables.yaml
+++ b/omines/datatables-bundle/0.1/config/packages/datatables.yaml
@@ -1,0 +1,14 @@
+# Latest documentation available at https://omines.github.io/datatables-bundle/#configuration
+datatables:
+    # Set options, as documented at https://datatables.net/reference/option/
+    options:
+        lengthMenu : [10, 25, 50, 100, 250, 500, 1000, 2500]
+        pageLength: 50
+        dom: "<'row' <'col-sm-12' tr>><'row' <'col-sm-6'l><'col-sm-6 text-right'pi>>"
+
+    template_parameters:
+        # Example classes to integrate nicely with Bootstrap 3.x
+        className: 'table table-striped table-bordered table-hover data-table'
+
+    # You can for example override this to "tables" to keep the translation domains separated nicely
+    translation_domain: 'messages'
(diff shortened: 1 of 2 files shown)


Nyholm:
Thank you. 

When I try to install this bundle using the recipe (explained [here](https://symfony.sh/r/github.com/symfony/recipes-contrib/172)), I get the following error: 

> Cannot autowire service "Omines\DataTablesBundle\Adapter\Doctrine\ORMAdapter": argument "$registry" of method "__construct()" has type "Symfony\Bridge\Doctrine\RegistryInterface" but this class cannot be loaded.          

Does you bundle miss a requirement?

curry684:
Hmm no I'm just seeing this same error myself. That dependency is completely optional, no idea why Flex is borking on it ([Travis isn't](https://travis-ci.org/omines/datatables-bundle)).

curry684:
Hang on, we're not actually running tests without ORM installed, that's why the autowiring is ofc going insane on it. I'll get it fixed.

curry684:
@nicolas-grekas can I ping you in for your opinion on this? I'm not sure whether this is a bug or a shortcoming in DI, or that I'm just expecting a bit too much magic from autowire.

We have an `ORMAdapter` *shared* service in this bundle, which is autowired to `Symfony\Bridge\Doctrine\RegistryInterface`. Composer.json suggests the `doctrine/doctrine-bundle` if you want to use this specific adapter. I optimistically expected the autowiring in this case to stub the service to an exception, causing it to crash at runtime with something like `Service X is unusable because it depends on unknown class Y`. Instead it crashes at compile time because it tries to autowire a service that may never be used.

For now manual wiring with optional dependency seems to be the only solution, but it seems like we could improve the autowiring behavior here?

nicolas-grekas:
> Service X is unusable because it depends on unknown class Y

wow that'd be awesome (or magic, can't decide yet)
it doesn't work this way currently (not sure it should really for now...)
so yes, manual wiring still has some use cases :)

But in this case, I bet it's because the service is public? You should make it private really, that will remove the failure.

curry684:
It'd be cool magic though   ;)   And not all that tough even I think, it could just wire up a `MissingDependencyServicePlaceholder` or something.

The service itself is private, but tagged and injected in a ServiceLocator so we can resolve it at runtime.

curry684:
@Nyholm I circumvented the issue in a [new release](https://github.com/omines/datatables-bundle/commit/47e4a402cfb3e09a207e530683c18c8d5230300f).

PR #174, opened 2017-12-04
add rokka/client-bundle to flex

chregu:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/rokka/client-bundle/1.0/config/packages/rokka.yaml
+++ b/rokka/client-bundle/1.0/config/packages/rokka.yaml
@@ -1,6 +1,6 @@
 rokka_client:
-    # Api-Key value given on registration of the user. Get your API key form https://rokka.io/
-    api_key:       "%env(ROKKA_API_KEY)%" # Required
+    # Api-Key value given on registration of the user. Get your API key from https://rokka.io/
+    api_key:      "%env(ROKKA_API_KEY)%" # Required
     # Organization value like "my-org" for use in calls to the API.
     organization: "%env(ROKKA_ORGANIZATION)%" # Required
 
--- a/rokka/client-bundle/1.0/config/packages/rokka.yaml
+++ b/rokka/client-bundle/1.0/config/packages/rokka.yaml
@@ -1,0 +1,8 @@
+rokka_client:
+    # Api-Key value given on registration of the user.
+    api_key:       "%env(ROKKA_API_KEY)%" # Required
+    # Api-secret value from user registration.
+    api_secret:    "%env(ROKKA_API_SECRET)%" # Required
+    # Organization value like "my-org" for use in calls to the API.
+    organization: "%env(ROKKA_ORGANIZATION)%" # Required
+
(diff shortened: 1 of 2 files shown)


rokka/client-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "\\Rokka\\RokkaClientBundle\\RokkaClientBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#1": "Get your API key form https://rokka.io/",
```
dbu: form => from ?

rokka/client-bundle/1.0/config/packages/rokka.yaml
```diff
@@ -0,0 +1,6 @@
+rokka_client:
+    # Api-Key value given on registration of the user. Get your API key form https://rokka.io/
```
dbu: form => from

rokka/client-bundle/1.0/config/packages/rokka.yaml
```diff
@@ -0,0 +1,6 @@
...
+    api_key:       "%env(ROKKA_API_KEY)%" # Required
+    # Organization value like "my-org" for use in calls to the API.
+    organization: "%env(ROKKA_ORGANIZATION)%" # Required
```
dbu: @chregu is the organization not optional? i thought the bundle works without organziation and then does e.g. not register the commands that only make sense when you have an organization. that way, you can create an organization with the commands, after installing the bundle with your api key.

rokka/client-bundle/1.0/config/packages/rokka.yaml
chregu: fixed

rokka/client-bundle/1.0/config/packages/rokka.yaml
chregu: Not in the bundle. It's marked as required (since ages ;)). Also for ease of use, it's good to have it there.

rokka/client-bundle/1.0/manifest.json
chregu: fixed

rokka/client-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "\\Rokka\\RokkaClientBundle\\RokkaClientBundle": ["all"]
```
Tobion: the leading slash is not neeeded and not used in the other recipes

stof: @fabpot should we enforce it in the validator ?

chregu: fixed that one. And yes, would be great, if the validator would catch that

PR #180, opened 2017-12-05
[LiipImagineBundle] Set default web directory path to public/

michaelperrin:
| Q             | A
| ------------- | ---
| License       | MIT

LiipImagineBundle doesn't work out of the box with Flex because the web directory is configured to use the `web/` directory by default in the bundle, while Flex directory structure uses `public/`.

This PR adds configuration to use the correct folder.

### DIFF
--- a/liip/imagine-bundle/1.8/config/packages/imagine.yaml
+++ b/liip/imagine-bundle/1.8/config/packages/imagine.yaml
@@ -1,4 +1,15 @@
-#liip_imagine:
+liip_imagine:
+    resolvers:
+        default:
+            web_path:
+                web_root: '%kernel.project_dir%/public'
+                cache_prefix: ~ # media/cache
+
+    loaders:
+        default:
+            filesystem:
+                data_root: '%kernel.project_dir%/public'
+#
 #    # valid drivers options include "gd" or "gmagick" or "imagick"
 #    driver: "gd"
 #


fabpot:
It would be great if the default could be auto-configured in the bundle itself.

michaelperrin:
The problem is that the same bundle version can be installed in a non-Flex Symfony 3.x install, using the `web/` dir instead.

fabpot:
I understand that, but guessing the right one should be easy enough, no?

michaelperrin:
I actually opened the PR on @stof proposal in the _symfony-flex_ channel on Slack.
Should I update LiipImagineBundle to detect whether the public folder exists or not and adapt the configuration on this? I am not sure whether it is a reliable way to check for a flex or non-flex setup.

Nyholm:
> Should I update LiipImagineBundle to detect whether the public folder exists or not and adapt the bundle's configuration depending on this?

Yes, the ideal way would be to have a configuration option for where the folder is. Default would be "`public` if that folder exists, otherwise `web`".

> I am not sure whether it is a reliable way to check for a flex or non-flex setup.

Lets not think of it as "flex or non-flex". Think of it as "where is the public folder?" =)

michaelperrin:
I will probably open a PR on LiipImagineBundle to implement this then. That would be a BC break (in case some people had a `public` dir in their project besides Symfony’s `web` one) but that should be fine as far as I make the PR in the upcoming 2.0 branch.

Feel free to close this PR then!

PR #184, opened 2017-12-09
[WIP] introduce missing cmf bundles

ElectricMaxxx:
| Q             | A
| ------------- | ---
| License       | MIT

By this PR i'd like to introduce the following cmf bundles:
- core-bundle
- block-buncle
- content-bundle
- menu-bundle
- resource-bundle
- sonata-phpcr-admin-integration-bundle
- routing-auto-bundle

ToDo:
* [ ] discuss with @dbu, @lsmith77 @wouterj  about a configuration to support user
* [ ] have a valid and easy configuration on https://github.com/symfony/recipes-contrib/pull/185


ElectricMaxxx:
@dbu you got some more ideas to put here?

Added the types to the bundles, where it was missing.

Nyholm:
Thank you for this PR. 

I see that you've added lots of bundles and no config. Flex is smart enough to be able to install these bundles without a recipe. So this PR does not really bring any value.

ElectricMaxxx:
Thanks @Nyholm 
i will wait now for @dbu to discuss what a nice configuration is, which isn't default at all, but should be set. From my POV, we do have a little step to go for users, when using CMF for the first time and i want to  make that step as smal as possible. That is the same as in https://github.com/symfony/recipes-contrib/pull/185

dbu:
what about having a single recipe for installing the whole cmf with phpcr-odm? basically what symfony-cmf/symfony-cmf does? if the individual bundles are installable without any configuration, we do indeed not need recipies.

wouterj:
I would not add these recipes indeed, as there is no configuration added by it. All these seperate bundles can also be configured by just configuring them in `cmf_core`.

It would be nice to have these `symfony-cmf/phpcr-odm-pack` things, which configures DoctrinePHPCRBundle + CoreBundle (allowing you to flex in the other CMF bundles, which will be configured automatically by the CoreBundle). However, let's wait a bit for some news from the Symfony core team, as packs will be revisited soon. :)

Nyholm:
What is the status of this PR now? Are we going for a pack?

wouterj:
I would say so.

ElectricMaxxx:
We discuss now about the packs: https://github.com/symfony-cmf/symfony-cmf/issues/259

PR #181, opened 2017-12-05
Added recipe for DoctrinePHPCRBundle

giovannialbero1992:
| Q             | A
| ------------- | ---
| License       | MIT

PR #185, opened 2017-12-09
add doctrine phpcr-bundle

ElectricMaxxx:
| Q             | A
| ------------- | ---
| License       | MIT

This one will add a reciepe for doctrine phpcr-bundle.

### DIFF
--- a/doctrine/phpcr-bundle/2.0/config/packages/doctrine_phpcr.yaml
+++ b/doctrine/phpcr-bundle/2.0/config/packages/doctrine_phpcr.yaml
@@ -1,0 +1,13 @@
+# Read the documentation: https://symfony.com/doc/2.0/cmf/bundles/phpcr_odm/configuration.html
+# Configure a doctrine dbal default connection before
+#doctrine_phpcr:
+#    odm:
+#        auto_mapping: true
+#        auto_generate_proxy_classes: "%kernel.debug%"
+#    session:
+#        backend:
+#            type: doctrinedbal
+#            connection: default
+#        workspace: default
+#        username: '%env(PHPCR_SESSION_USERNAME)%'
+#        password: '%env(PHPCR_SESSION_PASSWORD)%'
--- a/doctrine/phpcr-bundle/2.0/manifest.json
+++ b/doctrine/phpcr-bundle/2.0/manifest.json
@@ -1,0 +1,13 @@
+{
+    "bundles": {
+        "Doctrine\\Bundle\\PHPCRBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "src/": "%SRC_DIR%/"
+    },
+    "env": {
+        "PHPCR_SESSION_USERNAME": "admin",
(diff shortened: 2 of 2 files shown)


ElectricMaxxx:
@dbu you got some input here?

doctrine/phpcr-bundle/2.0/config/packages/doctrine_phpcr.yaml
```diff
@@ -0,0 +1,13 @@
...
+#    odm:
+#        auto_mapping: true
+#        auto_generate_proxy_classes: "%kernel.debug%"
+#    session:
+#        backend:
+#            type: doctrinedbal
+#            connection: default
+#        workspace: default
+#        username: '%env(PHPCR_SESSION_USERNAME)%'
+#        password: '%env(PHPCR_SESSION_PASSWORD)%'
```
fabpot: We do not copy/paste the default configuration as it means syncing it whenever something change upstream. Instead, we only list the ones which needs to be changed by the end user. If there is none, then the file can be removed.

ElectricMaxxx: Mh, ok sounds valid. But then i would have a question/feature request: is it possible to use env vars as default on configuration directly? I mean would `->defaultValue( '%env(PHPCR_SESSION_USERNAME)%')` work? Or shouldn't we have an extra `defaultEnvironmentValue('PHPCR_SESSION_USERNAME', optonalCast)` for that?

Nyholm: The bundle should not handle environment variables. The bundle should only care about its config. 

When defining your config in `Configuration.php` you should use `->defaultValue('foobar')`. When the user is using the config in `config.yaml` they might want to use `%env(FOO)%`.

ElectricMaxxx: So providing a configuration for both credential parts (username, password) set by envirionment variables here would be a good example?
Then i would remove the other configuration parts. But i think i would keep the decission for `doctrinedbal` and add a hint, that there is the place to switch to an other backend, and that dbal would need doctrine dbal configuration. Doing so we would have one working exampe then.

dbu:
if i understand correctly, flex basically makes the doctrine/phpcr-bundle uninstallable on its own. it will automatically enable the bundle, but the bundle can't run without a configured transport. but for the transport, we need a decision between jackalope-jackrabbit and jackalope-doctrine-dbal. that decision is another composer dependency.

we could provide two recipies for phpcr-odm with jackalope-jackrabbit and with dbal. for that we would have to create a repository for each of those with a composer.json that requires the phpcr bundle and the right jackalope version. then the recipies for those can also provide meaningful defaults.

if there is an easier solution it would be great. so if anybody has one, please tell!

Nyholm:
For the record: https://github.com/symfony-cmf/symfony-cmf/issues/259

maxhelias:
Same as #181

Nyholm:
Closing due inactivity. Feel free to comment or reopen. 

#eu-fossa

PR #186, opened 2017-12-09
Add DtcGridBundle

mmucklo:
| Q             | A
| ------------- | ---
| License       | MIT

This pull request adds a recipe for DtcGridBundle which has been updated to support symfony 4


mmucklo:
Thanks for testing it out,

The problem is that __config/packages/framework.yaml__ needs to have the twig templating engine enabled (as this bundle relies on it).  Is it standard practice to modify framework.yaml, or should I just work around it using a compiler pass and setter injection?

__config/packages/framework.yaml:__
framework:
    # ...
    templating:
        engines: ['twig']

PR #190, opened 2017-12-10
Adding mmucklo/queue-bundle

mmucklo:
| Q             | A
| ------------- | ---
| License       | MIT

Adding mmucklo/queue-bundle

### DIFF
--- a/mmucklo/queue-bundle/3.0/config/packages/dtc_queue.yaml
+++ b/mmucklo/queue-bundle/3.0/config/packages/dtc_queue.yaml
@@ -1,0 +1,9 @@
+dtc_queue:
+    default_manager: odm
+    record_timings: false
+    document_manager: default
+    entity_manager: default
+    priority_max: 255
+    priority_direction: desc
+    admin:
+        chartjs: 'https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.7.1/Chart.bundle.min.js'
--- a/mmucklo/queue-bundle/3.0/config/routes/dtc_queue.yaml
+++ b/mmucklo/queue-bundle/3.0/config/routes/dtc_queue.yaml
@@ -1,0 +1,2 @@
+dtc_queue:
+    resource: '@DtcQueueBundle/Resources/config/routing.yml'
--- a/mmucklo/queue-bundle/3.0/manifest.json
+++ b/mmucklo/queue-bundle/3.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Dtc\\QueueBundle\\DtcQueueBundle": ["all"]
+    },
+    "copy-from-recipe": {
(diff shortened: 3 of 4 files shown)


Nyholm:
Thank you for this. I think it looks good... or the code looks good. When installing this I got the following: 

screen shot 2017-12-12 at 09 20 02

Maybe not using the black background with blue text?

javiereguiluz:
@Nyholm do you agree with not fully using automation and putting everything on "What's Next" section? For example, why not create `config/routes/dtc_queue.yaml` instead of displaying this:

```
 +  * Then <fg=blue;bg=black>add</> the following to config/routes.yaml:
 +
 +<fg=blue;bg=black>dtc_grid:</>
 +    <fg=blue;bg=black>resource: '@DtcGridBundle/Resources/config/routing.yml'</>
 +<fg=blue;bg=black>dtc_queue:</>
 +    <fg=blue;bg=black>resource: '@DtcQueueBundle/Resources/config/routing.yml'</>
```

Same with the Twig config check, etc.  That shouldn't be part of the "What's Next". Add some check somewhere and display a proper error message.

And about the custom colors in "What's Next" ... even if this is recipes-contrib, I'd ban any use of colors to create a consistent experience for users.

mmucklo:
Didn’t know I could do the config/routes/ thing - are there some docs on that?

one caution is that these are admin routes which a prudent user might want to add a firewall rule around should I force the installation of the security bundle and try to auto add a firewall rule?  I suppose I could do it with checks inside of each route and a global rule to turn on or off a sort of local firewall for just my bundle (but it would still require the symfony security bundle to be installed)

The twig config check needs to go into config/packages/framework.yaml - am I permitted to edit that myself?  Seems kind of dangerous.  The bundle can be used without twig, but not the admin section (which is optional).

I can fix the coloring.

javiereguiluz:
Here are some examples of creating routing files:

* https://github.com/symfony/recipes/blob/master/javiereguiluz/easyadmin-bundle/1.17/config/routes/easy_admin.yaml
* https://github.com/symfony/recipes/blob/master/willdurand/js-translation-bundle/2.0/config/routes/js_translation.yaml

About the firewall config, we can't change that. You could define a prefix like "/admin/queue/" for the imported routes, or you could add a brief message in What's Next.

And about Twig dependency, not sure what's the best practice here. Tobias knows better what other recipes do. Maybe just add Twig as a dependency?

Nyholm:
> For example, why not create config/routes/dtc_queue.yaml instead of displaying this....

I agree, we should try to automate everything. I just did a quick review and a test run. 

> Colors stuff..

Sure. I'll make sure to enforce that. No colors it is =)

> Twig

Other bundles have `twig/twig` as dependency. No recipe in contrib touches the `framework` settings. I think that could be in "What's next"

PR #196, opened 2017-12-11
add recipes for Broadway

othillo:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/broadway/event-store-dbal/0.2/config/packages/broadway_event_store_dbal.yaml
+++ b/broadway/event-store-dbal/0.2/config/packages/broadway_event_store_dbal.yaml
@@ -1,0 +1,13 @@
+broadway:
+    event_store: broadway.event_store.dbal
+
+services:
+    broadway.event_store.dbal:
+        class: Broadway\EventStore\Dbal\DBALEventStore
+        arguments:
+            - "@doctrine.dbal.default_connection"
+            - "@broadway.serializer.payload"
+            - "@broadway.serializer.metadata"
+            - "events"
+            - false
+            - "@broadway.uuid.converter"
--- a/broadway/event-store-mongodb/0.2/manifest.json
+++ b/broadway/event-store-mongodb/0.2/manifest.json
@@ -3,7 +3,11 @@
         "config/": "%CONFIG_DIR%/"
     },
     "container": {
-        "mongodb_uri": "mongodb://127.0.0.1/",
-        "mongodb_database": "default"
+        "mongodb_uri": "%env(BROADWAY_EVENT_STORE_MONGODB_URI)%",
+        "mongodb_database": "%env(BROADWAY_EVENT_STORE_MONGODB_DATABASE)%"
+    },
+    "env": {
+        "BROADWAY_EVENT_STORE_MONGODB_URI": "mongodb://127.0.0.1/",
+        "BROADWAY_EVENT_STORE_MONGODB_DATABASE": "default"
     }
 }
--- a/broadway/event-store-mongodb/0.2/manifest.json
+++ b/broadway/event-store-mongodb/0.2/manifest.json
@@ -1,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "container": {
+        "mongodb_uri": "mongodb://127.0.0.1/",
+        "mongodb_database": "default"
+    }
+}
--- a/broadway/read-model-elasticsearch/0.3/config/packages/broadway_read_model_elasticsearch.yaml
+++ b/broadway/read-model-elasticsearch/0.3/config/packages/broadway_read_model_elasticsearch.yaml
@@ -1,0 +1,19 @@
+broadway:
+    read_model: broadway.read_model.elasticsearch.repository_factory
+
+services:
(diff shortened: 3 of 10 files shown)


broadway/event-store-dbal/0.1/config/packages/broadway_event_store_dbal.yaml
```diff
@@ -1,7 +0,0 @@
-broadway:
```
Nyholm: There is no need to remove old recipes.

broadway/event-store-dbal/0.2/config/packages/broadway_event_store_dbal.yaml
```diff
@@ -0,0 +1,14 @@
+broadway:
+    event_store: broadway.event_store.dbal
+
+services:
+    broadway.event_store.dbal:
+        class: Broadway\EventStore\Dbal\DBALEventStore
+        public: true
```
Nyholm: Do you really want this to be public by default?

broadway/event-store-dbal/0.2/config/packages/broadway_event_store_dbal.yaml
```diff
@@ -0,0 +1,14 @@
+broadway:
+    event_store: broadway.event_store.dbal
+
+services:
+    broadway.event_store.dbal:
```
Nyholm: Let's use `Broadway\EventStore\Dbal\DBALEventStore` as service id.

broadway/read-model-mongodb/0.2/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "container": {
+        "mongodb_host": "localhost",
+        "mongodb_port": "27107",
```
Nyholm: These parameters build up a DSN. We should consider using the DSN here directly. That would reduce the number of parameters.

othillo:
I'm a little confused about the FQCN usage. I understand it helps auto-wiring but then again Symfony suggests public bundles not to rely on auto-wiring (https://symfony.com/doc/current/service_container/autowiring.html#public-and-reusable-bundles). So for now I don't see the benefit, but, please, correct me if I'm wrong.

nicolas-grekas:
A service should get FQCN id only if it should be autowireable. That's the rule.

Nyholm:
Thank you. You are both correct. I thought one should always use FQCN no matter auto-wiring.

othillo:
@Nyholm I've done the changes you requested. Can you review it again?

broadway/event-store-mongodb/0.2/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "container": {
+        "mongodb_uri": "mongodb://127.0.0.1/",
+        "mongodb_database": "default"
```
Nyholm: I assume these will be different on different environments. Lets put them as environment variables: https://github.com/symfony/recipes#env-configurator

broadway/read-model-mongodb/0.2/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "container": {
+        "mongodb_uri": "mongodb://127.0.0.1/",
+        "mongodb_database": "default"
```
Nyholm: Same here. Consider using environment variables. 

Im not sure what happends when you reuse environment variables. Lets try =)

othillo:
I introduced environment variables for MongoDB. The Travis builds are failing but I don't understand why.

broadway/read-model-elasticsearch/0.3/config/packages/broadway_read_model_elasticsearch.yaml
```diff
@@ -0,0 +1,18 @@
...
+        class: Broadway\ReadModel\ElasticSearch\ElasticSearchRepositoryFactory
+        arguments:
+            - "@broadway.elasticsearch.client"
+            - "@broadway.serializer.readmodel"
+
+    broadway.elasticsearch.client:
+        class: Elasticsearch\Client
+        factory: ['@broadway.elasticsearch.client_factory', create]
+        arguments:
+            - "%elasticsearch%"
```
Nyholm: This parameter is undefined. You should consider put it as an environment variable.

Nyholm: Ping. Could you fix this and the PR is all good.

othillo: Unfortunately this parameter is a nullable array so we cannot use environment variables here. I declared it and set it to null by default.

othillo: @Nyholm are you ok with this?

Nyholm:
Excellent. Thank you

PR #198, opened 2017-12-12
Garlic Message Bus recipes

imaximius:
Q | A
-- | --
License | MIT


garlic/bus/1.0/config/packages/garlic.yaml
```diff
@@ -0,0 +1,13 @@
+enqueue:
```
Pierstoval: This corresponds to another package than Garlic, you shouldn't post this config here, as there is another one for Enqueue.

imaximius: Thanx! It just mistake. I'll change it

imaximius: @Pierstoval Already changed!

imaximius:
Bundle has the command that up and running consumers and producers to the message queue. I thought it'll be comfortable already have supervisor config for this purposes

Nyholm:
I do understand why you need supervisor. Im just cant figure out why you want that config in `my-symfony-project/config/`. Should that config be in `/etc/supervisor/config` or something like that?

garlic/bus/1.0/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+    "bundles": {
+        "Garlic\\Bus\\GarlicBusBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "RABBIT_HOST": "rabbit",
```
Nyholm: Where are these environment variables used?

PR #202, opened 2017-12-13
QandidateToggleBundle: do not enable SecurityBundle

othillo:
| Q             | A
| ------------- | ---
| License       | MIT

I thought it would be a good idea to also enable the SecurityBundle because it's a dependency for the QandidateToggleBundle. However when removing the QandidateToggleBundle this recipe also unregisters the SecurityBundle.


stof:
This enabling was indeed wrong. SecurityBundle was enabled by Flex thanks to the SecurityBundle recipe (which is also applied as it got installed).

othillo:
Thank you, @stof !

PR #203, opened 2017-12-13
Wisembly amqp

Taluu:
| Q             | A
| ------------- | ---
| License       | MIT

Update the flex recipe for Wisembly AmqpBundle (1.4 and 2.0 versions which were released today).

I wasn't sure if duplicating (as 1.4 and 2.0 have the same config), but I did it anyway for consistency.

### DIFF
--- a/wisembly/amqp-bundle/1.4/config/packages/wisembly_amqp.yaml
+++ b/wisembly/amqp-bundle/1.4/config/packages/wisembly_amqp.yaml
@@ -1,0 +1,16 @@
+wisembly_amqp:
+    console_path: '%kernel.project_dir%/bin/console'
+    broker: 'Wisembly\AmqpBundle\Broker\PhpAmqpLibBroker'
+    connections:
+        default: '%env(AMQP_URL)%'
+    gates:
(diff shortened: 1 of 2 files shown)


stof:
> I wasn't sure if duplicating (as 1.4 and 2.0 have the same config), but I did it anyway for consistency.

if they have the same config, there is no need to duplicate. A recipe versioned as 1.4 applies to `>= 1.4` (upper bound is only set by the next version of the recipe overriding it)

Taluu:
That's what I thought. I'll remove the 2.0 recipe then.

wisembly/amqp-bundle/1.4/config/packages/wisembly_amqp.yaml
```diff
@@ -0,0 +1,16 @@
+wisembly_amqp:
+    console_path: '%kernel.project_dir%/bin/console'
+    broker: 'Wisembly\AmqpBundle\Broker\PhpAmqpLibBroker'
```
Taluu: The only change from `1.3` is this, as using the alias `oldsound` is deprecated.

PR #204, opened 2017-12-13
Adds support for maba/webpack-bundle

mariusbalcytis:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/maba/webpack-bundle/1.0/manifest.json
+++ b/maba/webpack-bundle/1.0/manifest.json
@@ -1,0 +1,15 @@
+{
+    "copy-from-recipe": {
+        "assets/": "assets/"
+    },
+    "copy-from-package": {
+        "src/Resources/defaultConfig/latest/package.json": "package.json",
+        "src/Resources/defaultConfig/latest/webpack.config.js": "%CONFIG_DIR%/webpack.config.js"
+    },
+    "gitignore": [
+        "/node_modules/"
+    ],
+    "composer-scripts": {
+        "maba:webpack:install-node-modules": "symfony-cmd"
+    }
+}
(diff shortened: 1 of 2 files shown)


mariusbalcytis:
To test if bundle installation is working correctly:

```
composer create-project symfony/skeleton
cd skeleton
SYMFONY_ENDPOINT=https://symfony.sh/r/github.com/symfony/recipes-contrib/204 composer req maba/webpack-bundle
echo "index: {path: /, defaults: {_controller: 'Symfony\Bundle\FrameworkBundle\Controller\TemplateController:templateAction', template: 'base.html.twig'}}" > config/routes.yaml
echo '<script src="{{ webpack_asset("@assets/index.js") }}"></script>' >> templates/base.html.twig
echo '<link rel="stylesheet" href="{{ webpack_asset("@assets/index.js", "css") }}"/>' >> templates/base.html.twig
echo "body {background: green}" > assets/styles.css
echo "require('./styles.css'); alert('hello');" > assets/index.js
bin/console maba:webpack:compile
php -S 127.0.0.1:8000 -t public
```

You must have `npm` or `yarn` installed beforehand for this to work.

Instead of `bin/console maba:webpack:compile` one can run `bin/console maba:webpack:dev-server` to run webpack dev-server. It compiles assets on the fly, so after changing `index.js` file it will reload automatically.

It copies needed files (webpack config and `package.json`) and installs JavaScript packages always on composer install. Wrapper is used for installing packages as it chooses `yarn` or `npm` depending on the system.

Test itself adds webpack entry inside twig file, configures route for testing, adds JS and CSS files to test with.

When opened, browser will show green screen and alert hello message.

PR #205, opened 2017-12-14
Add Enqueue job queue pack

makasim:
| Q             | A
| ------------- | ---
| License       | MIT

A pack to install Enqueue's job queue with Doctrine ORM as a storage in a Symfony Flex application

### DIFF
--- a/enqueue/job-queue-pack/0.8/manifest.json
+++ b/enqueue/job-queue-pack/0.8/manifest.json
@@ -1,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
+    }
+}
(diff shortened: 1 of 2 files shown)


makasim:
@symfony-flex-server please review.

fabpot:
@symfony-flex-server review please (@makasim that's the right way :)).

makasim:
@fabpot I still miss something. Why hasn't it been merged?

enqueue/job-queue-pack/0.8/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "etc/": "%ETC_DIR%/"
```
msheakoski: This should be `"config/": "%CONFIG_DIR%/"` and the `etc/packages` dir in the recipe needs to be renamed to `config/packages`

PR #209, opened 2017-12-17
add recipe for liip/theme-bundle:1.5.*

lsmith77:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/liip/theme-bundle/1.5/config/packages/framework.yaml
+++ b/liip/theme-bundle/1.5/config/packages/framework.yaml
@@ -1,0 +1,3 @@
+framework:
+    templating:
+        engines: ['twig']
--- a/liip/theme-bundle/1.5/config/packages/theme.yaml
+++ b/liip/theme-bundle/1.5/config/packages/theme.yaml
@@ -1,0 +1,4 @@
+liip_theme:
+    themes: ['default']
+    active_theme: 'default'
+    load_controllers: false
--- a/liip/theme-bundle/1.5/post-install.txt
+++ b/liip/theme-bundle/1.5/post-install.txt
@@ -1,0 +1,6 @@
+  <bg=blue;fg=white>         Getting started using </><bg=blue;fg=white;options=bold>liip/theme-bundle</><bg=blue;fg=white>          </>
+
+  <options=bold>Configure</> the bundle and routes:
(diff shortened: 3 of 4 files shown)


lsmith77:
now can I ensure that `templating` is also enabled?

liip/theme-bundle/1.5/config/packages/theme.yaml
```diff
@@ -0,0 +1,4 @@
+liip_theme:
+    themes: ['default']
+    active_theme: 'default'
+    load_controllers: false
```
lsmith77: ok .. that doesn’t seem to do the trick :-/

liip/theme-bundle/1.5/config/packages/framework.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+    templating:
+        engines: ['twig']
```
lsmith77: ok .. that doesn’t seem to do the trick :-/

lsmith77: would it work if I fold it into the theme.yaml?

liip/theme-bundle/1.5/post-install.txt
```diff
@@ -0,0 +1,6 @@
+  <bg=blue;fg=white>         Getting started using </><bg=blue;fg=white;options=bold>liip/theme-bundle</><bg=blue;fg=white>          </>
+
+  <options=bold>Configure</> the bundle and routes:
+    1. You <options=bold>MUST</> verify the configuration in <comment>%CONFIG_DIR%/packages/theme.yaml</>.
+    2. You <options=bold>MAY</> configure additional features, see the README.md for instructions.
+    3. You <options=bold>MAY</> configure the routes as specified in the README.md if you enable the controller.
```
Taluu: You should remove the post-install

liip/theme-bundle/1.5/config/packages/framework.yaml
fabpot: Using the Symfony Templating layer as done here is "forbidden" in Symfony 3.4/4.0. Supporting Twig directly is what is epxected in recipes.

lsmith77: ok .. which requires a major refactoring of this Bundle .. which I am currently not planning to work on (see https://github.com/liip/LiipThemeBundle/issues/181).

so I guess I will close this PR

PR #210, opened 2017-12-17
[WIP] Add recipe for friendsofsymfony/user-bundle

Devtronic:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/friendsofsymfony/user-bundle/2.0/config/packages/fos_user.yaml
+++ b/friendsofsymfony/user-bundle/2.0/config/packages/fos_user.yaml
@@ -1,0 +1,8 @@
+# Read the documentation: https://symfony.com/doc/master/bundles/FOSRestBundle/index.html
+fos_user: ~
+#    db_driver: orm # other valid values are 'mongodb' and 'couchdb'
+#    firewall_name: main
+#    user_class: AppBundle\Entity\User
+#    from_email:
+#        address: "%env(MAILER_USER)%"
+#        sender_name: "%env(MAILER_USER)%"
--- a/friendsofsymfony/user-bundle/2.0/config/packages/fos_user.yaml
+++ b/friendsofsymfony/user-bundle/2.0/config/packages/fos_user.yaml
@@ -6,3 +6,7 @@ fos_user:
     from_email:
         address: "%env(MAILER_USER)%"
         sender_name: "%env(MAILER_USER)%"
+
+services:
+    validator.builder:
+        class: Symfony\Component\Config\Definition\Builder\ValidationBuilder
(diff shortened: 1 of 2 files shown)


friendsofsymfony/user-bundle/2.0/config/packages/fos_user.yaml
```diff
@@ -0,0 +1,8 @@
+# Read the documentation: https://symfony.com/doc/master/bundles/FOSUserBundle/index.html
+fos_user:
+#    db_driver: orm # other valid values are 'mongodb' and 'couchdb'
+    firewall_name: main
```
Pierstoval: Instead, it's better IMO to directly uncomment `db_driver`, `firewall_name` and `user_class` in the recipe as they are all mandatory anyway in the bundle.

Devtronic: Thanks, you're right. I changed the config in e129d94

friendsofsymfony/user-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "bundles": {
+        "FOS\\UserBundle\\FOSUserBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "MAILER_USER": ""
```
Pierstoval: I think `MAILER_USER` is inconsistent with possible configuration value. I think `MAILER_SENDER` would probably be more appropriate. A "user" option is more to configure the mailer with a user.

friendsofsymfony/user-bundle/2.0/config/packages/fos_user.yaml
```diff
@@ -0,0 +1,12 @@
...
+fos_user:
+    db_driver: orm # other valid values are 'mongodb' and 'couchdb'
+    firewall_name: main
+    user_class: AppBundle\Entity\User
+    from_email:
+        address: "%env(MAILER_USER)%"
+        sender_name: "%env(MAILER_USER)%"
+
+services:
+    validator.builder:
```
Pierstoval: That's absolutely not how the recipe should work. Instead, the package itself should require the Symfony package that installs this service, which is `symfony/validator` in this case.

This is a problem from the FOSUserBundle side that should be fixed before a recipe can be created.

Devtronic: uh... I understand, sorry.

friendsofsymfony/user-bundle/2.0/manifest.json
fabpot: I agree

friendsofsymfony/user-bundle/2.0/config/packages/fos_user.yaml
Nyholm: This service should be removed. You use add `symfony/validator` to the dependencies section in your composer.json.

Nyholm:
FYI. This is a duplicate of #270.

PR #211, opened 2017-12-18
Add Leapt ImBundle

jmsche:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/leapt/im-bundle/2.0/config/packages/leapt_im.yaml
+++ b/leapt/im-bundle/2.0/config/packages/leapt_im.yaml
@@ -1,0 +1,2 @@
+leapt_im:
+    cache_path: cache/im
--- a/leapt/im-bundle/2.0/config/packages/leapt_im.yaml
+++ b/leapt/im-bundle/2.0/config/packages/leapt_im.yaml
@@ -1,6 +1,7 @@
+# Read the documentation: https://github.com/leapt/im-bundle/blob/master/docs/configuration.rst
 leapt_im:
     cache_path: cache/im
-    # You can define reusable formats here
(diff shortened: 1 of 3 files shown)


leapt/im-bundle/2.0/config/packages/leapt_im.yaml
```diff
@@ -0,0 +1,6 @@
+leapt_im:
```
Nyholm: It could be a god idea to add a link to https://github.com/leapt/im-bundle/blob/master/docs/configuration.rst

leapt/im-bundle/2.0/config/packages/leapt_im.yaml
```diff
@@ -0,0 +1,6 @@
...
+    cache_path: cache/im
+    # You can define reusable formats here
+    formats:
+        #my_format:
+        #    thumbnail: 350x350^
```
Nyholm: Could we add something useful here instead of the comments?

Like: 
```
  formats:
    # resizes to fit in 100x100, only if bigger, and remove all crap (meta-data, ...)
    medium:
      thumbnail: 100x100>
```

leapt/im-bundle/2.0/config/packages/leapt_im.yaml
jmsche: Done

leapt/im-bundle/2.0/config/packages/leapt_im.yaml
jmsche: Nice idea :)

PR #212, opened 2017-12-19
[enqueue] Change etc dir to config.

makasim:
| Q             | A
| ------------- | ---
| License       | MIT

fix comment https://github.com/symfony/recipes-contrib/pull/205#discussion_r157673819


makasim:

nicolas-grekas:
don't you need also to rename the etc/ directory in the recipe?

makasim:
> don't you need also to rename the etc/ directory in the recipe?

that's true. fixed

PR #214, opened 2017-12-21
Add koff/i18n-form-bundle recipe

sadikoff:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Great thanks. 

Im looking at the config. Is there any config that 90% of the users will use? If so, I would be happy to include that and remove those comments. Maybe we want the `locales` to not be a comment? What do you think?

sadikoff:
I think we can use default configuration with single language selected, may be 'en'

PR #219, opened 2017-12-28
Added MsgPhp bundles

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

I tweaked the bundles to be installed without any errors. This should be the final step.

Working app: https://github.com/msgphp/symfony-demo-app
Blog post: https://medium.com/@ro0NL/building-a-new-symfony-user-bundle-b4fe5a9d9d80

It's a work in progress due the fact the used packages have no stable tag yet. 

For now;

```bash
$ composer create-project symfony/skeleton
$ cd skeleton
$ composer require msgphp/domain:@dev msgphp/eav:@dev msgphp/user:@dev msgphp/eav-bundle msgphp/user-bundle
```

### DIFF
--- a/msgphp/eav-bundle/0.2/config/packages/msgphp_eav.php
+++ b/msgphp/eav-bundle/0.2/config/packages/msgphp_eav.php
@@ -1,0 +1,13 @@
+<?php
+
+use MsgPhp\Eav\Entity\{Attribute, AttributeValue};
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return function (ContainerConfigurator $container) {
+    $container->extension('msgphp_eav', [
+        'class_mapping' => [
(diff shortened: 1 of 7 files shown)


ro0NL:
@Nyholm wow thanks :D one thing; i forced push the bundle repositories yesterday. They now live in the same monolith repo as for the components; https://github.com/msgphp/msgphp/tree/master/src/UserBundle

composer/PSR namespaces are all good. But the initial tags (0.x) are removed (all master now).

Does it mean for the recipe folder structure i can go with `/1.0/`? the current branch alias for all.

Nyholm:
Im not sure to be honest. Why dont you tag the current master as 0.3? That would make the recipes work

ro0NL:
Right.. tagging the monolith repo means all components are tagged too. Thats a new step for me :) but the next one yes.

As there are not tags today it be 0.1, thus `/0.1` folder. Ill do the PR ASAP.

PR #223, opened 2018-01-01
Remove endroid/import

endroid:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Thank you for the PR. 

Could you tell me why you want this to be removed?

endroid:
Hi @Nyholm. For the same reasons as the other two I dropped, see https://github.com/symfony/recipes-contrib/pull/168. I will create a separate import-bundle and I even think this bundle will not require any default configuration so I only removed this one. If the bundle requires configuration in the future I will add it under the name import-bundle.

Nyholm:
Wow, could you please make separate PRs for all your packages? Could you also make sure that this PR only "Remove endroid/import" which the title says. 

It is a bit hard to review when there is mixed features.

endroid:
Yes of course :) I will create a PR per bundle.

PR #230, opened 2018-01-03
Add pusher/pusher-php-server

endroid:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Thank you. I've looked at the source of this bundle. Couldn't you write a recipe for `pusher/pusher-php-server` and you wouldnt need the bundle at all, right?

PR #232, opened 2018-01-03
Add endroid/twitter

endroid:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/endroid/twitter/2.0/config/packages/endroid_twitter.yaml
+++ b/endroid/twitter/2.0/config/packages/endroid_twitter.yaml
@@ -1,6 +1,5 @@
 services:
     Endroid\Twitter\Client:
-        public: true
         arguments:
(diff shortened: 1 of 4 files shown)


Nyholm:
Coudn't this bundle be replaced by a recipe for `endroid/twitter`?

endroid:
Good point, I didn't think of the possibility of registering the service via the yaml file. I think since Flex there is no use for the bundle anymore.

endroid/twitter/2.0/config/packages/endroid_twitter.yaml
```diff
@@ -0,0 +1,8 @@
+services:
+    Endroid\Twitter\Client:
+        public: true
```
Nyholm: Are you sure?

PR #235, opened 2018-01-05
add SonataBlockBundle

aLkRicha:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sonata-project/block-bundle/3.9/config/packages/sonata_project/sonata_block.yaml
+++ b/sonata-project/block-bundle/3.9/config/packages/sonata_project/sonata_block.yaml
@@ -1,0 +1,2 @@
+sonata_block:
+    default_contexts: []
--- a/sonata-project/block-bundle/3.9/manifest.json
+++ b/sonata-project/block-bundle/3.9/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Sonata\\BlockBundle\\SonataBlockBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


sonata-project/block-bundle/3.9/config/packages/sonata_block.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+    templating:
+        engines: ['twig']
+sonata_block:
+    default_contexts: [cms]
```
covex-nn: `cms` context is never used anywhere, and it was renamed to `sonata_page_bundle` for SonataPageBundle (see https://stackoverflow.com/questions/22798670/what-is-context-in-sonata-block-bundle). So, i think `cms` should be removed from `default_contexts` here.

OskarStark:
//cc @greg0ire

aLkRicha:
i need a help with this:
```
Executing script cache:clear [KO]
 [KO]
Script cache:clear returned with error code 1
!!  
!!  In ContainerBuilder.php line 1018:
!!                                                       
!!    You have requested a non-existent service "twig".  
!!                                                       
!!  
!!  
```

it there another way then require twig-bundle in composer and enable twig service for framework?

sonata-project/block-bundle/3.9/config/packages/sonata_block.yaml
```diff
@@ -0,0 +1,2 @@
+sonata_block:
+    default_contexts: [sonata_page_bundle]
```
OskarStark: default_context or default_contextS?

Nyholm:
If the bundle depends on TwigBundle it should be in the requirements section of composer.json. Or handle the case where twig does not exit. 

It seams like the recipe is good but the bundles need to be updated.

sonata-project/block-bundle/3.9/config/packages/sonata_block.yaml
covex-nn: @OskarStark it is `default_contexts`. See [Configuration.php](https://github.com/sonata-project/SonataBlockBundle/blob/3.x/src/DependencyInjection/Configuration.php#L84). It is required, but if the value is left blank (`default_contexts: []`), there is no errors occured, at least when i use only SonataAdminBundle

May be `default_contexts` should not be required? `sonata_page_bundle` could be appended with recipe for SonataPageBundle!

aLkRicha:
@OskarStark @greg0ire so, we need to add TwigBundle to SonataBlockBundle?

greg0ire:
It probably should, since we are referencing it: https://github.com/sonata-project/SonataBlockBundle/search?utf8=%E2%9C%93&q=twigbundle&type=

aLkRicha:
@greg0ire ok, i'll create pull-request

OskarStark:
We release [3.92](https://github.com/sonata-project/SonataBlockBundle/releases/tag/3.9.2) of block-bundle

aLkRicha:
i think this will work after merging this https://github.com/symfony/recipes-contrib/pull/151

covex-nn:
@aLkRicha please move `config/packages/sonata_projects/sonata_block.yaml` to `config/packages` directory. And travis will be green =)

OskarStark:
I asked for a BlockBundle release @aLkRicha @covex-nn

aLkRicha:
@OskarStark @covex-nn i'm waiting this PR https://github.com/sonata-project/SonataBlockBundle/pull/457

covex-nn:
@OskarStark @aLkRicha since [3.10.0](https://github.com/sonata-project/SonataBlockBundle/releases/tag/3.10.0) there is no need to create a recipe for BlockBundle:

* Recipe is used only for new projects, it is used for initial configuration only
* Projects with BlockBundle < v3.10 are already configured
* New projects will use BlockBundle >= 3.10.0 with it's default configuration. See it for yourself:

    ```bash
    composer create-project symfony/skeleton .
    composer config extra.symfony.allow-contrib true
    composer require sonata-project/block-bundle
    ```

    There you will see error about `templating` service `The service "sonata.block.service.container" has a dependency on a non-existent service "templating"`. And this dependency must be removed.

But i can be wrong: may be we should create recipe for 3.9.*, what do you think?

covex-nn:
@OskarStark @aLkRicha https://github.com/sonata-project/SonataBlockBundle/pull/457 is merged and released with [3.11.0](https://github.com/sonata-project/SonataBlockBundle/releases/tag/3.11.0), but i'm still not sure if we need a recipe for BlockBundle

Nyholm:
So, should I close this?

jordisala1991:
Default context is no longer required, so many we don't need this recipe

jordisala1991:
And this one can be closed, it is not needed

PR #239, opened 2018-01-07
[WIP] Add psysh

theofidry:
| Q             | A
| ------------- | ---
| License       | MIT

Closes https://github.com/theofidry/PsyshBundle/issues/34


theofidry/psysh-bundle/3.3/config/packages/psysh.yaml
```diff
@@ -0,0 +1,6 @@
+psysh:
+    variables:
+        container: '@service_container'
+        kernel: '@kernel'
+        self: '@psysh.shell'
+        parameters: "@=service('service_container').getParameterBag().all()"
```
fabpot: If I understand correctly, those variables are always defined, right? If that's the case, they should be removed here.

theofidry: So the practice is to have those config files for non-default values?

theofidry:
Putting it as WIP as it seems there is some bits to rework for Symfony 4 :/

That said I would love a `psysh` alias, would it be possible to merge it in the official recipes instead as alias are not allowed in the contrib recipes?

Nyholm:
Can I close this PR for now?

PR #240, opened 2018-01-08
Add SonataCoreBundle with minimal configuration

aLkRicha:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sonata-project/core-bundle/3.9/config/packages/sonata_project/sonata_core.yaml
+++ b/sonata-project/core-bundle/3.9/config/packages/sonata_project/sonata_core.yaml
@@ -1,0 +1,5 @@
+sonata_core:
+    flashmessage: {}
+    form:
+        mapping:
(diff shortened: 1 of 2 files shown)


sonata-project/core-bundle/3.7/config/packages/sonata_project/sonata_core.yaml
```diff
@@ -0,0 +1,7 @@
+sonata_core:
+    flashmessage: {}
+    form:
+        mapping:
+            enabled: false
+framework:     # Remove this section
+    session: ~ # if you'll be using this in config/packages/framework.yaml
```
covex-nn: @aLkRicha session is enabled by default now, so you can remove `framework` configuration. See https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/3.3/config/packages/framework.yaml

PR #242, opened 2018-01-10
Add webinarium/symfony-lazysec package

webinarium:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/webinarium/symfony-lazysec/2.0/config/packages/lazysec.yaml
+++ b/webinarium/symfony-lazysec/2.0/config/packages/lazysec.yaml
@@ -1,8 +1,5 @@
 services:
-    _defaults:
-        autowire:      true
-        autoconfigure: true
-        public:        false
-
     # See https://github.com/webinarium/symfony-lazysec/wiki/User-provider
-    LazySec\Provider\GenericUserProvider: ~
+    LazySec\Provider\GenericUserProvider:
+        autowire: true
+        public: false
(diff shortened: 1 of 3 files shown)


webinarium/symfony-lazysec/2.0/config/packages/lazysec.yaml
```diff
@@ -0,0 +1,8 @@
+services:
+    _defaults:
+        autowire:      true
```
Nyholm: Do you really need _defaults and autoworker when you only congure one service?

(I have not looked at the bundle, maybe you do)

webinarium:
> Do you really need _defaults and autoworker when you only congure one service?

Hi @Nyholm,

Thank you for your feedback. You are absolutely right - no need for the defaults, so I simplified the config file there.

Best regards,
Artem

webinarium/symfony-lazysec/2.0/config/packages/lazysec.yaml
webinarium: Thank you for your feedback. You are absolutely right - no need for the defaults, so I have simplified the config file and made another commit.

nicolas-grekas:
shouldn't you also add a commented line like this one, with a comment?
`#    LazySec\Repository\UserRepositoryInterface: '@App\Repository\UserRepository'`

webinarium:
@nicolas-grekas, so true!
I didn't add that line because it would break an app without a repository prepared first, but it's slipped from my mind that I can add this line commented. :) Thank you for the hint, I have updated the config and pushed the update.

webinarium:
If there is anything else needed to get this PR merged, please let me know.

PR #244, opened 2018-01-10
Introduce algolia/search-bundle v3

julienbourdeau:
| Q             | A
| ------------- | ---
| License       | MIT

There is already a recipe for the Algolia Search Bundle and I have just released the version 3.0. Version 2 will be maintained (patch-only) but should not be installed for new projects.

This version introduce many new features but mostly:

* **Simple**: You can get started with only 5 lines of YAML
* **Extensible**: It let’s you easily replace services by implementing Interfaces
* **Standard**: It leverages `Normalizers` to convert entities for indexing
* **Dev-friendly**: It lets you disable HTTP calls easily (while running tests for examples)
* **Futur-ready**: It let’s you unsubscribe from doctrine events easily to use a messaging/queue system.

Documentation can be found [here](http://www.algolia.com/doc/api-client/symfony/) or in the [README](https://github.com/algolia/search-bundle/blob/master/README.md).

Any feedback is very welcomed 😃

### DIFF
--- a/algolia/search-bundle/3.0/config/packages/algolia_search.yaml
+++ b/algolia/search-bundle/3.0/config/packages/algolia_search.yaml
@@ -1,0 +1,4 @@
+# All available configuration can be found here:
+# https://www.algolia.com/doc/api-client/symfony/configuration/
+algolia_search:
+    prefix: '%env(APP_ENV)%_'
--- a/algolia/search-bundle/3.0/manifest.json
+++ b/algolia/search-bundle/3.0/manifest.json
@@ -1,0 +1,14 @@
+{
+    "bundles": {
+        "Algolia\\SearchBundle\\AlgoliaSearchBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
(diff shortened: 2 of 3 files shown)


algolia/algolia-search-bundle/2.2/config/packages/algolia.yaml
```diff
@@ -1,3 +0,0 @@
-algolia:
```
Nyholm: No need to remove the old recipe

julienbourdeau: I think it's important to remove it, people should use it anymore. I believe this is why the tests are failing. I'm not sure how to fix this 🤔

Nyholm: If I really want to install 2.2 for some reason, shouldn't I be able to use the old recipe?

julienbourdeau:
In the end, I have restored the recipe for 2.2. We can remove later if it makes sense, you're right @Nyholm it might be too early now. 👍

julienbourdeau:
Thanks @Nyholm 👌

PR #247, opened 2018-01-11
Add endroid/pdf recipe

endroid:
| Q             | A
| ------------- | ---
| License       | MIT


endroid/pdf/1.0/config/packages/endroid_pdf.yaml
```diff
@@ -0,0 +1,7 @@
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+        public: false
+
+    Endroid\Pdf\Pdf: ~
```
Nyholm: Do we really need to auto wire just one service? The config will be smaller if we don’t :)

endroid: Yes this is the only service that needs to be wired. How would you make it shorter? If I leave out line 7 the service is not found and if I leave out the _defaults the constructor arguments are not injected automatically. But if you think there is a part we can leave out while still having an autowired service, I'll be happy to remove that.

Nyholm: How about this: 
```
services: 
    Endroid\Pdf\Pdf: 
        autowire: true
```

??

endroid: Ah that's what you meant. I personally like to keep autowiring in the defaults in all my projects. I agree with you that it is not really necessary here, but it is a matter of personal preference.

Nyholm: Yes it is. So lets be consistent with all the other and make the service declaration easier to read. Then I'll be happy to approve this PR.

endroid: @Nyholm I tried the configuration as above but then Travis throws an error that Knp\Snappy\Pdf (which should be wired as the first argument) is not found. When I change the service definition back to the one with the _defaults key the error is gone again.

Do you have any idea why the first configuration works and the latter not?

PR #251, opened 2018-01-14
Add endroid/image-placeholder

endroid:
| Q             | A
| ------------- | ---
| License       | MIT


stof:
> but we still need to update declaration of ImagePlaceholder for each version

not really. You could use `_instanceof` to apply a tag on all providers, and then use `!tagged` to inject them (this assumes that the ImagePlaceholder can take an iterator in its constructor rather than doing multiple `addProvider` calls, but this would even allow doing lazy-loading)

endroid:
Nice :) I updated the service and configuration so providers are tagged and injected automatically.

PR #258, opened 2018-01-18
Added theofidry/alice-data-fixtures recipe

Toflar:
| Q             | A
| ------------- | ---
| License       | MIT


Toflar:
Hmm, I was told in the Symfony Slack `#symfony-flex` channel that if I install a composer package of type `symfony-bundle`, it should automatically detect and register the bundles. `theofidry/alice-data-fixtures recipe` is a `library` providing bridges in the `Bridge` namespace for multiple integrations. I think it's perfectly valid to register such a recipe. Otherwise one would have to register 10 packages with an individual `type` that all integrate the `library`. Sure, that's possible but I don't necessarily think that the approach that was chosen in this package is wrong. Wdyt?

Also, I've seen `symfony-pack` but the auto-validation does not say anything about that?

theofidry/alice-data-fixtures/1.0/manifest.json
```diff
@@ -0,0 +1,6 @@
+{
+    "bundles": {
+        "Nelmio\\Alice\\Bridge\\Symfony\\NelmioAliceBundle": ["test"],
```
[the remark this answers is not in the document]
Toflar: True, will update that once we have feedback regarding the whole "bundle in a subdirectory" question :)

fabpot:
Most bundles are self-contained and must be of type `symfony-bundle`, but having the bundle as a sub-dir of a library is fine by me.

Toflar:
Does that mean the Symfony Recipes Server Validation is going to be updated? Or was it already? Sorry, I'm confused now :-D

Nyholm:
Any updates here?

Toflar:
Was that a question for me or @fabpot? I haven't been told that the flex-bot now accepts anything else than `symfony-bundle` so I haven't been looking into it anymore.

Nyholm:
It was meant for you. 

> I haven't been told that the flex-bot now accepts anything else than symfony-bundle so I haven't been looking into it anymore.

Using `library` is okey. The review bot may not be happy but Flex supports it.

stof:
A recipe registering a bundle in a library package won't pass the flex-bot validation. But the Symfony core team can still decide to accept the recipe and merge it.

Toflar:
Ah, I didn't know that. Would've updated long ago if I had 😄 It's ready now, `nelmio/alice` has been added in the meantime so no second PR needed.
I don't know if I should add an example config file because at the moment, everywhere I use it, I use the defaults so I don't need to configure anything.

theofidry/alice-data-fixtures/1.0/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "bundles": {
+        "Fidry\\AliceDataFixtures\\Bridge\\Symfony\\FidryAliceDataFixturesBundle": ["test"]
```
theofidry: I would also add `dev` like done in [nelmio/alice](https://github.com/symfony/recipes-contrib/blob/55858fa45cf9dadba4247b1d354e6ef5ffc6937f/nelmio/alice/3.2/manifest.json#L3)

theofidry:
@Toflar is that it? As did you try to see if that was working? If so I'll be happy to see that PR merged and we could update the install instructions of [AliceDataFixtures](https://github.com/theofidry/AliceDataFixtures) :)

theofidry/alice-data-fixtures/1.0/manifest.json
Toflar: Added.

Toflar:
> @Toflar is that it? 

Yeah, what else do you expect?

theofidry:
Then it should be good for another review :)

Nyholm:
This recipe basically tells Flex that this `library` actually contains a bundle.

theofidry:
Thanks you guys 💚

PR #266, opened 2018-01-26
Display the list of packages being required during testing

stof:
| Q             | A
| ------------- | ---
| License       | MIT


stof:
@fabpot `.travis.yml` should be considered special in the validator, as it is not part of a recipe, so using the `.yml` extension is fine (it is even required for Travis to work).
Can you exclude it from the validation ?

fabpot:
merge now

Nyholm:
Since we just scrape the packages from https://symfony.sh, we do not support when a recipe deals with multiple packages. 

We could solve it with adding some HTML comments on the PR page that travis scrapes

stof:
@Nyholm the fact that the Travis setup might be broken when there is multiple packages is something I was suspecting as well.
Adding this output in the Travis setup makes it possible to know what command is running, and was a first step for me towards checking what happens for multiple packages.

Nyholm:
@fabpot Is it okey if I merge this manually or is flex bot required to merge all PRs?

PR #269, opened 2018-01-31
Add KNPLabs Dictionary bundle recipe

PedroTroller:
| Q             | A
| ------------- | ---
| License       | MIT

Recipe for https://github.com/KnpLabs/DictionaryBundle

### DIFF
--- a/knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
+++ b/knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
@@ -1,0 +1,17 @@
+knp_dictionary:
+    dictionaries: []
+
+# Examples
+# knp_dictionary:
+#     dictionaries:
+#         my_dictionary:
+#             type: 'key_value'
+#             content:
+#                 "foo": "foo_value"
+#                 "bar": "bar_value"
+#                 "baz": "baz_value"
+
+#         my_callable_dictionary:
+#             type: 'callable'
+#             service: 'app.service.id'
+#             method: 'getSomething'
--- a/knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
+++ b/knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
@@ -1,17 +1,2 @@
 knp_dictionary:
     dictionaries: []
-
-# Examples
-# knp_dictionary:
-#     dictionaries:
-#         my_dictionary:
-#             type: 'key_value'
-#             content:
-#                 "foo": "foo_value"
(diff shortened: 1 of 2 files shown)


knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
```diff
@@ -0,0 +1,17 @@
+knp_dictionary:
+    dictionaries: []
+
+# Examples
+# knp_dictionary:
```
stof: I'm -1 for this commented example. Files copied into the project are not meant to be documentation. And this commented section is not something they can uncomment, as it would make the YAML file invalid (duplicate key)

knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
```diff
@@ -0,0 +1,17 @@
+knp_dictionary:
+    dictionaries: []
+
+# Examples
```
Nyholm: Do we need this example docs? Could we do a link the documentation instead?

knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
PedroTroller: @stof :+1: fixed

knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
```diff
@@ -0,0 +1,2 @@
+knp_dictionary:
+    dictionaries: []
```
Nyholm: Is it likely that a user will add more dictionaries here?

What would happen if we did not use this config?

PedroTroller: .... nothing ....

knplabs/dictionary-bundle/2.0/config/packages/knp_dictionary.yaml
PedroTroller: Already removed :wink:

Nyholm:
> > What would happen if we did not use this config?
> .... nothing ....

I would like this recipe to bring some value to the user. Could we add something to the config will add value for 80% of the users?

If not, flex is smart enough to install this bundle without a recipe.

stof:
@Nyholm AFAICT, the bundle is useless if you don't configure some dictionaries in the config file. It would probably work, but allowing you to access values in your dictionaries does not bring value if you don't have any.

PR #272, opened 2018-02-03
Added Client and Secret keys

chindit:
| Q             | A
| ------------- | ---
| License       | MIT

Added Client and Secret key values in config.

Very useful for most of the calls which require these two values.


Nyholm:
Thank you. Is it common that most users need both `developerKey` and `key`/`secret`?

chindit:
I personally never used `developerKey` as all the services I use require `key` and `secret`.
But I didn't want to remove `developerKey` entry as I'm not the original author of the recipe.
Anyway, leaving the `developerKey` blank will not cause any issue.

PR #273, opened 2018-02-03
added recipe for KnpPaginatorBundle

felipyamorim:
| Q             | A
| ------------- | ---
| License       | MIT

Small recipe for [KnpPaginatorBundle](https://github.com/KnpLabs/KnpPaginatorBundle)


felipyamorim:
Suggestions to fix version 3.4? KnpPaginatorBundle needs to require translation as dependency ?

felipyamorim:
Thank you for the help

The config file is the default specified in the documentation, but i understand your point, you think is unnecessary this recipe ? I only made this to automate the installation of the same.

stof:
if the config is the default one, there is no reason to configure it explicitly. Relying on the defaults is better (as it means you don't have to maintain this config in your project)

Nyholm:
I'll close this due inactivity.

PR #282, opened 2018-02-07
Add support for Roukmoute Hashids bundle

roukmoute:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/roukmoute/hashids-bundle/1.4/manifest.json
+++ b/roukmoute/hashids-bundle/1.4/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Roukmoute\\HashidsBundle\\RoukmouteHashidsBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}
--- a/roukmoute/hashids-bundle/1.4/config/packages/roukmoute_hashids.yaml
+++ b/roukmoute/hashids-bundle/1.4/config/packages/roukmoute_hashids.yaml
@@ -1,0 +1,3 @@
+# Read the documentation: https://github.com/roukmoute/HashidsBundle/blob/master/README.md
+roukmoute_hashids:
+    salt: 'change me'


javiereguiluz:
@roukmoute thanks for this contribution! We try to minimize the number of env vars because they are very problematic to manage for lots of people. So, I'm asking if you think that these four config options should really be env vars or they could be just normal config params. We do that in most of our recipes. For example in https://github.com/symfony/recipes-contrib/blob/master/doctrine/mongodb-odm-bundle/3.3/config/packages/doctrine_mongodb.yaml we just define two env vars for some really special params ... but the rest are normal params. Thanks!

roukmoute:
@javiereguiluz Of course I can do that :)
In fact in my repoy, this is like you suggest.
I will fix it.

roukmoute:
Done @Nyholm :)

roukmoute:
Done @Nyholm :)

roukmoute:
I try but currently I have some troubles to test it in local:
```
                                                                                                                            
  [Composer\Downloader\TransportException]                                                                                  
  The "https://symfony.sh/r/github.com/symfony/recipes-contrib/ulid" file could not be downloaded (HTTP/1.1 404 Not Found)  
                                                                                                                            
```

roukmoute/hashids-bundle/1.4/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "Roukmoute\\HashidsBundle": ["all"]
```
Nyholm: Travis [complains here](https://travis-ci.org/symfony/recipes-contrib/jobs/341340179#L610). This line should be: 
`Roukmoute\\HashidsBundle\\RoukmouteHashidsBundle`

roukmoute:
@Nyholm What do you want me to do?

Now the error is caused by `DoctrineExtension`:
```bash
To configure the ORM layer, you must first install the doctrine/orm package  
```

I can add in my `require-dev` → `doctrine/orm`.
Even if for me it is not to me that I add this extension.

What do you suggest?

Nyholm:
Yeah, why do you get this error? Do you require doctrine/orm in your bundle?

stof:
@Nyholm that's actually related to our DoctrineBundle recipe. The DoctrineBundle recipe configures the ORM layer, but the ORM is an optional dependency of DoctrineBundle, not a required one (so people wanting to use only the DBAL layer have to remove stuff after installing the recipe to get things working).

@roukmoute your dependency on DoctrineBundle alone looks weird to me though. You don't need the DBAL layer at all. Your only place depending on DoctrineBundle is for the param converter, and this will only work with the ORM enabled. So you should either have a requirement on the ORM (not only on DoctrineBundle) or make DoctrineBundle an optional dependency and disable the param converter when you don't have both DoctrineBundle and the ORM in the project.

roukmoute:
:pray:

roukmoute:
Finally done @Nyholm!
https://travis-ci.org/symfony/recipes-contrib/jobs/348224644

roukmoute:
Yes it will be the next version, I will do that quickly :+1:

fabpot:
@symfony-flex-server review again

PR #286, opened 2018-02-09
Add CommandSchedulerBundle recipe

J-Mose:
| Q             | A
| ------------- | ---
| License       | MIT
| Version       | 2.0

Adding recipe for https://packagist.org/packages/jmose/command-scheduler-bundle


J-Mose:
Yes, it is. But I think users will customize it most of the time.

Should i comment it and add a link to the documentation ?

J-Mose:
@Nyholm, I did a little cleaning in the configuration file to leave only the two most important parameters (usually customized) and added some post-install instructions.

PR #289, opened 2018-02-10
add onurb/doctrine-yuml-bundle recipe

Nono1971:
| Q             | A
| ------------- | ---
| License       | MIT

should be OK this time


stof:
having no requirement on any Symfony component in your bundle looks weird to me (and it also means that there is nothing enfoncing compatible SF versions btw)

Nono1971:
The main requirement is symfony itself... (better for a symfony bundle :D )

But with the sf4 micro-framework logic, you're right, I should specify the few required packages (like routes or profiler). Since sf4, everything needed was already there with symfony. I'm not sure that's it's still true now ^^
 I'll check it in the evening after work (the one paying my salary ;) ) ... and will adapt my composer.json :)

Thanks for your usefull remark ! :)

PR #291, opened 2018-02-14
Add webinarium/datatables-bundle

webinarium:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/webinarium/datatables-bundle/2.3/manifest.json
+++ b/webinarium/datatables-bundle/2.3/manifest.json
@@ -1,4 +1,7 @@
 {
+    "bundles": {
+        "DataTables\\DataTablesBundle": ["all"]
+    },
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/"
     }


webinarium/datatables-bundle/2.3/config/packages/datatables.yaml
```diff
@@ -0,0 +1,4 @@
+services:
+
+    # See https://github.com/webinarium/DataTablesBundle/wiki
+    DataTables\DataTablesInterface: '@datatables'
```
Nyholm: Couldn’t this service configuration be inside the bundle?

webinarium: True, can put it inside the bundle. Does it make sense to have a recipe just for bundle registration? Maybe not, please advise.

Nyholm: No, it would not make sense to have a recipe for just bundle registration. Flex is smart enough to install that bundle. 

However, a general idea that I really like is to not use a bundle at all. I like to use a flex recipe to register the services you need and then that's it =)
Looking at your bundle though I see that you do need a compiler pass so moving to "no bundle"  might not be a good it for you.

PR #293, opened 2018-02-15
Update jms/serializer-bundle configuration

covex-nn:
| Q             | A
| ------------- | ---
| License       | MIT

With this PR new directory `%kernel.project_dir%/config/metadata/jms_serializer` will be created, where serialization metadata configuration should be stored for objects with `App` namespace prefix


covex-nn:
@goetas please, review changes

goetas:
From the technical point of view is correct. Since symfony "wiped-out" the AppBundle, this makes sense to me. 
But I will still keep the section commented (and keep it just as a suggestion) as most of the people are using the annotations.

So for me is :-1:  as it is.

jms/serializer-bundle/2.0/config/packages/jms_serializer.yaml
```diff
@@ -2,12 +2,9 @@ jms_serializer:
...
-#                path: "@MyFooBundle/Resources/config/serializer"
-#            another-name:
-#                namespace_prefix: "My\\BarBundle"
-#                path: "@MyBarBundle/Resources/config/serializer"
+    metadata:
+        auto_detection: true
+        directories:
+            app:
+                namespace_prefix: "App"
+                path: "%kernel.project_dir%/config/serializer"
```
goetas: Is this the recommended directory to put metadata configurations?

jms/serializer-bundle/2.0/config/packages/jms_serializer.yaml
```diff
@@ -2,12 +2,9 @@ jms_serializer:
...
-#        auto_detection: false
-#        directories:
-#            any-name:
-#                namespace_prefix: "My\\FooBundle"
-#                path: "@MyFooBundle/Resources/config/serializer"
-#            another-name:
-#                namespace_prefix: "My\\BarBundle"
-#                path: "@MyBarBundle/Resources/config/serializer"
+    metadata:
+        auto_detection: true
```
goetas: do not remember if `auto_detection: true` and `directories` work well together

covex-nn: they work together, i checked
see https://github.com/schmittjoh/JMSSerializerBundle/blob/master/DependencyInjection/JMSSerializerExtension.php#L130

goetas: great! I forgot about it! :)

jms/serializer-bundle/2.0/config/packages/jms_serializer.yaml
goetas: Probably the dir name should be `jms_serializer`, to avoid conflicts with the symfony serializer

covex-nn: There is no recommended directory for such metadata configurations. But there was a discussion about a place of `translations` directory:

First it was moved to `/config/translations` (see https://github.com/symfony/recipes/pull/249), but then it was moved back to the root of repository (see https://github.com/symfony/recipes/pull/253) because `Discoverability of this directory is important because it's one of those dirs that could be browsed by non backend developers (such as assets/ and templates/).` (c)

I've chosen `config/serilizer` because metadata configuration files are stored in `<bundle-directory>/Resources/config/serializer` for bundles and discoverabilty of `config/serialier` directory is not important as for translations, and because metadata configuration is configuration =)

goetas: What about  `%kernel.project_dir%/config/metadata/jms_serializer` ?

goetas:
I see you have commented back some part of it, should you comment everything?

`auto_detection` is true by default, no need to specify it

jms/serializer-bundle/2.0/config/packages/jms_serializer.yaml
covex-nn: Sounds good!
Cache dir is `%kernel.cache_dir%/jms_serializer` and metadata configuration will be in `%kernel.project_dir%/config/metadata/jms_serializer`. Similar and unique )

goetas: >Cache dir is `%kernel.cache_dir%/jms_serializer` and metadata configuration will be in > `%kernel.project_dir%/config/metadata/jms_serializer`. Similar and unique )

:+1:

covex-nn:
I thought `# auto_detection: false` and `auto_detection: true` is the same. And with `auto_detection: true` there is only one commented block.

I'd like to to leave `%kernel.project_dir%/config/metadata/jms_serializer` directory uncommented to be able to create a recipe with serializer metadata configuration. To make it available out from a box

goetas:
I will comment everything just to not overload the user... If the user is using annotations (by experience most of them do), this section should be commented.

I'm ok with suggesting nice defaults (ready to be un-commented by the user) but the user should not have to do extra operations most of the time.

covex-nn:
I will try to argue one more time (the last one =)

Before Flex we had to create `AppBundle` in `src` directory and `jms/serializer-bundle` anyway would check `src/AppBundle/Resources/config/serializer` dir because of `auto_detection: true`. With my proposal a number of directories with metadata configuration is the same. So, it is not a overload, IMO. But if most of users are using annotations, may be you are right and i should comment everything (i am not sure =)

Anyway, i'd like to show you an examle: https://github.com/covex-nn/recipes-contrib/commit/2e5da5cbe9121d9f4d53190c8d6036c2b15a9b07. This is a commit with recipe for `sonata-project/notification-bundle` (bundle itself must be released before i will create a PR, and this commit is not last, i guess).

* A recipe contains: (1) Bundle configuration, (2) Entity `App\Entity\SonataNotificationMessage` class, (3) Serializer metadata configuration
* Bundle does not require `doctrine/doctrine-bundle` and `jms/serializer-bundle`, they are optional

After `composer require sonata-project/notification-bundle`, entity class will be copied by Flex in `src/Entity` directory and if DoctrineBundle in already installed, after `bin/console doctrine:schema:update`, database will be updated. At the moment i cannot use `@ JMS\Serializer\AnnotationGroup` annotation, because of error:

> [Semantical Error] The annotation "@ JMS\Serializer\Annotation\Groups" in class App\Entity\SonataNotificationMessage does not exist, or could not be auto-loaded.

But after `composer require jms/serializer-bundle` serialization will be available and already set up correctly, because `Entity.SonataNotificationMessage.xml` was copied from recipe too. That is the reason why i suggest not to comment `config/metadata/jms_serializer` directory.

If you are still not agree with me, i will comment new directory in configuration.

stof:
We could keep the directory mapping uncommented IMO. If the directory does not exist or does not have mapping in it, JMSSerializerBundle should be smart enough to detect this at build time, and so there won't be any performance impact to have it configured even when using annotations AFAIK.

goetas:
Do nor remember if `JMSSerializerBundle` is  smart enough to not trigger errors if the dir does not exists or similar....

covex-nn:
`JMSSerializerBundle` will throw RuntimeException if directory from `jms_serializer.metadata.directories` does not exists, thats true. But `%kernel.project_dir%/config/metadata/jms_serializer` will be created by Flex after a recipe will be applied

goetas:
Then I'm ok with it!
:+1:

covex-nn:
I've got a error message when i tried to create a recipe for `sonata-project/notification-bundle`:

> sonata-project/notification-bundle/3.4/config/metadata/jms_serializer/Entity.SonataNotificationMessage.xml
> Underscore notation is required for file names under config/

So, if metadata directory will be inside `config/`, then this PR will be useless: it will be imposible to put metadata files there with recipe =( I will move this directory from `config/` to the root

jms/serializer-bundle/2.0/manifest.json
```diff
@@ -3,6 +3,7 @@
         "JMS\\SerializerBundle\\JMSSerializerBundle": ["all"]
     },
     "copy-from-recipe": {
-        "config/": "%CONFIG_DIR%/"
+        "config/": "%CONFIG_DIR%/",
+        "metadata/": "metadata/"
```
Nyholm: Will this always work? Don't we need a prefix?

covex-nn: I guess it will always work, because `%CONFIG_DIR% === 'config'`

covex-nn:
@Nyholm i planned to put `metadata/` directory into `config/`, but files like `SonataNotificationMessage.orm.xml` and `Entity.SonataNotificationMessage.xml` did not pass Flex Server validation. And now it is looks like i am trying to use non-standard directory for a popular bundle. This is not right, i guess. I will create an issue to discuss it.

covex-nn:
@Nyholm metadata is configuration, that describes data for some purposes. Metadata configuration for `jms/serializer-bundle` could be described via annotations, but sometimes JMS Serializer is optional and it is not possible to use that annotations. It cases like #302 metadata configuration must be stored somewhere in external files. That files could be updated by developer

Nyholm:
Since “metadata” in this case actually is config, I would put it in the config dir. Let’s investigate in the issue with camelCase or snake_case. 

Good that you opened that issue.

PR #294, opened 2018-02-15
Add Picoss Yousign Bundle

picoss:
| Q             | A
| ------------- | ---
| License       | MIT

Add recipe for Picoss Yousign bundle that provides integration for
[Yousign API](http://developer.yousign.fr).

Yousign API offers a powerfull SOAP API to add a process of signature or archive document.

See [https://yousign.com/](https://yousign.com/) for more informations.


picoss/yousign-bundle/2.0/config/packages/picoss_yousign.yaml
```diff
@@ -0,0 +1,7 @@
+picoss_yousign:
+    env: '%env(YOUSIGN_ENV)%'
+    api_key: '%env(YOUSIGN_API_KEY)%'
+    username: '%env(YOUSIGN_USERNAME)%'
+    password: '%env(YOUSIGN_PASSWORD)%'
+#    soap_options:
+#        tarce: 1
```
Nyholm: Tarce?

Why do we need commented out config? I suggest removing this.

picoss: soap_options are not required, but it's just to show that it's possible to pass some options tu the saop client used by the API.
Do you think it should be explain only in the documentation rather than as a comment in the config file ?

And your right about the "tarce", it should be trace.

stof: yeah, putting it in the documentation is better than commented here, especially if it is not config that most users will need to use.

picoss: Ok, I'll remove them.

Thanks

PR #296, opened 2018-02-17
Add MarvelApiBundle recipe

ikoene:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/ikoene/marvel-api-bundle
https://packagist.org/packages/ikoene/marvel-api-bundle


Nyholm:
It looks good. But shouldn’t the service definition be in the bundle itself? And you use the environment variables in the bundles config.

ikoene:
It actually is [inside the bundle](https://github.com/ikoene/marvel-api-bundle/blob/master/Resources/config/services.yml). I though copying it over was a way to expose the definition to the user. I'll remove it.

ikoene:
`Recipe is not needed as it only registers bundle `

@Nyholm That is incorrect because environment variables are also added.

ikoene:
@Nyholm I should've read your comment more carefully. Moved the use of env variables outside the bundle. Should be good now.

PR #298, opened 2018-02-20
[WIP] Added php-task

wachterjohannes:
| Q             | A
| ------------- | ---
| License       | MIT

This PR add the php-task bundle to enable asynchronous task execution.

### DIFF
--- a/php-task/task-bundle/1.2/config/packages/task.yaml
+++ b/php-task/task-bundle/1.2/config/packages/task.yaml
@@ -1,8 +1,0 @@
-task:
-    storage:                 doctrine # One of "array"; "doctrine"
-    run:
-        mode:                'off' # One of "off"; "listener"
-    locking:
-        enabled:             false
-    executor:
-        type:                inline # One of "inside"; "separate"
--- a/php-task/task-bundle/1.2/config/packages/task.yaml
+++ b/php-task/task-bundle/1.2/config/packages/task.yaml
@@ -4,11 +4,5 @@ task:
         mode:                'off' # One of "off"; "listener"
     locking:
         enabled:             false
-        # ttl:                 600
-        # storages:
-        #     file:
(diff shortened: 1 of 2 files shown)


php-task/task-bundle/1.2/config/packages/task.yaml
```diff
@@ -0,0 +1,14 @@
+task:
+    storage:                 doctrine # One of "array"; "doctrine"
+    run:
+        mode:                'off' # One of "off"; "listener"
+    locking:
+        enabled:             false
+        # ttl:                 600
```
Nyholm: Try to avoid having configuration as comments like this.

Nyholm:
You also need to update composer.json and mark you package as “symfony-bundle”. 
The static validation and Travis needs to be green so I can merge it.

wachterjohannes:
@Nyholm I have seen that :) will update that in the evening!

wachterjohannes:
@Nyholm this should now be valid

stof:
@wachterjohannes you still have a bunch of commented config setting, which should not be part of the file added in the project (the console_path for instance is something that 99% of users will never need to change as your bundle should be able to have a default value suited for all projects following the Symfony structure).

The file added by the recipe becomes part of the userland project, and so is maintained by users after that. Using it as documentation is not a good idea (as documentation should be maintained by you and updated when the bundle changes)

wachterjohannes:
@stof ou shit :) completely forgotten that! thanks for the hint

wachterjohannes:
@Nyholm and the bundle will also be added to the kernel?

Nyholm:
Yes, it will =)

wachterjohannes:
@Nyholm ok then its not necessary :) i would like to see some default config in the application when installing a bundle - but thats ok if not!

PR #300, opened 2018-02-21
Add recipe for nella/monolog-tracy-bundle

mhujer:
Bundle URL: https://github.com/nella/monolog-tracy-bundle

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/nella/monolog-tracy-bundle/2.0/manifest.json
+++ b/nella/monolog-tracy-bundle/2.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Nella\\MonologTracyBundle\\MonologTracyBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}
--- a/nella/monolog-tracy-bundle/2.0/config/packages/nella_monolog_tracy.yaml
+++ b/nella/monolog-tracy-bundle/2.0/config/packages/nella_monolog_tracy.yaml
@@ -1,0 +1,7 @@
+monolog:
+    handlers:
+        blueScreen:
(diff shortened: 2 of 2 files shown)


mhujer:
I think that this should pass the validation, as the bundle [is licensed](https://github.com/nella/monolog-tracy-bundle/blob/master/LICENSE.md) under all three MIT/GPL2/GPL3.


**Report from Symfony bot:**
```
Version "2.0" has license "GPL-2.0-only", but only MIT-like/Apache licenses are accepted
Version "2.0" has license "GPL-3.0-only", but only MIT-like/Apache licenses are accepted
```

fabpot:
@symfony-flex-server review again please

fabpot:
@symfony-flex-server review one more time please, I messed up, sorry :)

fabpot:
Issue with the license is fixed now.

mhujer:
Thank you very much @fabpot 👍

nella/monolog-tracy-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Nella\\MonologTracyBundle\\MonologTracyBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "log/": "%VAR_DIR%/log/"
```
mhujer: If I understand it correctly, as the `/var/` directory is gitignored in `symfony/skeleton`, it would be better to let the library automatically create the directory, instead of creating it in recipe, right?

mhujer:
R4R, I successfully tested the recipe locally.

Travis fails only for Symfony 3.4 as the bundle allows only `~4.0`.

PR #305, opened 2018-03-01
Add twital bundle recipe

goetas:
| Q             | A
| ------------- | ---
| License       | MIT

This is a recipe for https://github.com/goetas/twital-bundle

A fully compatible Twig template engine with enhanced syntax.

### DIFF
--- a/goetas/twital-bundle/1.3/config/packages/twital.yaml
+++ b/goetas/twital-bundle/1.3/config/packages/twital.yaml
@@ -1,0 +1,7 @@
+goetas_twital:
+    full_twig_compatibility: true # see http://twitalbundle.readthedocs.io/en/latest/#full-twig-compatibility
+
+## add here your custom extension handlers, by default Twital parses *.html.twital, *.xml.twital and *.xhtml.twital files
+## more info about it on http://twitalbundle.readthedocs.io/en/latest/#source-adapters
+#    source_adapter:
+#        - { service: twital.source_adapter.html5, pattern: ['/\.ext\.twital$/', '/\.ext2\.twital$/'] }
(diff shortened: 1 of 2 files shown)


goetas/twital-bundle/1.2/config/twital.yaml
```diff
@@ -0,0 +1,10 @@
+framework:
+    templating:
+        engines: ['twig', 'twital']
+# optional configurations for file extension matching
```
Nyholm: Not sure we should include this optional config. I would much rather see a link to the bundle's documentation

goetas:
Hi! Thanks for keeping an eye on this!  Now it is ready! 

(Was not easy! I feel that since all the dependencies become optional, it is not easy to track what is available and what is not).

goetas/twital-bundle/1.3/config/packages/twital.yaml
```diff
@@ -0,0 +1,7 @@
+goetas_twital:
+    full_twig_compatibility: true # see http://twitalbundle.readthedocs.io/en/latest/#full-twig-compatibility
+
+## add here your custom extension handlers, by default Twital parses *.html.twital, *.xml.twital and *.xhtml.twital files
```
goetas: @Nyholm If you think is a better idea is can completely remove this comment and let the users find out this detail via documentation.

Was thinking to put just a note "By default Twital parses *.html.twital, *.xml.twital and *.xhtml.twital files, more info about it on http://twitalbundle.readthedocs.io/en/latest/#source-adapters"

goetas:
@Nyholm anything missing with this PR that I should improve?

goetas:
Thanks!

PR #306, opened 2018-03-03
Update machy8/webloader-bundle

Machy8:
| Q             | A
| ------------- | ---
| License       | MIT

- Changed directory structure
- Added web_loader.yaml with pre-defined fields


machy8/webloader-bundle/1.1/config/packages/web_loader.yaml
```diff
@@ -0,0 +1,4 @@
+web_loader:
+    pathPlaceholders:
+    filesCollections:
```
Nyholm: Why are these needed? Do they do anything?

Machy8: Yes. Path placeholders and files collections are the most used configuration parameters in the webloader. So they are here, prearranged, yet empty.

stof: be careful, they are `null` here, not an empty string, due to the way Yaml works. The empty string must be represented as a quoted value. So make sure your bundle handles null values, or update this.

Machy8: Weloader have condition check for the null values https://github.com/Machy8/webloader-bundle/blob/master/DependencyInjection/WebLoaderExtension.php#L43

Machy8:
@Nyholm could you please merge this?

PR #310, opened 2018-03-10
Base config for package

c0ntax:
| Q             | A
| ------------- | ---
| License       | MIT


c0ntax/env-providers-bundle/0.1/config/packages/env_providers.yaml
```diff
@@ -0,0 +1,3 @@
+c0ntax_env_providers:
+    array:
+        return_null_if_empty: true
```
Nyholm: Is this the same as the default config?

c0ntax: It is, yes. Is that a bad thing?

I've fixed the SF4 issue. How can I kick of the the Travis check again?

stof: if if is the same than the default value, and the default is expected to work fine for most people, it is better to *not* configure it in the project (relying on the default instead), and then this recipe is useless (as the default auto-generated recipe will be able to register the bundle)

c0ntax: I've always been a fan of making things more explicit in config rather than having mystery defaults that people have to hunt around for, but it's just a personal preference. I guess I should delete this pull request then unless I add anything to the bundle that merits more setup?

Nyholm: > I guess I should delete this pull request then unless I add anything to the bundle that merits more setup?

Yes, for the reasons @stof mentioned.

OskarStark:
Should be closed

PR #313, opened 2018-03-14
add SteamAuthenticationBundle recipe

knojector:
| Q             | A
| ------------- | ---
| License       | MIT


knojector:
The Bundle requires the security component but refering to the contribution guideline it's not valid if the bundle requires the component. Is there a solution for this?

The 3.4 Travis fails because of the required minimum version of Symfony 4.0.

knojector/steam-authentication-bundle/1.0/config/packages/knojector_steam_authentication.yaml
```diff
@@ -0,0 +1,5 @@
+knojector_steam_authentication:
+    api_key: ''
```
Nyholm: What about adding environment parameter here?

knojector/steam-authentication-bundle/1.0/config/packages/knojector_steam_authentication.yaml
```diff
@@ -0,0 +1,5 @@
+knojector_steam_authentication:
```
Nyholm: Maybe a link to your documentation so I know where to go if I need support/docs?

knojector/steam-authentication-bundle/1.0/config/packages/knojector_steam_authentication.yaml
```diff
@@ -0,0 +1,5 @@
...
+    api_key: ''
+    login_route: ''
+    login_redirect: ''
+    user_class: ''
```
Nyholm: route, redirect and user_class are all require right?

knojector: Yes, they are all required. Do I have to make any changes to this?

Nyholm: No, you are fine. Thanks

Nyholm:
This looks good. Thanks!

PR #319, opened 2018-03-17
Add nelmio alice

theofidry:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/nelmio/alice/3.2/config/packages/nelmio_alice.yaml
+++ b/nelmio/alice/3.2/config/packages/nelmio_alice.yaml
@@ -1,0 +1,3 @@
+nelmio_alice:
+    functions_blacklist:
+        - 'current'
--- a/nelmio/alice/3.2/manifest.json
+++ b/nelmio/alice/3.2/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Nelmio\\Alice\\Bridge\\Symfony\\NelmioAliceBundle": ["dev", "test"]
+    },
+    "copy-from-recipe": {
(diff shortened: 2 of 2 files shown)


theofidry:
I've putted a message about it in #symfony-flex on Slack but I think I should put it here as well:

Is this `symfony-bundle` really required? It feels wrong to put `symfony-bundle` instead of `library` as this library could very well add a Laravel or framework X bridge. And Composer does not allow an array for `type`...

theofidry:
For the record:

>fabpot:
>having a lib and a bundle in the same package is fine, but the validator cannot detect that. Anyway, we can still merge the PR, so no worries.

curry684:
Agreed this should not be a showstopper.

theofidry:
@fabpot the tests are not passing because the Travis tests relies on:

```
$(curl -s $SYMFONY_ENDPOINT | sed -En 's/.*composer req "([^"]+)".*/\1/p')
```

to retrieve the package name. However because the package type is currently a blocker for the Flex Server, you only get:

```
There are some validation errors
Tested Recipe: github.com/symfony/recipes-contrib#319

Package nelmio/alice (3.2)
Type must be "symfony-bundle" as the manifest registers a bundle, "library" detected
Modified Packages
nelmio/alice
```

So there's the following options:

- I can update the regex to retrieve the package name from another part of the text when the current on fails (here we can get it from the `Package nelmio/alice (3.2)` bit
- We can make the Flex Server either more tolerant to that issue or still display the other informations on validation errors

theofidry:
❤️

PR #320, opened 2018-03-19
Added recipe for the Prooph pack

gquemener:
| Q             | A
| ------------- | ---
| License       | MIT

Initial conversation is [here](https://github.com/prooph/service-bus-symfony-bundle/issues/41)


Nyholm:
Let me know when you are ready for a review

gquemener:
Hello @Nyholm,

I will do, it might take some time though, discussions are still opened on our side about what we would consider as good default configuration for the prooph bundles.

Thanks.

nicolas-grekas:
Hello, thanks for working on this.
I'm not sure binding recipe to a pack makes sense.
Can't you bind the recipe to the bundle instead?

gquemener:
Hello @nicolas-grekas,

Thanks for your suggestion.
FYI, we've started discussing this topic [here](https://github.com/prooph/service-bus-symfony-bundle/issues/41#issuecomment-374908821).

What we're trying to achieve is to provide a set of basic configuration to user requiring the `prooph/pdo-event-sourcing-pack` in their Symfony project, in order to jump faster into the domain description and not care (too much, at least for a start) about how configuration.

In the mentioned discussion, I've written a section about recipe dependencies, it would be great if you could take some time to review it and share your thoughts about it, please.

I'll copy it here to save you some time:
## Bundle-wide vs Pack/Skeleton-wide recipes

It seems that Symfony Flex can execute recipes no matter if the package is a [library](https://github.com/symfony/recipes/tree/master/symfony/routing/4.0) or a [bundle](https://github.com/symfony/recipes/tree/master/symfony/framework-bundle/3.3).

However, I think this system has a flaw because it creates recipe dependencies.
For instance, requiring `symfony\routing` without `symfony/framework-bundle`, would copy [this config directory](https://github.com/symfony/recipes/tree/master/symfony/routing/4.0/config) that contains [`symfony/framework-bundle` configuration](https://github.com/symfony/recipes/blob/master/symfony/routing/4.0/config/packages/routing.yaml#L1) into the project config directory.
I'm not sure why the Core Symfony team has done that, I guess they considered that if you're using `symfony/flex` then you're in a Symfony project that MUST requires `symfony/framework-bundle`.

In the case of Prooph, if we follow the same example, it would mean creating a `prooph\pdo-event-store` recipe that would create a mysql event store and register it using `prooph\event-store-symfony-bundle` configuration (thus making an implicit dependencies between the two).

My suggestion would be to define one skeleton/pack-wide recipe, executed only when requiring the skeleton/pack. The downside is that this recipe wouldn't be executed when requiring packages manually, but is it really a problem?

Another suggestion would be to define recipes ONLY for prooph bundles.
For example, we could provide *commented* configuration in the `prooph/event-store-symfony-bundle` recipe to register a pdo event store (with explanation that user MUST require the `prooph/pdo-event-store` package in order to uncomment this section).

nicolas-grekas:
The framework-bundle is special, it's a requirement of all and any bundle. You cannot compare to it.
Adding a recipe to a pack feels odd for two reasons:
- packs can be unpacked (see https://medium.com/@fabpot/symfony-4-unpack-the-packs-a672d06850eb)
- installing the bundles would not configure them (but one would expect to)

You can still provide the pack, and also provide recipe(s) for the bundle(s), these are separate things.

gquemener:
In the case of Prooph, the [`prooph/pdo-event-store`](https://github.com/prooph/pdo-event-store) package provides an event store adapter.

Thus, if I understand correctly, we would have to provide the [Symfony configuration for this adapter](https://github.com/symfony/recipes-contrib/pull/320/files#diff-f75788829a5af887d830fc2aa08200d1) in the `prooph/pdo-event-store`, right?

But this configuration will only work if user has also required the [`prooph/event-store-symfony-bundle`](https://github.com/prooph/event-store-symfony-bundle) package.

On the other hand, putting this configuration in the recipe of the `prooph/event-store-symfony-bundle` package would not make sense, as it would not work if the `prooph/pdo-event-store` package hasn't been required.

Isn't it odd?

With a pack recipe, my intent was to ensure that the provided configuration will work (because the pack would contain both bundle and adapter).

nicolas-grekas:
Here is a plan:
- for each store, you provide a recipe that create the store-specific services
- for the bundle, you ship it with a default config that references the pdo store
- you patch the bundle to make it throw a useful exception in case the configured `event_store` is not found

this way:
- when the pack you want to provide is installed everything works (with mysql)
- when a new store is installed, some services its related services are created (but unused of course, so might need some additional help)
- when the bundle is installed standalone, the exception hints the dev that something is missing

That's roughly what we do in similar situations in core.

prooph/service-bus-symfony-bundle/0.6/config/packages/prooph_service_bus.yaml
```diff
@@ -0,0 +1,12 @@
...
+    command_buses:
+        app_command_bus: ~
+    event_buses:
+        app_event_bus: ~
+
+services:
+    _defaults:
+        autowire: true
+
+    Prooph\Bundle\ServiceBus\CommandBus: '@prooph_service_bus.app_command_bus'
```
gquemener: Maybe alias id should be `Prooph\ServiceBus\CommandBus` to be consistent with the following alias? Any thought @UFOMelkor?

gquemener: It will be used in user's controller:
```php

use Prooph\Bundle\ServiceBus\CommandBus;
// or 
use Prooph\ServiceBus\CommandBus;

function __invoke(CommandBus $bus)
{
   $bus->dispatch(...);
}
```

prooph/event-store-bus-bridge/3.1/config/packages/prooph_event_store_bus_bridge.yaml
```diff
@@ -0,0 +1,16 @@
+services:
+    _defaults:
+        autowire: true
+
+    Prooph\EventStoreBusBridge\EventPublisher:
+        tags:
+            - { name: 'prooph_event_store.app.plugin' }
+
+    # Uncomment these lines if your event store support transaction
```
gquemener: The default event store `Prooph\EventStore\Pdo\MySqlEventStore` doesn't implement `Prooph\EventStore\TransactionalEventStore`, is it ok for you to have this comment block then @UFOMelkor ?

prooph/service-bus-symfony-bundle/0.6/config/packages/prooph_service_bus.yaml
UFOMelkor: `Prooph\Bundle\ServiceBus\CommandBus` is an implementation detail of the bundle and should not be used directly. We should tag it as internal.

-> Use `Prooph\ServiceBus\CommandBus`

prooph/event-store-bus-bridge/3.1/config/packages/prooph_event_store_bus_bridge.yaml
```diff
@@ -0,0 +1,16 @@
+services:
+    _defaults:
+        autowire: true
+
+    Prooph\EventStoreBusBridge\EventPublisher:
```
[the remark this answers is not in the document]
gquemener: Indeed, configuration might benefit from being a bit more explicit here.

In that case, I would remove the `Prooph\ServiceBus\EventBus` service alias.

gquemener:
@UFOMelkor Indeed, it'd be better to have explicit service definition (thus, not using autowiring) as provided defaults. 

Instant view of services dependencies, less jump back and force between service definitions and classes, no undesired service aliases.

I'm working on it.

gquemener:
I'm not sure why the build do not pass, as the `prooph/event-store-bus-bridge:^3.1` requirement [should be resolved](https://packagist.org/packages/prooph/event-store-bus-bridge#v3.1.0) :thinking: 

However, I've explicitly defined dependencies accross the service definitions, remove unecessary aliases, renamed `app` to `default` (to make it clear that this configuration is coming from the recipe, which provides "default" configuration) and added some more documentation to improve DX.

It's getting quite hard to refactor this configuration localy.
Getting https://github.com/prooph/prooph-pack/pull/1 merged first would help a lot IMO.

gquemener:
@Nyholm I've had to modify the travis configuration file to make the build work with pull request containing many recipes modifications.

I think there might be an issue with the symfony validation CI step, could it be possible to whitelist the `.travis.yml` file please? I haven't found any open source repo for this recipe server to do it myself.

image

Thanks, and have a nice weekend!

gquemener:
Did you have some time to think about my last comment @Nyholm?

gquemener:
I wouldn't like to look impatient, but I fear this PR get forgotten.

Two questions arise here:

- Is this Prooph Symfony pack still an idea promoted by the maintainers of the Prooph components? (I believe @UFOMelkor, @prolic or @codeliner might answer this question)
- Does the fix to the Travis build execution step (making it able to run on a PR containing multiple repositories contributions, AFAIR the previous configuration only worked for PR containing a single repository contribution) makes sense to the maintainers of the `symfony/recipes-contrib` repository? (I believe @Nyholm or @nicolas-grekas might have an answer for that)

Thank you very much for your time,
IMO, a decision should be made by the end of next week, 2018-04-20, or I will simply close the PR in order not to "pollute" this repository.

codeliner:
> Is this Prooph Symfony pack still an idea promoted by the maintainers of the Prooph components? 

yes, of course

gquemener:
> >    Is this Prooph Symfony pack still an idea promoted by the maintainers of the Prooph components?

> yes, of course

I realise now that this question could be understood with a passive agressive tone, so I'm sorry if you read it that way @codeliner.

The real meaning was "Is it high enough in your backlog priorities to treat it now?".

I'm in contact with @UFOMelkor, so hopefully we can move on soon.

Cheers!

UFOMelkor:
Do you think we can make the `CreateEventStreamCommand` generic and move it into the event-store bundle?

gquemener:
That would be usefull indeed @UFOMelkor.

However, It wouldn't be trivial.
For example, event stream are automatically created when using one stream per aggregate (thus making the command useless if the repository is configured as so).

gquemener:
@UFOMelkor After thinking about it, this is not really possible and as the documentation states (http://docs.getprooph.org/event-store/implementations/pdo_event_store/variants.html#3-2-1-2-5), one should manually create the event stream when using the single stream strategy.

As this is the one configured by default in the recipe, I think it makes sense to provide the command.

On the other hand, we could add some documentation in the command to reference the relevant Prooph documentation chapter.

Another solution would be to configure the AggregateStreamStrategy as the default one in the recipe (and set the repository as "one stream per aggregate"), like this the event streams are automatically generated.

As the documentation states:

> This stream strategy is the most performant of all, but it will create a lot of database tables, which is something not everyone likes (especially DB admins).

prooph/event-store-symfony-bundle/0.4/config/packages/prooph_event_store.yaml
```diff
@@ -0,0 +1,14 @@
+prooph_event_store:
+    stores:
+        default:
+            event_store: 'app.event_store.default'
+            # repositories:
```
Nyholm: Are these comments really needed? 
We usually recommend to put this in documentation instead.

.travis.yml
```diff
@@ -8,10 +8,12 @@ cache:
 matrix:
     fast_finish: true
     include:
-        - php: 7.1
-          env: SKELETON_VERSION="^3.4"
-        - php: 7.1
-          env: SKELETON_VERSION="^4.0"
+        -
```
Nyholm: This is only style changes. Please open a separate PR for this.

.travis.yml
```diff
@@ -25,4 +27,4 @@ install:
     - composer config extra.symfony.allow-contrib true
 
 script:
-    - composer req "${PACKAGES}"
+    - composer req $(echo "$PACKAGES" | tr --delete '\n')
```
Nyholm: Not sure this will work... The regex is just fetching one package...

prooph/event-store-bus-bridge/3.1/config/packages/prooph_event_store_bus_bridge.yaml
```diff
@@ -0,0 +1,19 @@
...
+        arguments:
+            - '@prooph_service_bus.default_event_bus'
+        tags:
+            - { name: 'prooph_event_store.default.plugin' }
+
+    # The app.event_store.default service must be configured with a
+    # Prooph\EventStore\TransactionalEventStore instance in order to user the following plugin.
+    # Read more: http://docs.getprooph.org/event-store/implementations/pdo_event_store/variants.html#3-2-1-2
+    #
+    # Prooph\EventStoreBusBridge\TransactionManager:
```
Nyholm: I do not like config in comments like this. If you think that 90% of the users will use this config you should un-comment this. If not I think these lines should be removed.

prooph/event-store-symfony-bundle/0.4/src/Command/CreateEventStreamCommand.php
```diff
@@ -0,0 +1,29 @@
...
+    protected function configure()
+    {
+        $this->setName('event-store:event-stream:create')
+            ->setDescription('Create event_stream.')
+            ->setHelp('This command creates the event_stream');
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $eventStore = $this->getContainer()->get('prooph_event_store.app');
```
Nyholm: Maybe use constructor injection instead?

prooph/service-bus-symfony-bundle/0.6/config/packages/prooph_service_bus.yaml
```diff
@@ -0,0 +1,25 @@
...
+    _defaults:
+        public: false
+
+    Prooph\ServiceBus\CommandBus: '@prooph_service_bus.default_command_bus'
+
+    # Autoregister your message handlers here
+    # Make sure App\Messaging\* services are not overriden in `config/services.yaml`
+    # Read more at https://github.com/prooph/service-bus-symfony-bundle/blob/master/doc/routing.md
+    #
+    # command_handlers:
```
Nyholm: Same here.. I think you should un-comment this.

.travis.yml
gquemener: I'm not sure to understand what you mean by "The regex is just fetching one package". For this PR, `$PACKAGES` contains:

```
prooph/event-store-bus-bridge:^3.1
prooph/event-store-symfony-bundle:^0.4
prooph/pdo-event-store:^1.7
prooph/service-bus-symfony-bundle:^0.6
``` 

Which [will not be handled properly](https://travis-ci.org/symfony/recipes-contrib/jobs/357352746#L593) if using `composer req "${PACKAGES}"`.

The piped `tr` will allows to merge lines into a single line:
```
prooph/event-store-bus-bridge:^3.1 prooph/event-store-symfony-bundle:^0.4 prooph/pdo-event-store:^1.7 prooph/service-bus-symfony-bundle:^0.6
``` 

which will be correctly handled by `composer req`.

ramondelafuente:
Any news on this one?

.travis.yml
Nyholm: Sorry for not being clear. `$PACKAGES` will never contain more that one package.

gquemener: Well, it does on this PR :thinking:

gquemener:
Ok, for handling the travis configuration in a separate PR.

I'm taking care of your other comments, thanks.

gquemener:
It's done @Nyholm, thanks for your comments :+1:

ramondelafuente:
👍 Yay!

prooph/pdo-event-store/1.7/config/packages/prooph_pdo_event_store.yaml
```diff
@@ -0,0 +1,22 @@
...
+        class: Prooph\EventStore\Pdo\MySqlEventStore
+        arguments:
+            - '@prooph_event_store.message_factory'
+            - '@app.event_store.pdo_connection.mysql'
+            - '@app.event_store.mysql.persistence_strategy'
+
+    app.event_store.pdo_connection.mysql:
+        class: \PDO
+        arguments:
+            - 'mysql:host=%env(MYSQL_HOST)%;dbname=%env(MYSQL_DBNAME)%'
```
Nyholm: You could use a environment variable like "MSQL_DSN" here instead.

Nyholm:
Any updates here?

kochen:
> The php doc seems to answer no to this question: http://php.net/manual/en/ref.pdo-mysql.connection.php (it might be worth trying it though).

This indeed seems to be the case.

> The factory method is a good idea indeed, but creates a dependency with the Doctrine bundle.

Then it's not worth fixing at the moment.


I suggest we go with what you have now and it needed/require we can always improve it later.

ramondelafuente:
It looks like everyone in this thread is happy at this point... Is there anything else blocking this now? Or maybe it's a release cycle that it's waiting on? (just asking)

gquemener:
Hello :wave: 

Considering this summer announcement<sup>1</sup> about dropping support on the prooph/service-bus package, I'd say this PR would need to be reworked before merged...

Any thought @UFOMelkor @codeliner?

Should it be simply closed?

---
<sup>1</sup> [here](https://www.sasaprolic.com/2018/08/the-future-of-prooph-components.html) and [here](https://www.sasaprolic.com/2018/08/the-future-of-prooph-components.html)

prolic:
No, it's still supported for a while and used a lot.

codeliner:
What @prolic said + the next service bus version could be a community driven project using symfony messenger but adding CQRS semantics on top of it. A future prooph pack could provide that. Anyway, service-bus is maintained until 31.12.2019.

OskarStark:
So IIRC this one is ready to merge?

fabpot:
If that's the case, then the `WIP` in the title must be removed.

gquemener:
Travis is failing because this PR contains recipes for many packages at the same time. I had provided [a solution](https://github.com/symfony/recipes-contrib/pull/320/commits/00ba2ea88da938e4a9fa2f5d2881bd8665b88040), but @Nyholm [asked me to revert it](https://github.com/symfony/recipes-contrib/pull/320#pullrequestreview-119643848).

I am not sure what to do to fix the build then :thinking:, would you have a suggestion @OskarStark ?

I'm removing the WIP flag

OskarStark:
Nice 👍

OskarStark:
The problem with travis could be, that you submitted many recipes in one PR and travis expects only one recipe per PR?

gquemener:
> The problem with travis could be, that you submitted many recipes in one PR and travis expects only one recipe per PR?

Yes it is. The `$PACKAGES` var will contain "new line" delimited packages list which `composer require` don't understand (it needs a space delimited packages list).

I managed to get this space delimited list by using `composer req $(echo "$PACKAGES" | tr --delete '\n')`

PR #321, opened 2018-03-19
Removing incompativle environment variable usage.

aequasi:
| Q             | A
| ------------- | ---
| License       | MIT


aequasi:
The way we were doing it never worked, just defaulting to array now, letting the user change it normally if they want. Doesn't need to be an environment variable.

PR #322, opened 2018-03-20
Add 1.5 version for Roukmoute Hashids bundle

roukmoute:
| Q             | A
| ------------- | ---
| License       | MIT

Valid for flex `^4.0`


Nyholm:
Thank you for this PR. I see that it is exactly that same recipe as for version 1.4. There is no need to add a recipe like this if nothing changed. 

Users that download 1.5 of your bundle will be able to use the 1.4 recipe. That is even true for users downloading 2.1. 

Im going to close this PR. If I missed something, please tell me and I'll be happy to open it again.

PR #326, opened 2018-03-23
Doctrine PHPCR dbal Pack

ElectricMaxxx:
start on working to get the doctrine-phpcr-dbal-pack running

| Q             | A
| ------------- | ---
| License       | MIT


ElectricMaxxx:
selection_205
uhh it looks the way we want it. yea.

But i am really unsure if i used the correct solution. @fabpot @nicolas-grekas should i create a pack instead? do i have the chance to create a pack in contrib section?

@dbu what do you think?

ElectricMaxxx:
The travis build fails, as `doctrine/phpcr-odm` isn't stable yet.

ElectricMaxxx:
@Nyholm this would be the follow up i said: https://github.com/symfony/recipes-contrib/pull/327

ElectricMaxxx:
@Nyholm having that modifying in head i would have one proposasal for flex: insteand of simply copying the config, we should start merging them in a kind of builder like we do in prepare method of our bundle extension. Just for the case we need it.

stof:
bundles are already merging the semantic config coming from multiple files (and they can control how the merging is done thanks to the Config component). I'm not sure we should try to have Flex doing a similar thing (which would break removal too)

doctrine/phpcr-dbal-symfony-pack/1.0/config/packages/doctrine_phpcr_dbal.yaml
```diff
@@ -0,0 +1,25 @@
...
+                nodes: doctrine_cache.providers.phpcr_nodes
+            parameters:
+                jackalope.check_login_on_server: false
+        workspace: '%env(PHPCR_WORKSPACE)%'
+        username: '%env(PHPCR_USER)%'
+        password: '%env(PHPCR_PASSWORD)%'
+    # enable the ODM layer
+    odm:
+        auto_mapping: true
+        auto_generate_proxy_classes: true'
```
covex-nn: Extra quotation mark

doctrine/phpcr-dbal-symfony-pack/1.0/manifest.json
```diff
@@ -0,0 +1,16 @@
+{
+    "bundles": {
+        "Doctrine\\Bundle\\DoctrineBundle\\DoctrineBundle": ["all"],
```
stof: this is wrong. You should not register bundles from other packages. The recipes of these other packages will already handle it (and this might cause issues with double-registration, or broken unregistration on removal)

ElectricMaxxx: you are right

Nyholm:
What is the status of this PR?

ElectricMaxxx:
@nyholm i will have a look for the red stuff, should be green as the bundle gone stable last week.

covex-nn:

ElectricMaxxx:
can somebody re run the tests please?

fabpot:
@ElectricMaxxx Done

ElectricMaxxx:
Thanks @fabpot ! @stof it looks green now, anything to add?

PR #327, opened 2018-03-23
Symfony CMF dbal Implementation

ElectricMaxxx:
create pack for symfony-cmf using doctrine phpcr with its dbal implem…entation

| Q             | A
| ------------- | ---
| License       | MIT

would fix: https://github.com/symfony-cmf/symfony-cmf/issues/259


symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
```diff
@@ -0,0 +1,24 @@
+framework:
```
Nyholm: You should not add config for other bundles.

symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
```diff
@@ -0,0 +1,24 @@
...
+    dynamic:
+        persistence:
+            phpcr:
+                enabled: true
+
+cmf_content:
+    persistence:
+        phpcr: ~
+
+cmf_core:
```
Nyholm: Why can't you put this is a symfony-cmf/core-bundle recipe?

dbu: the idea is to have one pack for all the cmf things. is that the wrong approach?

should we have packs for each bundle and then a meta-pack that requires all the other packs?

symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
```diff
@@ -0,0 +1,24 @@
...
+    multilang:
+        locales: [en,de]
```
dbu: this looks more like a demo example than a most common configuration. i think the pack should provide a minimal useful configuration, not contain demo things. as its an important feature, it could be there but commented out with some short explanation. afaik that is ok for recipies to provide commented out examples for the most important things.

dbu:
does this depend on #326, or how is phpcr set up?

symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
Nyholm: One pack to rule them all... Are the bundles to tightly coupled together? Say that I only want to install cmf-core, shouldn't I get some help from a recipe?

symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
Nyholm: >  i think the pack should provide a minimal useful configuration

Yes

> afaik that is ok for recipies to provide commented out examples for the most important things.

Generally no. I would rather see a link to some documentation. In rare cases I make exceptions, say when there is mandatory to add some config but the exact value is impossible to guess.

symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
ElectricMaxxx: it would be easier, when all other bundles would require `core-bundle` as a base container, which serves the dependencies. But that isn't. And hiding the the dependencies and the recipe in a package isn't that clean i think.

symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
ElectricMaxxx: So i would eventually point this one to an parameter array which is empty here.

ElectricMaxxx: So i would eventually point this one to an parameter array which is empty here.

symfony-cmf/cmf-phpcr-dbal-pack/1.0/config/packages/cmf_phpcr_dbal.yaml
dbu: most bundles depend on core-bundle being present, and that is modeled in composer.json - but from that, CmfCoreBundle is not instantiated in the kernel. afaik those bundles do not require services from core-bundle however, just php code.

the core-bundle handles sharing some configuration that is relevant for all cmf bundles through prepend. 

thus, separate packs would work. i just feel often you want all the main cmf bundles to have a more complete thing. can packs depend on other packs? then we could have separate packs for each bundle and this pack bring them all together.

ElectricMaxxx:
@Nyholm anything more i can do?

symfony-cmf/routing-bundle/2.0/config/packages/cmf_routing_bundle.yaml
```diff
@@ -1,11 +0,0 @@
-# Read the documentation: https://symfony.com/doc/master/cmf/bundles/routing/configuration.html
```
Nyholm: Why is the config for routing bundle removed in this PR?

PR #328, opened 2018-03-25
Added recipe for Buzz

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

Note, version 0.17 is not released yet.

### DIFF
--- a/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
+++ b/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
@@ -1,6 +1,5 @@
 # Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
 services:
-
     # Clients
     Buzz\Client\Curl:
         arguments:
--- a/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
+++ b/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
@@ -3,7 +3,7 @@ services:
     _defaults:
         public: false
 
-    # Tell the auto wire system to use "FuleGetContents" client as default client.
+    # Tell the auto wire system to use "FileGetContents" client as default client.
     Buzz\Client\BuzzClientInterface: '@Buzz\Client\FileGetContents'
 
     # Clients
--- a/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
+++ b/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
@@ -1,5 +1,8 @@
 # Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
 services:
+    _defaults:
+        public: false
+
     # Clients
     Buzz\Client\Curl:
         arguments:
--- a/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
+++ b/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
@@ -31,3 +31,5 @@ services:
 
     Buzz\Middleware\LoggerMiddleware:
         arguments: ['@?logger']
+        tags:
+            - { name: monolog.logger, channel: buzz }
--- a/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
+++ b/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
@@ -4,6 +4,7 @@ services:
         public: false
 
     # Tell the auto wire system to use "FileGetContents" client as default client.
+    # Set this to any of the other clients below (e.g. Curl or MultiCurl) to use a different default client.
     Buzz\Client\BuzzClientInterface: '@Buzz\Client\FileGetContents'
 
     # Clients
--- a/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
+++ b/kriswallsmith/buzz/0.17/config/packages/buzz.yaml
@@ -1,0 +1,28 @@
+# Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
+services:
+
+    # Clients
+    Buzz\Client\Curl:
+        arguments:
+            - timeout: 10
+
+    Buzz\Client\FileGetContents:
+        arguments:
+            - timeout: 10
+
+    Buzz\Client\MultiCurl:
+        arguments:
+            - timeout: 10
+
+    # Browser
+    Buzz\Browser:
+        arguments: ['@Buzz\Client\Curl']
+        calls:
+            - ['addMiddleware', ['@Buzz\Middleware\ContentLengthMiddleware']]
+            - ['addMiddleware', ['@Buzz\Middleware\LoggerMiddleware']]
+
+    # Middlewares
+    Buzz\Middleware\ContentLengthMiddleware: ~
+
+    Buzz\Middleware\LoggerMiddleware:
+        arguments: ['@?logger']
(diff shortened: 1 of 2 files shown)


Nyholm:

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
```diff
@@ -0,0 +1,27 @@
+# Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
+services:
```
stof: the file should have defaults to make services private

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
```diff
@@ -0,0 +1,27 @@
...
+    # Clients
+    Buzz\Client\Curl:
+        arguments:
+            - timeout: 10
+
+    Buzz\Client\FileGetContents:
```
stof: should all these classes be available for autowiring ? If no, you should rather give these another name.

stof:
Btw, should we actually create services for all clients ?

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
Nyholm: Yeah. I think so. One may only be interested in using clients.

Nyholm:
I think all clients should be registered as a service. I cannot decide what client that is "preferred" and there is no "cost" of register clients as services.

Nyholm:
Thank you for the review. I've updated the recipe

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
Nyholm: Fixed

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
stof: but then, they should still typehint the client interface, not specific implementation. currently, the interface is not autowirable here, while implementations are.

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
```diff
@@ -0,0 +1,30 @@
...
+        arguments:
+            - timeout: 10
+
+    Buzz\Client\MultiCurl:
+        arguments:
+            - timeout: 10
+
+    # Browser
+    Buzz\Browser:
+        arguments: ['@Buzz\Client\Curl']
```
stof: I would add a comment before this line to explain how to switch client (or before the client interface alias in case you create one to select the client)

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
```diff
@@ -0,0 +1,30 @@
...
+        calls:
+            - ['addMiddleware', ['@Buzz\Middleware\ContentLengthMiddleware']]
+            - ['addMiddleware', ['@Buzz\Middleware\LoggerMiddleware']]
+
+    # Middlewares
+    Buzz\Middleware\ContentLengthMiddleware: ~
+
+    Buzz\Middleware\LoggerMiddleware:
+        arguments: ['@?logger']
```
stof: what about adding the `monolog.logger` tag to log in a dedicated channel (`buzz` looks like a sensible default) instead of the default `app` one ? It will not hurt when using the fallback logger rather than MonologBundle.

stof:
@Nyholm the drawback of using the recipe to configure services rather than updating the BuzzBundle is that you cannot provide tag-based registering of middleware (which could be autoconfigured too). Not sure whether it is worth it though.

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
Nyholm: That is an excellent idea. Thanks!

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
Nyholm: So many great suggestions. Of course we should!

Nyholm:
> @Nyholm the drawback of using the recipe to configure services rather than updating the BuzzBundle is that you cannot provide tag-based registering of middleware (which could be autoconfigured too). Not sure whether it is worth it though.

You are correct. I cannot specify container injection tags without a bundle. But Im not sure I would like to have it though. We do not use it with HTTPlug bundle becuase it is rare (or is it?) that you would like the same middleware/plugin on all clients.

However, I do not plan to create (or maintain) a BuzzBundle. The already awesome HTTPlugBundle is compatible with Buzz. Or.. to be strict, it is Buzz that is compatible with HTTPlugBundle.

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
```diff
@@ -0,0 +1,35 @@
+# Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
+services:
+    _defaults:
+        public: false
+
+    # Tell the auto wire system to use "FuleGetContents" client as default client.
```
stof: typo in the comment

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
```diff
@@ -0,0 +1,35 @@
...
+    Buzz\Client\BuzzClientInterface: '@Buzz\Client\FileGetContents'
+
+    # Clients
+    Buzz\Client\Curl:
```
stof: I would not use class names as service ids for the client implementation, so that only the interface can be autowired.

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
```diff
@@ -0,0 +1,35 @@
+# Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
+services:
+    _defaults:
+        public: false
+
+    # Tell the auto wire system to use "FuleGetContents" client as default client.
+    Buzz\Client\BuzzClientInterface: '@Buzz\Client\FileGetContents'
```
stof: the previous version of the PR was using Curl as default implementation. Was it expected to change ?

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
Nyholm: But what if you really want a specific client? Say that for some requests you want the MultiCurl and for some you want normal Curl... 

Is there any performance issues in using class names as I do?

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
Nyholm: Good question. If I do `new Browser()` I will get `FileGetContents` as client. I thought the service definition should be the same. 

I think we use `FileGetContents` because that one "will always work". Curl has a dependency on a PHP extension.

kriswallsmith/buzz/0.17/config/packages/buzz.yaml
Nyholm: Fixed!

stof:
> Do you want me to update the middlewares as well?

well, middlewares are very unlikely to get typehinted in user code, so the argument about pushing towards best practices does not qualify here. The argument about cleaning the list of autowirable types still holds though.
I'm not strong on this change. @weaverryan what do you think ?

Nyholm:
I do only use auto wring for Browser and the BuzzClientInterface.

Im ready for a new review

Nyholm:
Thank you for the review

PR #329, opened 2018-03-27
Added nyholm/psr7

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

@stof Is this okey to "claim" all these Httplug and PSR (or soon to be PSR) interfaces. 

@dbu and @xabbuh might also be interested to see this.

### DIFF
--- a/nyholm/psr7/0.3/config/packages/nyholm_psr7.yaml
+++ b/nyholm/psr7/0.3/config/packages/nyholm_psr7.yaml
@@ -1,0 +1,30 @@
+# Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
+services:
+    _defaults:
+        public: false
+
+    # Register services for auto wire
+    Interop\Http\Factory\RequestFactoryInterface: '@nyholm.psr7.message_factory'
+    Interop\Http\Factory\ResponseFactoryInterface: '@nyholm.psr7.message_factory'
+    Http\Message\MessageFactory: '@nyholm.psr7.message_factory'
+    Interop\Http\Factory\ServerRequestFactoryInterface: '@nyholm.psr7.server_request_factory'
+    Http\Message\StreamFactory: '@nyholm.psr7.stream_factory'
+    Interop\Http\Factory\StreamFactoryInterface: '@nyholm.psr7.stream_factory'
+    Interop\Http\Factory\UploadedFileFactoryInterface: '@nyholm.psr7.uploaded_file_factory'
+    Http\Message\UriFactory: '@nyholm.psr7.uri_factory'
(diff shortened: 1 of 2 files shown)


stof:
hmm, claiming them as soon as this implementation is installed might cause issues, especially in case some libraries are forcing to use this implementation (Buzz 0.17 for instance).

And claiming the HTTPlug ones would create conflicts with HTTPlugBundle (as these claims would likely win over the bundle, breaking the fact that the bundle provides configuration for them)

Nyholm:
Yeah, I felt it could be kind of dangerous or lead to unwanted behaviors. 

For users that only have `nyholm/psr7` installed they would obviously like to add autowiring. That is why I leave these in the config.

Nyholm:
I usually ask people not to have config in docs like this. I want to argue that it makes sense it this case... 

@stof could give your opinion/review?

nyholm/psr7/0.3/config/packages/nyholm_psr7.yaml
```diff
@@ -0,0 +1,32 @@
+# Read more about client options at: https://github.com/kriswallsmith/Buzz/blob/master/doc/client.md
```
stof: this doc link looks wrong. It goes to a different package.

Nyholm: Thank you. Copy/paste error

nyholm/psr7/0.3/config/packages/nyholm_psr7.yaml
```diff
@@ -0,0 +1,31 @@
...
+    ## Register nyholm/psr7 services for autowiring with Interop\Http
+    #Interop\Http\Factory\RequestFactoryInterface: '@nyholm.psr7.message_factory'
+    #Interop\Http\Factory\ResponseFactoryInterface: '@nyholm.psr7.message_factory'
+    #Interop\Http\Factory\ServerRequestFactoryInterface: '@nyholm.psr7.server_request_factory'
+    #Interop\Http\Factory\StreamFactoryInterface: '@nyholm.psr7.stream_factory'
+    #Interop\Http\Factory\UploadedFileFactoryInterface: '@nyholm.psr7.uploaded_file_factory'
+    #Interop\Http\Factory\UriFactoryInterface: '@nyholm.psr7.uri_factory'
+
+    ## Register nyholm/psr7 services for autowiring with HTTPlug
+    #Http\Message\MessageFactory: '@nyholm.psr7.message_factory'
```
xabbuh: Wouldn't it make sense to add aliases for `RequestFactory` and `ResponseFactory` too?

Nyholm: Yeah, you might be correct. 
Thank you.

PR #331, opened 2018-03-27
[MsgPhpUserBundle] enable doctrine mapping

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

Looking at some other entity-providing bundles..., they all do this: enable doctrine mapping by default. It doesnt break without doctrine (it's metadata-ish).. so seems a good DX improvement :)

### DIFF
--- a/msgphp/user-bundle/0.2/src/Entity/User/User.php
+++ b/msgphp/user-bundle/0.2/src/Entity/User/User.php
@@ -2,11 +2,12 @@
 
 namespace App\Entity\User;
 
+use Doctrine\ORM\Mapping as ORM;
 use MsgPhp\User\Entity\User as BaseUser;
 use MsgPhp\User\UserIdInterface;
 
 /**
- * @final
+ * @ORM\Entity()
  */
 class User extends BaseUser
 {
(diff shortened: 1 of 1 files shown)


ro0NL:
failure fixed on my side :}

Nyholm:
Looks good. I’m just waiting on Travis.

ro0NL:
So am i :)

msgphp/user-bundle/0.2/src/Entity/User/User.php
```diff
@@ -2,14 +2,16 @@
...
 use MsgPhp\User\Entity\User as BaseUser;
 use MsgPhp\User\UserIdInterface;
 
 /**
- * @final
+ * @ORM\Entity()
  */
 class User extends BaseUser
 {
+    /** @ORM\Id() @ORM\Column(type="msgphp_user_id") */
```
Nyholm: I just saw this now. 

I’ve never seen two ORM at the same line. Does this work?

ro0NL: Yep :+1:

PR #334, opened 2018-03-28
update to 0.3

laurent-bientz:
| Q             | A
| ------------- | ---
| License       | MIT


wandi/easyadmin-plus-bundle/0.3/config/packages/wandi_easyadmin_plus.yaml
```diff
@@ -0,0 +1,12 @@
+#wandi_easy_admin_plus:
```
Nyholm: We should avoid having config in comments like this. I would prefer to see a link to the docs instead.

PR #337, opened 2018-04-04
[MsgPhpUserBundle] add post-install.txt

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/msgphp/user-bundle/0.3/post-install.txt
+++ b/msgphp/user-bundle/0.3/post-install.txt
@@ -4,5 +4,5 @@
 
   * Install the Symfony Maker Bundle using <info>composer require maker --dev</>. It's required for the next step.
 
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you. 
Maybe you should be a bit more verbose why I should install maker bundle?

Also, there is no need to move recipes from 0.2 to 0.3. You should **copy** or just modify the 0.2 version. It will apply on 0.3 automatically.

ro0NL:
Updated the post install text :) about 0.2 => 0.3; the `make:user` is only available on my side as of user-bundle:^0.3 (hence the rename). Again im killing 0.2 :)

I kept the eav bundle at 0.2. As you mentioned this will work fine with 0.3

Nyholm:
Sure sure. Could you resolve the merge conflict and I'll be happy to merge.

ro0NL:
Done,

PR #338, opened 2018-04-04
Recipe for ShivasVersioningBundle no longer required

mdriessen:
| Q             | A
| ------------- | ---
| License       | MIT

We were using the .env file as a replacement for the parameters.yml file in Symfony 3 to store and update the version number of the application. This worked fine for the development environment, but not for production.

So we modified the bundle to not rely on the .env file being present and as a side effect the console command doesn't need to be run anymore after each install. Since the bundle only needs to be registered now (which should happen by default), this recipe is no longer required.


stof:
Note that removing the recipe will not remove the configuration of running the command for existing projects using the old version of the recipe (as the recipe runs only on initial install of the bundle, not on updates)

stof:
btw, you just won a free review of your bundle improvements, to go further.

mdriessen:
Thanks for the review, I'll take a look at that! I realise that removing the recipe doesn't remove the configuration for existing projects, but since the bundle isn't working as intended on production even with the configuration that's probably not a problem (no .env file so bump currently doesn't do anything and no version number is loaded).

mdriessen:
@stof The bundle has been updated, could you merge this PR?

PR #341, opened 2018-04-10
 [MsgPhpEavBundle] Add ORM GeneratedValue annotation

ro0NL:
| Q             | A
| ------------- | ---
| License       | MIT

`msgphp_*_id` resolves to integer by default, so adding `@ORM\GeneratedValue()` seems a good DX improvement. At least it's more obvious to remove then to add.


fabpot:
@symfony-flex-server review

PR #346, opened 2018-04-13
Add recipe for sebtm/inwx-api-bundle

SebTM:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sebtm/inwx-api-bundle/0.1/manifest.json
+++ b/sebtm/inwx-api-bundle/0.1/manifest.json
@@ -1,0 +1,14 @@
+{
+  "bundles": {
+    "SebTM\\INWX\\InwxApiBundle": [
+      "all"
+    ]
+  },
+  "env": {
+    "INWX_API_ENVIRONMENT": "test",
+    "INWX_API_USERNAME": "dummy",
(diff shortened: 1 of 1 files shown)


sebtm/inwx-api-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+    "bundles": {
+        "SebTM\\INWX\\InwxApiBundle": [
```
Nyholm: Could you write this in one line?

SebTM:
Hey @Nyholm,

I directly access the values by "env"-function. Is there an example or documentation how this is done in best practice?

Thanks,
SebTM

Nyholm:
I think it is up to you and your bundle to decide. If you directly access the environment variables you need to require the DotEnv component. Your way of doing it is not at all wrong. But that is not how most bundles do it. 

Im find with approving this recipe if you like to. If you want to add a config file that is also very fine.

SebTM:
@Nyholm Would be ok if you merge it I will reconsider this for one of the next releases 👍

PR #347, opened 2018-04-14
Added recipe for pablok/supercache-bundle

PabloKowalczyk:
| Q             | A
| ------------- | ---
| License       | MIT


pablok/supercache-bundle/0.5/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "PabloK\\SupercacheBundle\\SupercacheBundle": ["all"]
```
Nyholm: Do we really need to enable this bundle in `"all"` environments?

PabloKowalczyk: IMO there is nothing wrong with this because user can disable caching functionality on config level, as done in `pablok/supercache-bundle/0.5/config/packages/dev/supercache.yaml`.

Nyholm: Yeah, I see. Sorry for not been more clear in my last comment. 

I meant: 
You can avoid having `supercache.enabled: false` in the dev config if you just enable this bundle in production. 

Im not sure what way is recommended. I just wanted to highlight that fact and see what you thought about it.

PabloKowalczyk: Bundle was enabled in all envs to avoid copying `cache_dir: '%kernel.project_dir%/public/webcache'` to every env config, but 99% of usages are only in `prod` env and i think it will be better to enable and configure bundle only in prod env.
Thanks for reminder :)

PR #351, opened 2018-04-19
Add SpecShaperEncryptBundle recipe

danabrey:
| Q             | A
| ------------- | ---
| License       | MIT


specshaper/encrypt-bundle/1.1/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+    "container": {
+        "encrypt_key": "%env(GDPR_ENCRYPT_KEY)%"
+    },
+    "env": {
+        "GDPR_ENCRYPT_KEY": "%generate(secret, 32)%"
```
Nyholm: Shouln't you pass this to your bundles config instead of adding it to the container like that?

Nyholm:
Any updates here?

danabrey:
Your comment is correct, that would be better. I submitted this from following the setup instructions from https://github.com/mogilvie/EncryptBundle, so it would be on @mogilvie to decide whether to make that change.

PR #355, opened 2018-04-26
Add sapient-bundle recipe

lepiaf:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/lepiaf/sapient-bundle/0.1/config/packages/sapient.yaml
+++ b/lepiaf/sapient-bundle/0.1/config/packages/sapient.yaml
@@ -1,0 +1,9 @@
+sapient:
+    # By default, bundle does nothing without initial configuration.
+    # Run command sapient:configure to get initial configuration.
+    # See http://sapient-bundle.readthedocs.io/en/latest/installation.html
+    guzzle_middleware:
+        unseal: false
+        verify: false
+    sealing_public_keys: ~
+    verifying_public_keys: ~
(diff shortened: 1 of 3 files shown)


Nyholm:
Thank you for the PR. I see that you enable the bundle for "prod" and "dev". Is there a reason why you did not use "all"?

lepiaf:
Hi, 

I did not use `all` because I thought it would be difficult to test functionally an API. For example, I want to test a `GET /api/user/1`, I want to check the json structure.

lepiaf:
I have a subscriber that listen on all kernel response and transform it to return an encrypted body.

Nyholm:
I see your reasoning. I would like my test environment to be as similar to production as possible. By disabling encryption on tests will fail to detect encryption related bugs. 

I would say that best practice is to enable the bundle on all environments. If you agree with me, then the recipe is not needed since Flex is smart enough to install this bundle anyways. 

If you **strongly** disagree, Im okey with merging the PR in its current state. 

What do you want to do?

lepiaf:
Ok, I agree with you.

One more question, does Symfony Flex create a config file for bundle? 

After installing bundle, user need to run a command to configure the bundle. I was thinking it might be easy to have an minimum config file and then copy the generated configuration.

For example, minimum config file do nothing. User will enable ability one by one.
```
sapient:
```

- http://sapient-bundle.readthedocs.io/en/latest/installation.html#symfony-4-without-recipe
- http://sapient-bundle.readthedocs.io/en/latest/reference.html

lepiaf:
I've added a post install message to tell user what to do next.

PR #360, opened 2018-05-11
Add recipe for phpfastcache/phpfastcache-bundle

Geolim4:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/phpfastcache/phpfastcache-bundle/3.0/config/packages/phpfastcache.yaml
+++ b/phpfastcache/phpfastcache-bundle/3.0/config/packages/phpfastcache.yaml
@@ -11,8 +11,3 @@ phpfastcache:
 #
 # See https://github.com/PHPSocialNetwork/phpfastcache-bundle/wiki for instructions on
 # how to configure the bundle.
-#
-#        memorycache:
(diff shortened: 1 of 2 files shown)


Geolim4:
Okay, Travis is failing because the bundle is not yet stable. Let's stand by until the first stable release.

phpfastcache/phpfastcache-bundle/3.0/config/packages/phpfastcache.yaml
```diff
@@ -0,0 +1,18 @@
...
+            type: Files
+            parameters:
+                path: "%env(PHPFASTCACHE_FILECACHE_PATH)%"
+
+# Define your drivers here. The example below connects to database 0 of the default Redis server.
+#
+# See https://github.com/snc/SncRedisBundle/blob/master/Resources/doc/index.md for instructions on
+# how to configure the bundle.
+#
+#        memorycache:
```
Nyholm: This is not needed. You have an example with the `filecache` which is great!

I suggest you to remove the section with `memorycache` but keep you comment and reference to the docs.

PR #392, opened 2018-06-07
Add recipe for redirection.io proxy

qboot:
| Q             | A
| ------------- | ---
| License       | MIT

This PR add a recipe for [redirection.io proxy](https://github.com/redirectionio/proxy-symfony)

### DIFF
--- a/redirectionio/proxy-symfony/0.1/config/packages/redirectionio.yaml
+++ b/redirectionio/proxy-symfony/0.1/config/packages/redirectionio.yaml
@@ -1,0 +1,5 @@
+# Documentation: https://github.com/redirectionio/proxy-symfony
+redirection_io:
+    connections:
+        my_agent: tcp://127.0.0.1:20301
+        # another_agent: unix:///var/run/redirectionio.sock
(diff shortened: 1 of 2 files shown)


redirectionio/proxy-symfony/0.1/config/packages/redirectionio.yaml
```diff
@@ -0,0 +1,7 @@
+# redirection.io Configuration
```
lyrixx: I don't think this is useful. The file is named `redirectionio.yaml`, and the main configuration key is `redirection_io`.
I would add a link to the documentation instead.

redirectionio/proxy-symfony/0.1/config/packages/redirectionio.yaml
```diff
@@ -0,0 +1,7 @@
...
+redirection_io:
+    connections:
+        # add here all your connection options as follow:
```
lyrixx: I would remove this comment. The config is already saying that.
And I would also remove the next comments **or** I would add the following comment:
`#another_agent: unix:///var/run/redirectionio.sock`

qboot:
@lyrixx I've updated my PR with your feedback 👍

PR #395, opened 2018-06-12
add hwi/oauth-bundle

andreybolonin:
| Q             | A
| ------------- | ---
| License       | MIT


hwi/oauth-bundle/0.6/config/packages/hwi_oauth.yaml
```diff
@@ -0,0 +1,29 @@
+hwi_oauth:
+    # list of names of the firewalls in which this bundle is active, this setting MUST be set
+    firewall_names: [secured_area]
+
+    # an optional setting to configure a query string parameter which can be used to redirect
+    # the user after authentication, e.g. /connect/facebook?_destination=/my/destination will
+    # redirect the user to /my/destination after facebook authenticates them.  If this is not
```
Nyholm: Do we really need all these comments and configuration also as a comment? I would like to see a link to the documentation instead. 

Also, remove all "inactive" configuration

hwi/oauth-bundle/0.6/post-install.txt
```diff
@@ -0,0 +1,20 @@
+<bg=blue;fg=white>                                          </>
+<bg=blue;fg=white> What's next for HWIOAuthBundle? </>
+<bg=blue;fg=white>                                          </>
+
+  * <fg=blue>Configure</> your application:
+    1. Enable sessions by uncommenting the entire session section in <comment>config/packages/framework.yaml</>
```
Nyholm: This instruction is not really reliable. Maybe write exactly what to do here or refer to documentation. 

> Enable sessions by adding `framework.session.enabled: true`  in <comment>config/packages/framework.yaml

andreybolonin:
@Nyholm done

hwi/oauth-bundle/0.6/config/packages/framework.yaml
```diff
@@ -0,0 +1 @@
+framework.session.enabled: true
```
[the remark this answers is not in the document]
Nyholm: We should not modify other bundles configuration. 

Please remove this. 
If it is really needed, write something in the post-install.

Nyholm:
Good. I like this. 

Could we provide a better example? When I try to install this I get the following error. 
screen shot 2018-06-22 at 18 06 34


Maybe we add "facebook" as default, What do you think?

andreybolonin:
@Nyholm yep

hwi/oauth-bundle/0.6/config/packages/hwi_oauth.yaml
```diff
@@ -0,0 +1,14 @@
+hwi_oauth:
+    # list of names of the firewalls in which this bundle is active, this setting MUST be set
+    firewall_names: [secured_area]
```
Nyholm: I like this comment. But may I suggest to change this value to "main"? The default security config looks like: 

```
security:
    # https://symfony.com/doc/current/security.html#where-do-users-come-from-user-providers
    providers:
        in_memory: { memory: ~ }
    firewalls:
        dev:
            pattern: ^/(_(profiler|wdt)|css|images|js)/
            security: false
        main:
            anonymous: true

            # activate different ways to authenticate

            # http_basic: true
            # https://symfony.com/doc/current/security.html#a-configuring-how-your-users-will-authenticate

            # form_login: true
            # https://symfony.com/doc/current/security/form_login_setup.html

    # Easy way to control access for large sections of your site
    # Note: Only the *first* access control that matches will be used
    access_control:
        # - { path: ^/admin, roles: ROLE_ADMIN }
        # - { path: ^/profile, roles: ROLE_USER }

```

There is no firewall named "secured_area". What do you think?

andreybolonin:
@Nyholm thank you for review )
what about to merge?

Nyholm:

Nyholm:
The review bot seams to be on vacation. I'll be back when it is fixed.

andreybolonin:

fabpot:
@symfony-flex-server review please (hey, that's me, be nice :))

PR #399, opened 2018-06-20
Move config to dev directory for nelmio/alice

ossinkine:
| Q             | A
| ------------- | ---
| License       | MIT

Move the config to a dev directory since the bundle is enabled only for dev and test environments.
@theofidry


ossinkine:
Initial MR #319

ossinkine:
This is a bad idea because config should be placed in test directory too

PR #401, opened 2018-06-25
Add the recipe for ticketing-bundle

Mapsred:
| Q             | A
| ------------- | ---
| License       | MIT


maps_red/ticketing-bundle/0.2/config/packages/ticketing.yaml
```diff
@@ -0,0 +1,58 @@
...
+    # Enable the seen/unseen display for tickets
+    enable_history:       true
+
+    # Set a role that will have access to the "staff" tickets
+    restricted_tickets_role: ROLE_USER
+
+    # if true, all new tickets will only be accessible to the owner and the restricted_ticket_role. If false, all new tickets will be public
+    enable_ticket_restriction: true
+    entities:
+        ticket:               App\Entity\Ticket
```
Nyholm: These entities does not exists. Maybe you should provide an implementation of some basic entities?

Mapsred: These entities are supposed to be created by the ``make:ticketing`` command from the bundle (in the manifest under composer-scripts) 
But is it better to initiate them directly in the recipe ?

Nyholm: Im not sure, you are way more knowledgeable about the bundle. 
But `make:ticketing` will run every time you do a composer update ([docs](https://github.com/symfony/recipes#composer-scripts-configurator)). Is that really want you want? 

By using the command to generate entities you get more control over the output. That is up to you. Just make sure the bundle does not complain if these entites does not exist.

Mapsred: Ah indeed, thanks for the review, I updated the recipe

PR #404, opened 2018-06-29
add felds/quick-mailer-bundle recipe

felds:
| Q             | A
| ------------- | ---
| License       | MIT


felds/quick-mailer-bundle/4.0/config/packages/quickmailer.yaml
```diff
@@ -0,0 +1,18 @@
...
+    from: { name: "Sender Name", email: "sender@example.com" }
+
+    # # Reply-to info (optional)
+    # reply_to: { name: "Reply-to Name", email: "reply-to@example.com" }
+
+    templates:
+        # # Short-form (enabled by default)
+        # demo: path/to/template.html.twig
+
+        # # Long-form
```
Nyholm: Do you really need both these examples? I think one is enough, the rest could be read in the docs.

felds/quick-mailer-bundle/4.0/config/packages/quickmailer.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+quickmailer:
+    # Sender info (required)
+    from: { name: "Sender Name", email: "sender@example.com" }
+
+    # # Reply-to info (optional)
+    # reply_to: { name: "Reply-to Name", email: "reply-to@example.com" }
+
+    templates:
+        # # Short-form (enabled by default)
```
Nyholm: Just write: 

This creates a template named “demo”

felds:
@Nyholm

Those are great tips! Thanks! I'll be using them!

About my last PR, I closed it because I changed something in the package itself and I had no way to re-run the `continuous-integration/symfony/pr` hook without changing something here (i think).

felds/quick-mailer-bundle/4.0/config/packages/quickmailer.yaml
```diff
@@ -0,0 +1,13 @@
+# For the full configuration, visit:
```
Nyholm: Could you rename this file to felds_quickmailer.yaml?

felds: then will we have to change the config root to `felds_quickmailer` too?

Nyholm: No no. Just the file name

felds: I find that a little misleading. Are there many packages using a file name different that the config root?
I assume the same conditions that could lead to a filename conflict would also break the config…

Nyholm: The config root has really nothing to do with this. The file name of the config file (in the recipe) should be prefixed with the organization. 

I see that this is not documented anywhere, but this is a practice done by many (most?) packages.

felds: good to know :)

I'll fix this in a minute

Nyholm:
I see you decided to change the config root anyways. Im fine with that, but it was not necessary.

felds:
-- Felds

PR #406, opened 2018-07-03
Adding netbull's recipe

netbull:
| Q             | A
| ------------- | ---
| License       | MIT


netbull/security-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "NetBull\\SecurityBundle\\NetBullSecurityBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "composer-scripts": {
+        "vendor/bin/browscap-php browscap:fetch": "php-script",
```
Nyholm: Could you tell me why these lines are needed?

netbull: the idea is to use browscap for creating browser fingerprint and store which will be used to ban after X attempts and for X seconds. Browscap need to create his cache after installing and that's why the commands are needed.

Nyholm: Yeah, I understand that. But I do not see that your bundle uses browscap other then in composer.json.

netbull: Ahh, I'm still developing... after 30 min it will be ready.. :/

netbull: now it should be fine :)

Nyholm:
Great. Can you make sure travis is happy? 
It seams like you depend on doctrine bundle.

netbull:
The Travis should be happy /at least for Symfony 4/, but I can not make rebuild without useless commit...

PR #411, opened 2018-07-05
Add support for NewRelic 2.0

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

We are about to release 2.0. It is currently in beta. 

This will fix https://github.com/ekino/EkinoNewRelicBundle/issues/179


Nyholm:
FYI @jderusse

ekino/newrelic-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Ekino\\NewRelicBundle\\EkinoNewRelicBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "#1": "Get the REST API key of your NewRelic Account on https://docs.newrelic.com/docs/apis/rest-api-v2/requirements/api-keys. Using an API key is optional.",
+        "NEWRELIC_API_KEY": "FIXME"
```
maxhelias: post-install ? For indicate the configuration to the user ?

Nyholm: Thank you. 
I decided not to since this is optional. 

It is only used if you want to use deployment notifications.

ekino/newrelic-bundle/2.0/config/packages/ekino_new_relic.yaml
```diff
@@ -0,0 +1,6 @@
+ekino_new_relic:
+    enabled: false
+    application_name: 'My Symfony Application' # Remove this to use the value from php.ini
+    api_key: "%env(NEWRELIC_API_KEY)%"
```
jderusse: I suggest to add a comment like `# This configuration is only required when using the "newrelic:notify-deployment"`

ekino/newrelic-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Ekino\\NewRelicBundle\\EkinoNewRelicBundle": ["all"]
```
jderusse: What's about `["prod"]` and moving `packages/ekino_new_relic.yaml` to `packages/prod/ekino_new_relic.yaml`?

ekino/newrelic-bundle/2.0/config/packages/ekino_new_relic.yaml
Nyholm: Yeah, but what if the bundle adds a new feature that uses the API key? 
I do not want to update the recipe then, right?

Or am I overthinking it?

ekino/newrelic-bundle/2.0/manifest.json
Nyholm: The 1.3 recipe did that. I chose not to for several reasons: 
- If you have invalid config you will only notice that in production
- You cannot write code that interact with services provided by the bundle. 
- We got a great BlackholeInteractor that we should use as often as we can. =)

Nyholm:

Nyholm:
Version 2.0 is finally released and this PR is ready for a final review.

PR #413, opened 2018-07-08
Add AdminLTE bundle 0.9

kevinpapst:
| Q             | A
| ------------- | ---
| License       | MIT

Adds the AdminLTE theme bundle 0.9

Testing this bundle:
```
composer config extra.symfony.allow-contrib true
export SYMFONY_ENDPOINT=https://symfony.sh/r/github.com/symfony/recipes-contrib/413
composer req "kevinpapst/adminlte-bundle:^0.9"
```
See https://symfony.sh/r/github.com/symfony/recipes-contrib/413

About AdminLTE bundle:
- Packagist: https://packagist.org/packages/kevinpapst/adminlte-bundle
- Repo: https://github.com/kevinpapst/AdminLTEBundle
- Demo app: https://github.com/kevinpapst/AdminLTEBundle-Demo

**EDIT:** Sorry for all the noise below, but I had some troubles with the composer dependencies (due to the 3.4 and 4.x demo applications) and had to create a couple of releases before it worked.


kevinpapst/adminlte-bundle/0.9/config/packages/admin_lte.yaml
```diff
@@ -0,0 +1,45 @@
+#
+# Configuration file for the AdminLTE bundle
+#
+# For more information about the bundle settings visit:
+# https://github.com/kevinpapst/AdminLTEBundle/blob/master/Resources/docs/configurations.md
+#
+admin_lte:
+    options:
```
Nyholm: Is this all the default config? 

There is no need to add that in a recipe. The bundle will work equally well without it.

kevinpapst/adminlte-bundle/0.9/config/packages/admin_lte.yaml
```diff
@@ -0,0 +1,45 @@
...
+        # boxed layout (true) or full-screen (false)
+        boxed_layout: false
+        # initial collapsed state of the sidebar
+        collapsed_sidebar: false
+        # whether the logo collapses or not
+        mini_sidebar: false
+        # configurable tabs in the control_sidebar (right screen), see:
+        # https://github.com/kevinpapst/AdminLTEBundle/blob/master/Resources/docs/control_sidebar.md
+        #control_sidebar:
+        #    home:
```
Nyholm: Is this something 80% of the user will uncomment and use? 

Im not a big fan of config in comments like this.

kevinpapst/adminlte-bundle/0.9/config/packages/admin_lte.yaml
kevinpapst: Hi @Nyholm - thanks for your review! 
Yes its the default config and it isn't necessarily needed. Personally I believe a full default config helps in setting up a module, but I can remove it if it isn't accepted otherwise. 

For sure every user will adapt at least some of the values in here. But I am fine with adapting it if this is against the official ruleset (which I might have missed, I couldn't find any recommendations in the docu about what should be in recipe config and what not). 

Is there any drawback from having a default config (besides compilation time in dev env)?

kevinpapst/adminlte-bundle/0.9/config/packages/admin_lte.yaml
kevinpapst: I have seen some of your your comments about "comments in configs" in other PRs as well. 

Two questions:
- If I remove the commented configs but leave the docu link, would that be okay?
- What about using `copy-from-package` instead of `copy-from-recipe`? I have seen that only a couple of recipes use it, but that would remove the "documentation" from the recipe repo. Would that be better?

kevinpapst:
Hi @Nyholm - thanks for your review! 
You are totally right, the config should NOT replace a good documentation ;-) that's why I have written quite some documentation:
- https://github.com/kevinpapst/AdminLTEBundle/tree/master/Resources/docs 

Especially covering the configuration:
https://github.com/kevinpapst/AdminLTEBundle/blob/master/Resources/docs/configurations.md
with sub-chapters for e.g. the control_sidebar (which is commented in the config right now).

As written in the other comments I believe that a good default configs helps in setting up the bundle, but I am totally fine adapting it to the guidelines! Are you against URLs in the config as well, to point the devs into the right direction?

Nyholm:
> Are you against URLs in the config as well, to point the devs into the right direction?

URLs are perfect to put here. 

> If I remove the commented configs but leave the docu link, would that be okay? Or would you rather remove both?

Doc links are very much appreciated. 

> What about using copy-from-package instead of copy-from-recipe?

The `copy-from-package` is used for code/files. Not really for config. It would be rare to use in your scenario. (https://github.com/symfony/recipes#copy-from-package-configurator)

> Is there any drawback from having a default config (besides compilation time in dev env)?

There is not extra compilation time in dev =)

The biggest drawback would be redundancy. And they are not really doing anything if the default config is in a recipe... If you removed it, nothing would change. 

See this PR as an example. It is the NewRelic bundle (https://github.com/symfony/recipes-contrib/pull/411/files)
There are a lot more configuration options but I believe they are not used by 80% of the users. But `exceptions` and `deprecations` are two config keys that I think are likely to be changed. 

------------

The idea of a recipe is that it should be a good template to start working with the bundle. The idea is not to expose all possible configuration. 
I do not know your bundle, but if you believe that `control_sidebar` is a config key most of your users will use and modify; then it make sense to have a good example setting.

kevinpapst:
Thanks for the clarification @Nyholm !

I adjusted the config and removed all inline docs and some default values. 
What is left are now the keys that either have to be configured by the developers or which will be adapted by > 80% of all users.

kevinpapst:
Thanks for accepting the PR! :-)

I have one question though which is left and hope you can clarify that:
Do I have to adjust the recipe for every release? Only for major releases? Or only if something fundamental (like new required configs) changed? 
Background: I don't understand how the version in the directory here is linked to the bundles release version. Does Flex search for the nearest available version in the recipe path? Is there a documentation available for recipe contributors?

kevinpapst:
I've read that part, the sentence `where version is the minimum version supported by the recipe` is a bit vague. I hope there won't be any breaking changes in the future (at least for the config).

Thanks again for all your support :)

PR #415, opened 2018-07-09
Rename imagine.yaml to liip_imagine.yaml

axzx:
| Q             | A
| ------------- | ---
| License       | MIT


xabbuh:
@fabpot Should we update the validator to check that the file name equals the root config node (I thought we were already doing that)?

axzx:
@Nyholm no, there are only 2 files in the config directory;)

Nyholm:
> @fabpot Should we update the validator to check that the file name equals the root config node (I thought we were already doing that)?

I think that is a good idea. It is one of the few things that I still have to check manually.

PR #419, opened 2018-07-12
Added recipe for the Comsave Salesforce-outbound-message-bundle

Dirk-Scholten:
| Q             | A
| ------------- | ---
| License       | MIT


Dirk-Scholten:
Am I supposed to include my config folder with my bundle? I thought this was bad practise? But I don't see how Travis would ever pass if I don't...

kevinpapst:
The problem is in your bundle. Your [composer.json](https://github.com/comsave/salesforce-outbound-message-bundle/blob/master/composer.json) is conflicting with the recipe integration builds.
There are two demo projects built with your bundle, both using PHP 7.1 (which already conflict with your PHP requirement, but that is ignored on Travis). 
One build is against Symfony 3.4 and one against Symphony 4.0, see these lines in the Travis build:
```
$ export SKELETON_VERSION="^3.4"
$ export SKELETON_VERSION="^4.0"
```
While building my own recipe I had the same problems and these requirements don't seem to be documented (I just figured from the Travis build). The only solution I could come up with: check if you really need these strict dependencies in your bundle and lower them if possible.

comsave/salesforce-outbound-message-bundle/0.3/config/comsave_outbound_message.yaml
```diff
@@ -0,0 +1,7 @@
+comsave_salesforce_outbound_message:
+    wsdl_directory: 'src/Company/Resources/wsdl'
+    document_paths:
+        Account:
+            path: 'src/Company/Document/Account'
+        Product2:
+            path: 'src/Company/Document/Product2'
```
Nyholm: These classes does not exist. Does it make sense to add them in the recipe?

Dirk-Scholten: That's correct, I don't think it makes a lot of sense to add them. They are very specific for each users own implementation. I just added the paths like this to give a fairly clear example of what a working configuration would look like. 

Does it make more sense to make them a more obscure example so it's more clear that they are just examples? Something like `path/to/document/Account`

PR #433, opened 2018-07-24
add the recipe for version 0.6

kasou:
| Q             | A
| ------------- | ---
| License       | MIT


wandi/easyadmin-plus-bundle/0.6/config/packages/wandi_easyadmin_plus.yaml
```diff
@@ -0,0 +1,60 @@
+#wandi_easy_admin_plus:
```
Nyholm: We should not have config in comments like this. Either remove it or uncomment it.

Nyholm:
Closing due inactivity. 

#eu-fossa

PR #443, opened 2018-07-31
Firewall was changed to main.

mathieutu:
Q | A
-- | --
License | MIT

In a previous commit the firewall was changed to main in the configuration. This should be changed too.


mathieutu:
I think travis is broken, I've just updated some text in post install...

PR #398, opened 2018-06-19
add dekalee mailjet bundle configuration

nicolasThal:
| Q             | A
| ------------- | ---
| License       | MIT

PR #445, opened 2018-08-02
WandiMailerBundle recipe

jacquesndl:
| Q             | A
| ------------- | ---
| License       | MIT


jacquesndl:
Hello, can an admin merge this PR ? No reply since Aug 2. Thanks

wandi/mailer-bundle/1.0/config/packages/wandi_mailer.yaml
```diff
@@ -0,0 +1,4 @@
+wandi_mailer:
+    sender:
+        name: 'example'
```
maxhelias: You can also use the env variable for define the name & address like this PR : https://github.com/symfony/recipes-contrib/pull/398/files and then explain the next step in the post-install

jacquesndl: @maxhelias thanks for the advice

wandi/mailer-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Wandi\\MailerBundle\\WandiMailerBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "WANDI_MAILER_SENDER_NAME": "Zinédine Zidane",
+        "WANDI_MAILER_SENDER_ADDRESS": "zinedine.zidane@france.fr"
```
maxhelias: I do not think it's a good idea to set these values ​​by default. Something more generic would be better like `test@test.com`

Nyholm: I agree with Max. Do a more generic example. That will make it more obvious that these values should be changes. 

After that, Im good with merging this PR.

jacquesndl: Ok, it's changed. Thanks.

jacquesndl:
Hello, can an admin merge this PR ? No reply since Nov 14th. Thanks

PR #447, opened 2018-08-04
add recipe for whiteoctober/WhiteOctoberPagerfantaBundle

gabiudrescu:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/white-october/pagerfanta-bundle/1.0/config/packages/white_october_pagerfanta_bundle.yaml
+++ b/white-october/pagerfanta-bundle/1.0/config/packages/white_october_pagerfanta_bundle.yaml
@@ -1,0 +1,4 @@
+white_october_pagerfanta:
+    exceptions_strategy:
+        out_of_range_page:        ~
+        not_valid_current_page:   ~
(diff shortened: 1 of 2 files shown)


gabiudrescu:
@Nyholm 

I have opened a PR to fix the problem with the missing dependency, waiting for it to be accepted and merged.

Regarding the configuration, the repo readme says, I quote:

```
This project is not yet configured with Symfony Flex, so this change to config/bundles.php won't be done automatically.
```

And I don't remember to have got this bundle automatically added to my bundle list when I installed it without my PR to flex. But I will try again.

My main desire was to get flex to create the pagerfanta.yaml on my behalf so I can customize at a later point.

Is there any chance to get this PR merged as it is?

Nyholm:
There is no difference for the user if we merge this PR or not. The bundle will still be installed and work exactly the same way, right?

emodric:
@Nyholm Currently, for some reason, the bundle is not autoactivated if installed with flex, no auto generated recipe is executed.

I haven't been able to deduce why.

emodric:
The reason was non standard PSR-4 config: https://github.com/whiteoctober/WhiteOctoberPagerfantaBundle/pull/205

Nyholm:
Great. I seen you use PSR-4 now. The bundle will be installed with Flex. We dont need a recipe with the default config. 

#eu-fossa

PR #448, opened 2018-08-07
add recipe for PUGXGeoFormBundle

garak:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/pugx/geo-form-bundle/0.2/config/packages/pugx_geo_form.yaml
+++ b/pugx/geo-form-bundle/0.2/config/packages/pugx_geo_form.yaml
@@ -1,0 +1,4 @@
+pugx_geo_form:
+    # see https://developers.google.com/maps/documentation/javascript/geocoding#GeocodingRequests for possible values for "region"
+    region: EU
+    useSsl: false
--- a/pugx/geo-form-bundle/0.2/config/packages/pugx_geo_form.yaml
+++ b/pugx/geo-form-bundle/0.2/config/packages/pugx_geo_form.yaml
@@ -1,4 +1,4 @@
 pugx_geo_form:
     # see https://developers.google.com/maps/documentation/javascript/geocoding#GeocodingRequests for possible values for "region"
     region: EU
-    useSsl: false
+    useSsl: true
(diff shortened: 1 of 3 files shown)


pugx/geo-form-bundle/0.2/config/packages/pugx_geo_form.yaml
```diff
@@ -0,0 +1,4 @@
+pugx_geo_form:
+    # see https://developers.google.com/maps/documentation/javascript/geocoding#GeocodingRequests for possible values for "region"
+    region: EU
+    useSsl: false
```
Nyholm: Shouldn't this be true by default?

garak: We always kept it false as default. I guess it's to avoid possible problems with ssl. Anyway I'm not really a maps expert and so I'm open to suggestions

Nyholm: At [geocoder-php](https://github.com/geocoder-php/Geocoder) we dropped all non-secure endpoints last year. There is no reason why one should use insecure endpoints. 

I understand why you keep the option but using safe default would be better. 

That is just an opinion, Im happy to merge now if you want to.

Nyholm:
Could you have a look at travis? 
It seams like there are some issues. Both with Symfony 4 and another bug.

garak:
It looks like my bundle depends on `%locale%` parameter, that's missing in travis build.
Is there a way to force such parameter in travis?
I found a way to replace such parameter in bundle's config when missing, but it's involving expression language component and I don't like to add such requirement just for a single parameter. Of course, if you know an alternate solution, let me know.

Nyholm:
You should not "force such parameter". 
If you create a new symfony project and then install this bundle you will get this error. The bundle needs to require the missing dependencies or make sure it does not fail without them.

garak:
Let me know if you want me to squash commits or if your can squash them from here. Thanks!

Nyholm:
Thank you

PR #451, opened 2018-08-15
Knpuniversity oauth2 client bundle

ckrack:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/knpuniversity/oauth2-client-bundle/1.20/config/packages/knpu_oauth2_client.yaml
+++ b/knpuniversity/oauth2-client-bundle/1.20/config/packages/knpu_oauth2_client.yaml
@@ -1,16 +1,11 @@
 knpu_oauth2_client:
-    # configure your clients as described here: https://github.com/knpuniversity/oauth2-client-bundle#configuration
     clients:
-        # # will create a service: "knpu.oauth2.client.facebook_main"
-        # first, run `composer require league/oauth2-facebook`
-        # facebook_main:
+        # configure your clients as described here: https://github.com/knpuniversity/oauth2-client-bundle#configuration
+        # make sure to composer require the package for the provider type
+        # clientname: 
         #     # this will be one of the supported types
-        #     type: facebook
-        #     client_id: '%env(FB_ID)%'
-        #     client_secret: '%env(FB_SECRET)%'
+        #     type: some_type
+        #     client_id: '%env(OAUTH_SOMETYPE_ID)%'
+        #     client_secret: '%env(OAUTH_SOMETYPE_SECRET)%'
         #     # the route that you're redirected to after
-        #     # see the controller example below
-        #     redirect_route: connect_facebook_check
-        #     # route parameters to pass to your route, if needed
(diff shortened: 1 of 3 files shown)


knpuniversity/oauth2-client-bundle/1.20/config/packages/knpu_oauth2_client.yaml
```diff
@@ -0,0 +1,16 @@
...
+        #     # this will be one of the supported types
+        #     type: facebook
+        #     client_id: '%env(FB_ID)%'
+        #     client_secret: '%env(FB_SECRET)%'
+        #     # the route that you're redirected to after
+        #     # see the controller example below
+        #     redirect_route: connect_facebook_check
+        #     # route parameters to pass to your route, if needed
+        #     redirect_params: {}
+        #     graph_api_version: v2.12
```
weaverryan: I think we should ONLY include the link, but *not* include all of the `facebook_main` and details below that. Those details could become out of date, and unless you are using Facebook exactly, they won't be useful.

knpuniversity/oauth2-client-bundle/1.20/config/packages/post-install.txt
```diff
@@ -0,0 +1,8 @@
+<bg=blue;fg=white>                                          </>
+<bg=blue;fg=white> What's next for KnpUOAuth2ClientBundle? </>
+<bg=blue;fg=white>                                          </>
+
+  * <fg=blue>Configure</> your oauth2 clients:
+    Add clients as described at <comment>https://github.com/knpuniversity/oauth2-client-bundle#configuration</> in your security firewalls section.
```
weaverryan: How about:

> * Download & configure your OAuth2 clients in `config/packages/knpu_oauth2_client` https://github.com/knpuniversity/oauth2-client-bundle#configuring-a-client

And I think we don't need the next bullet point below.

knpuniversity/oauth2-client-bundle/1.20/manifest.json
```diff
@@ -0,0 +1,14 @@
...
+	},
+	"copy-from-recipe": {
+			"config/": "%CONFIG_DIR%/"
+	},
+	"env": {
+			"#1": "Register for a facebook oauth app:",
+			"#2": "https://developers.facebook.com/apps/",
+			"FB_ID": "facebook-app-id",
+			"FB_SECRET": "facebook-secret"
+	}
```
weaverryan: Because we don't know what client they'll use, I think we shouldn't make any changes to the `.env` file. Instead, we should update the README to talk about this.

knpuniversity/oauth2-client-bundle/1.20/config/packages/knpu_oauth2_client.yaml
ckrack: Will update this, let's hope we get it through, as the guidelines say that there shouldn't be 'empty' configs.

PR #458, opened 2018-08-19
[OverblogGraphQLBundle] Init 0.12 and fix doc link in post install

mcg-web:
| Q             | A
| ------------- | ---
| License       | MIT


overblog/graphql-bundle/0.12/post-install.txt
```diff
@@ -0,0 +1,4 @@
+<bg=blue;fg=white> Next for OverblogGraphQLBundle </>
+    1. Define your schema, read documentation <comment>https://github.com/overblog/GraphQLBundle/blob/master/docs/definitions/index.md</>
```
Nyholm: Why not write "0.12" instead of master?

mcg-web: It's just because 0.12 is still under development, we removed a config entry since 0.12 so we must create a new recipe before merging the change.

PR #460, opened 2018-08-23
Add lireincore/imgcache-bundle recipe

lireincore:
| Q             | A
| ------------- | ---
| License       | MIT


lireincore/imgcache-bundle/0.1/config/packages/lireincore_imgcache.yaml
```diff
@@ -0,0 +1,14 @@
+lireincore_imgcache:
+    driver: gd
+#    srcdir: '%kernel.project_dir%/path/to/source/images'
+    destdir: '%kernel.project_dir%/public/thumbs'
+    webdir: '%kernel.project_dir%/public'
+#    baseurl: 'https:\\sitename.com'
```
OskarStark: you should remove the comments here, if you would like to provide additional information you can add a text file which will be prompted after the installation (and btw. it should be `https://`, not `\\`

Nyholm: I agree with @OskarStark Please remove the config in comments.

lireincore: done

lireincore/imgcache-bundle/0.1/post-install.txt
```diff
@@ -0,0 +1,5 @@
+<bg=blue;fg=white>                                   </>
+<bg=blue;fg=white> Next for LireinCoreImgCacheBundle </>
+<bg=blue;fg=white>                                   </>
+
+  * <fg=blue>Read</> the documentation at <comment>https://github.com/lireincore/LireinCoreImgCacheBundle/blob/master/README.md</>
```
fabpot: I would remove the post install file. We are only using it when there is something to do manually. Someone installing this bundle knows where the doc is.

lireincore: done

PR #426, opened 2018-07-23
[WIP] Added contao/core-bundle

sheeep:
| Q             | A
| ------------- | ---
| License       | MIT

PR #464, opened 2018-08-30
Add contao/core-bundle and contao/installation-bundle

leofeyer:
| Q             | A
| ------------- | ---
| License       | MIT

This PR replaces #426.


leofeyer:
The tests are failing, because `contao/core-bundle` requires `friendsofsymfony/http-cache`, which requires `php-http/client-implementation`. It is up to the end user to select a client implementation, therefore the packages only have a dev requirement for a concrete implementation, which however is not loaded by `composer require`. 😢

leofeyer:
@Nyholm Any idea how we could fix this?

PR #474, opened 2018-09-06
Added Yokai namespace packages

yann-eugone:
| Q             | A
| ------------- | ---
| License       | MIT


yann-eugone:
Packages may be more than simply a bundle : library with symfony Bundle integration.

As I see, with ApiPlatform, there is no problem with this

- [ApiPlatform recipe](https://github.com/symfony/recipes/blob/master/api-platform/core/2.1/manifest.json#L3)
- [ApiPlatform package type](https://github.com/api-platform/core/blob/2.1/composer.json#L3)

Do someone knows why it is required that the composer package is declared **as** a Symfony bundle ?

maxhelias:
I think this is not a requirement. It's for Flex in the case where there is no recipe, it will try to create one via the definition of the autoload with the `composer.json`

Can you confirm @Nyholm ?

Nyholm:
Yeah, That is correct.

yokai/messenger-bundle/1.0/config/packages/yokai_messenger.yaml
```diff
@@ -0,0 +1,46 @@
+yokai_messenger:
+    channels:
+        # Each channel has its own global configuration
+        swiftmailer:
```
Nyholm: Is this the same as default configuration?

yokai/enum-bundle/2.2/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "Yokai\\EnumBundle\\YokaiEnumBundle": ["all"]
```
Nyholm: Flex is smart enough to install this bundle without a recipe. Just make sure to specify "symfony-bundle" as a type in composer.json

yokai/messenger-bundle/1.0/config/packages/yokai_messenger.yaml
yann-eugone: The whole file is not the default configuration : because of the example, but yes, channels are all disabled by default (usage of `canBeEnabled` config helper)

yokai/enum-bundle/2.2/manifest.json
yann-eugone: I understand that. In the next major version of this repository, it will become less coupled to symfony, and transform to a library.

As for api platform, I was expecting to distribute everything in the same repository : the library, the symfony bridge, ...

I was expecting that i can add a bundle recipe for a package that is not marked as a bundle explicitely. But regarding the automated tests, I believe it is not possible. Do you confirm ?

Nyholm: > I was expecting that i can add a bundle recipe for a package that is not marked as a bundle explicitely.

You can, but that is an edge case. 

> But regarding the automated tests, I believe it is not possible. Do you confirm ?

Im not sure what you mean? The review bot will not be happy if you submit a `library`, that is correct.

Nyholm:

Nyholm:
I cannot approve this unless the review bot is happy. For example. You can never depend on `symfony/symfony`

yann-eugone:
closing. will split this in several PRs this next week

PR #482, opened 2018-09-09
add troopers/alertify-bundle recipe

lenybernard:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/troopers/alertify-bundle/3.0/config/packages/troopers_alertify.yaml
+++ b/troopers/alertify-bundle/3.0/config/packages/troopers_alertify.yaml
@@ -1,0 +1,10 @@
+#Please learn more about alertify contexts https://github.com/Troopers/AlertifyBundle/blob/master/Resources/doc/configuration.md
+troopers_alertify:
+    contexts:
+        front:
+            engine: "notie"
+            layout: "top"
+            translationDomain: "messages"
+            options:
+                alertTime: 10
+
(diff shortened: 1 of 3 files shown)


troopers/alertify-bundle/3.0/config/packages/troopers_alertify.yaml
```diff
@@ -0,0 +1,13 @@
+#troopers_alertify:
```
Nyholm: Please do not add configuration in comments like this. 
You should add some default config that is good for 80% of the users.

lenybernard: actually, there is already a default config (https://github.com/Troopers/AlertifyBundle/blob/master/DependencyInjection/Configuration.php#L30), I added this to help user to overwrite the configuration as this is a frequent need.

Nyholm: That is great!

But we should not bloat this configuration file with comments. Please add a link to your docs where the user can see how to configure the bundle. 

I would also like you to remove these config in comments before I can approve your recipe.

lenybernard: I've made some changes to avoid comments and add the default config. The travis seems to fail because of twig dependancy in `framework.yaml`

PR #484, opened 2018-09-11
[doctrine_mongodb] adding fallback env vars

Lctrs:
| Q             | A
| ------------- | ---
| License       | MIT


javiereguiluz:
Although the Symfony Bot says that the param should be add via "configurator", in the related Doctrine ORM recipe we don't do that (https://github.com/symfony/recipes/blob/528f95306348400bc55a50d883176c6a249c2256/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml#L1-L6) and it has been approved as an official recipe ... so maybe we can make an exception here too?

PR #486, opened 2018-09-15
NelmioAliceBundle > fix bug "There is no extension able to load the configuration for 'nelmio_alice'" in "prod" environment

kniziol:
| Q             | A
| ------------- | ---
| License       | MIT
| Environment       | prod

Required to avoid bug while running Symfony app in `prod` environment:

`Fatal error: Uncaught Symfony\Component\DependencyInjection\Exception\InvalidArgumentException: There is no extension able to load the configuration for "nelmio_alice" (in /var/www/application/config/packages/nelmio_alice.yaml). Looked for namespace "nelmio_alice", found "framework", "twig", "swiftmailer", "security", "doctrine_cache", "doctrine", "doctrine_migrations", "monolog", "sensio_framework_extra" in /var/www/application/vendor/symfony/config/Loader/FileLoader.php on line 168`

The `NelmioAliceBundle` is available in both `dev` and `test` environments. Is there any other possibility to not duplicate `nelmio_alice.yaml` configuration file in `config/packages/test` directory?


kniziol:
@theofidry I've placed the `nelmio_alice.yaml` configuration file in both `config/packages/dev/nelmio_alice.yaml` and  `config/packages/test/nelmio_alice.yaml` directories. What should I do more? :)

theofidry:
Erf sorry I forgot to ping you as well. That was all from me but I opened https://github.com/symfony/recipes-contrib/pull/487 as I didn't expect you to be that fast :)

The only small difference is that the `nelmio_alice.yaml` under `test` actually imports the one from `dev` to avoid duplicated content which could lead to confusion

kniziol:
@theofidry Import instead of duplicate - that's good idea. I solved it in the same way in my project, but doesn't commit with `import`. I don't know why. I pushed e69f062 right now to fix it 🤓

dunglas:
The Flex bot error is a false positive. This can be merged

PR #487, opened 2018-09-18
Register Nelmio Alice configuration only for dev and test environments

### LINKED ISSUE #455: Possible nelmio/alice recipie misconfiguration
I am new to this stuff so tell me if I am wrong.

Regarding https://github.com/symfony/recipes-contrib/tree/master/nelmio/alice/3.2

manifest.json says:
```` json
"bundles": {
        "Nelmio\\Alice\\Bridge\\Symfony\\NelmioAliceBundle": ["dev", "test"]
    },
````
But directory structure is `config/packages/nelmio_alice.yaml`. This mean that it will be on in all environments including `prod` which is not listed in `bundles` so it fails to load the configuration for 'nelmio_alice'.

Tell me if I am right and I will try to provide PR for it to fix.

theofidry:
| Q             | A
| ------------- | ---
| License       | MIT

Closes #486
Closes #455

The CI is failing because the library is of the type `library` instead of `bundle`.


dunglas:
The Flex bot error is a false positive.

PR #492, opened 2018-09-20
Added core23/shariff-bundle

core23:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/core23/shariff-bundle/3.0/config/packages/core23_shariff.yaml
+++ b/core23/shariff-bundle/3.0/config/packages/core23_shariff.yaml
@@ -1,0 +1,15 @@
+core23_shariff:
+    services:
+        facebook:
+            app_id:  '%env(FACEBOOK_API_ID)%'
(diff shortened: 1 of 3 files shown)


core23/shariff-bundle/3.0/config/packages/core23_shariff.yaml
```diff
@@ -0,0 +1,15 @@
...
+            app_id:  '%env(FACEBOOK_API_ID)%'
+            secret:  '%env(FACEBOOK_API_SECRET)%'
+
+sonata_block:
+    blocks:
+        core23_shariff.block.share:
+
+doctrine_cache:
+    providers:
+        core23_shariff:
```
Nyholm: It looks a bit weird that you've hard coded dependency to `doctrine_cache.providers.core23_shariff`. You could let the user provide an id to a cache service. 

(Also consider using PSR6)

PR #493, opened 2018-09-20
Added core23/gdpr-bundle

core23:
| Q             | A
| ------------- | ---
| License       | MIT


core23/gdpr-bundle/0.1/config/packages/core23_gdpr.yaml
```diff
@@ -0,0 +1,3 @@
+sonata_block:
+    blocks:
+        core23_gdpr.block.information: ~
```
fabpot: Looks like it depends on a dep that depends on symfony/templating, which is not supported by Flex.

PR #507, opened 2018-09-23
Add a recipe for dunglas/doctrine-json-odm

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/dunglas/doctrine-json-odm
A recipe is mandatory because the bundle is provided in the library, thus cannot be auto-discovered by Flex.

### DIFF
--- a/dunglas/doctrine-json-odm/0.1/post-install.txt
+++ b/dunglas/doctrine-json-odm/0.1/post-install.txt
@@ -1,6 +1,0 @@
-<bg=blue;fg=white>              </>
-<bg=blue;fg=white> What's next? </>
-<bg=blue;fg=white>              </>
-
-  * The <comment>json_document</> column type is now available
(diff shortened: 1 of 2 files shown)


dunglas:
Failures are intended. I tested the recipe locally and it works.

dunglas/doctrine-json-odm/0.1/post-install.txt
```diff
@@ -0,0 +1,6 @@
+<bg=blue;fg=white>              </>
+<bg=blue;fg=white> What's next? </>
+<bg=blue;fg=white>              </>
+
+  * The <comment>json_document</> column type is now available
+  * Go to https://github.com/dunglas/doctrine-json-odm for examples and docs
```
fabpot: I would remove this post install as it does not give any useful information the user won't already have.

dunglas: done

PR #508, opened 2018-09-24
Add a recipe for sylius/refund-plugin

Zales0123:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sylius/refund-plugin/0.4/config/packages/sylius_refund.yaml
+++ b/sylius/refund-plugin/0.4/config/packages/sylius_refund.yaml
@@ -1,0 +1,2 @@
+imports:
+    - { resource: "@SyliusRefundPlugin/Resources/config/app/config.yml" }
--- a/sylius/refund-plugin/0.4/config/routes/sylius_refund.yaml
+++ b/sylius/refund-plugin/0.4/config/routes/sylius_refund.yaml
@@ -1,0 +1,2 @@
+sylius_refund:
+    resource: "@SyliusRefundPlugin/Resources/config/routing.yml"
--- a/sylius/refund-plugin/0.4/manifest.json
+++ b/sylius/refund-plugin/0.4/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Sylius\\RefundPlugin\\SyliusRefundPlugin": ["all"]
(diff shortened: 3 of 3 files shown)


pamil:
For Sylius plugin we use `sylius-plugin` as a Composer package type instead of `symfony-bundle` in order to get them more discoverable at [Packagist](https://packagist.org/explore/?type=sylius-plugin). However, the Flex recipe validation prohibits that, is there any possibility to allow other package types?

Zales0123:
Up ^^ We would really like to make our plugins accessible via SymfonyFlex :)

fabpot:
I've just added support for `sylius-plugin` as a synonym for `symfony-bundle` in the Symfony Flex server, so you can switch back to it if you want to.

sylius/refund-plugin/0.4/config/packages/sylius_refund.yaml
```diff
@@ -0,0 +1,2 @@
+imports:
+    - { resource: "@SyliusRefundPlugin/Resources/config/app/config.yml" }
```
fabpot: Out of curiosity, why is it not imported automatically by the bundle?

pamil: Do you mean by loading it in a DI extension?

nicolas-grekas: yes, that :)

pamil: This is a file containing the configuration for other bundles, like `sylius_grid`, `sylius_resource` or `winzou_state_machine` etc.

Loading it in extension's `load` method would not work as it does not have access to other bundles. Loading it in extension's `prepend` method would work, but then we had an issue with end-user trying to add a different config for another bundle having `disallowNewKeysInSubsequentConfigs` in configuration or no way to delete / reorder a collection of items in config. 

Importing the bundle's config file explicitly in the end-user config makes it possible to copy that config and customise it to one's need, while still allowing for config changes in the new releases. Also, it uses YAML which everyone is familiar with instead of arrays in PHP in DI extension, but that's only a side-effect.

PR #512, opened 2018-09-26
Add a recipe for sylius/admin-order-creation-plugin

Zales0123:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sylius/admin-order-creation-plugin/0.6/manifest.json
+++ b/sylius/admin-order-creation-plugin/0.6/manifest.json
@@ -4,9 +4,5 @@
     },
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/"
-    },
-    "composer-scripts": {
-        "assets:install --symlink %PUBLIC_DIR%": "symfony-cmd",
-        "cache:clear": "symfony-cmd"
     }
 }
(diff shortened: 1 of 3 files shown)


sylius/admin-order-creation-plugin/0.6/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "Sylius\\AdminOrderCreationPlugin\\SyliusAdminOrderCreationPlugin": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "composer-scripts": {
+        "assets:install --symlink %PUBLIC_DIR%": "symfony-cmd",
+        "cache:clear": "symfony-cmd"
```
fabpot: These 2 lines should be removed as there are already part of framework bundle.

pamil: Indeed, it has resulted in two `assets:install` commands.

Zales0123: 👍 Fixed 🚀

PR #529, opened 2018-10-17
Add EWZRecaptchaBundle

franmomu:
| Q             | A
| ------------- | ---
| License       | MIT


excelwebzone/recaptcha-bundle/1.5/config/packages/ewz_recaptcha.yaml
```diff
@@ -0,0 +1,4 @@
+# See https://github.com/excelwebzone/EWZRecaptchaBundle for full configuration
+ewz_recaptcha:
+    public_key:  '%env(GOOGLE_RECAPTCHA_SITE_KEY)%'
```
franmomu: I'm not sure about this, I reused the env vars defined in `google/recaptcha` because it is a dependency, but I don't know if it should rely on those variables or define their own.

Nyholm: I agree with you. 
What if the Google/recaptcha changes their recipe. I think you should define your own env variables and then let the user change them. That will make the recipe more unit.

PR #550, opened 2018-10-27
updating msgphp user receipt for mysql/maria db as well as msgphp pull 214

pascalwacker:
| Q             | A
| ------------- | ---
| License       | MIT

Key columns only support a max length of 191 characters in mysql/maria db on default, using the `utf8mb4` charset and `utf8mb4_unicode_ci` collate.

This change ensures that msgphp user bundle will work out of the box as well as prepare for https://github.com/msgphp/msgphp/pull/214


ro0NL:
@pascalwacker can you update eav-bundle as well please :)

PR #558, opened 2018-11-12
Add recipe for scheb/two-factor-bundle

scheb:
| Q             | A
| ------------- | ---
| License       | MIT

Recipe for version 3.x of https://github.com/scheb/two-factor-bundle


scheb:
Decided to go with the security_tokens configuration (since that's relatively specific to the application) and the routes.

PR: https://github.com/symfony/recipes-contrib/pull/558

PR #560, opened 2018-11-20
sylius/refund-plugin recipe improvements

Zales0123:
| Q             | A
| ------------- | ---
| License       | MIT

It is meant to automatize 3rd point from [installation instruction](https://github.com/Sylius/RefundPlugin/blob/master/README.md#installation).

### DIFF
--- a/sylius/refund-plugin/0.4/manifest.json
+++ b/sylius/refund-plugin/0.4/manifest.json
@@ -4,5 +4,8 @@
     },
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/"
+    },
+    "copy-from-package": {
+        "src/Resources/views/SyliusAdminBundle/": "templates/bundles/SyliusAdminBundle/"
     }
 }


sylius/refund-plugin/0.4/manifest.json
```diff
@@ -1,8 +1,18 @@
 {
     "bundles": {
+        "Sylius\\Bundle\\CoreBundle\\SyliusCoreBundle": ["all"],
+        "Sylius\\Bundle\\GridBundle\\SyliusGridBundle": ["all"],
+        "Sylius\\Bundle\\MailerBundle\\SyliusMailerBundle": ["all"],
+        "Sylius\\Bundle\\ResourceBundle\\SyliusResourceBundle": ["all"],
```
Nyholm: These bundles is not part of the package. We should never modify other bundles in this recipe.

Nyholm:
Any updates here?

Nyholm:
Friendly ping

Zales0123:
Sorry for the long response time, I've totally missed it :/ I had no time to work on this for some time recently, but for sure I want to make it work and will try as fast as possible :)

Zales0123:
Ok, so it's been a long time, sorry for that 🖖 Right now it won't work, as the plugin requires Sylius to work and Sylius does not have a recipe defined... So I'll allow myself to close it (in the end, it's just an improvement), to not litter the PR's list and when we would have a recipe for Sylius and/or its bundles itself (I hope sooner than later ;))  I'll come back with it. Thanks a lot for Flex and Symfony! 🚀

PR #565, opened 2018-11-23
Add receipes for Enqueue 0.9.

makasim:
| Q             | A
| ------------- | ---
| License       | MIT

fixes https://github.com/php-enqueue/enqueue-dev/issues/600

### DIFF
--- a/enqueue/enqueue-bundle/0.9/config/packages/test/enqueue.yaml
+++ b/enqueue/enqueue-bundle/0.9/config/packages/test/enqueue.yaml
@@ -1,0 +1,5 @@
+enqueue:
+    default:
+        transport: 'null:'
+        client:
+            traceable_producer: true
--- a/enqueue/wamp/0.9/manifest.json
+++ b/enqueue/wamp/0.9/manifest.json
@@ -1,0 +1,5 @@
+{
+    "env": {
+        "ENQUEUE_DSN": "wamp:"
+    }
+}
--- a/enqueue/async-command/0.9/config/packages/enqueue_async_command.yaml
+++ b/enqueue/async-command/0.9/config/packages/enqueue_async_command.yaml
@@ -1,0 +1,3 @@
+enqueue:
+    default:
+        async_commands: true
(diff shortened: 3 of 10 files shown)


makasim:
@Nyholm not yet

enqueue/enqueue-bundle/0.9/config/packages/test/enqueue.yaml
```diff
@@ -0,0 +1,5 @@
+enqueue:
+    default:
+        transport: 'null:'
```
OskarStark: `null:` or `null://`

enqueue/wamp/0.9/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "env": {
+        "ENQUEUE_DSN": "wamp:"
```
OskarStark: `wamp:` or `wamp://` ?

OskarStark:
Can someone please rerun the travis job?

enqueue/enqueue-bundle/0.9/config/packages/test/enqueue.yaml
makasim: `null:` is a valid scheme.

OskarStark: I know, but then we should use it here too:
https://github.com/symfony/recipes-contrib/pull/565/files#diff-d1033b7e1d7ea30fd6e9eae6a48948e8R9

I am against mixing it with or without `//`

Nyholm:
I reran travis. I know that we recently made an update to support multiple packages in one PR. A rebase will fix that.

Nyholm:
When I test this I get: 

>  A given scheme "wamp" is not supported. Maybe it is a custom driver, make sure you registered it with "Enqueue\Client\Resources::addDriver".

Is that expected? If I removed the second `ENQUEUE_DSN` from my `.env` it all works great.

OskarStark:
Yes I will fix that in another PR, we only want to expose the ENQUEUE_ENV with the bundle and not for each transport... 

see https://github.com/symfony/recipes-contrib/pull/568#issuecomment-442729478 and the following comments

PR #568, opened 2018-11-29
support enqueue 0.9

OskarStark:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/enqueue/enqueue-bundle/0.9/config/packages/enqueue.yaml
+++ b/enqueue/enqueue-bundle/0.9/config/packages/enqueue.yaml
@@ -1,0 +1,4 @@
+enqueue:
+    default:
+        transport: '%env(ENQUEUE_DSN)%'
+        client: ~
--- a/enqueue/enqueue-bundle/0.9/config/packages/test/enqueue.yaml
+++ b/enqueue/enqueue-bundle/0.9/config/packages/test/enqueue.yaml
@@ -1,0 +1,5 @@
+enqueue:
+    default:
+        transport: 'null://'
+        client:
+            traceable_producer: true
--- a/enqueue/enqueue-bundle/0.9/manifest.json
+++ b/enqueue/enqueue-bundle/0.9/manifest.json
@@ -1,0 +1,11 @@
+{
+    "bundles": {
+        "Enqueue\\Bundle\\EnqueueBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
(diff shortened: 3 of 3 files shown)


Nyholm:
Isn’t this a duplicate of #565?

OskarStark:
> Isn’t this a duplicate of #565?

You are absolutely right, thank you!
We can close it here, but on question, how is the best practice:

We have enqueue-bundle which installs added this env: `ENQUEUE_DSN` and now, for every transport I install, I get a new part in .env.dist file and a new defined `ENQUEUE_DSN`. This can result in weird results, because the last registered ENV "wins":

```
###> enqueue/enqueue-bundle ###
ENQUEUE_DSN=amqp+ext://guest:guest@rabbitmq:5672
###< enqueue/enqueue-bundle ###

###> enqueue/null ###
ENQUEUE_DSN=null://
###< enqueue/null ###

###> enqueue/sqs ###
ENQUEUE_DSN=sqs://
###< enqueue/sqs ###
```

In this case it uses the sqs one.... which is not properly configured and results in an exception.

Imo we should only expose/create/add this specific ENV only for the bundle itself.
For the transports it looks more like a documentation on how this DSN should start to get it to work, but a text-output after adding a transport looks more appropriate to me.

//cc @makasim

makasim:
The alternative would be to use separate DSN names and the bundle config, for example, AMQP_DSN, REDIS_DSN, SQS_DSN and so no and config:

```
enqueue:
    amqp:
        transport: '%env(AMQP_DSN)%'
```

There some disadvantages too:

* It's not that clear which one is a default
* Many config files 
* DSN names could collide with other recipes.

OskarStark:
My proposal would be to only use `ENQUEUE_DSN` which avoids collisions with other envs/bundles and after installing a transport show a message like:

```
To use SQS transport, update ENQUEUE_DSN like this:

    ENQUEUE_DSN=sqs://.....

```

makasim:
I like it.

OskarStark:
> I like it.
😄 

I would provide a PR after #565 was merged and will do this for 0.8 version, too

makasim:
@Nyholm it's ready to go.

makasim:
@Nyholm what can I do to get this merged?

Nyholm:
Great. Thank you

OskarStark:
Please merge this instead: https://github.com/symfony/recipes-contrib/pull/565

Nyholm:
> Please merge this instead: #565

Thank you. I just read them both and was just about to ask.

PR #576, opened 2018-12-06
Add mailjet/mailjet-bundle recipe

lbassin:
| Q             | A
| ------------- | ---
| License       | MIT

Related : https://github.com/mailjet/mailjetBundle/issues/7

### DIFF
--- a/mailjet/mailjet-bundle/1.0/config/packages/mailjet.yaml
+++ b/mailjet/mailjet-bundle/1.0/config/packages/mailjet.yaml
@@ -1,4 +1,4 @@
 mailjet:
-    api_key:    "%env(MJ_APIKEY_PUBLIC)%"
-    secret_key: "%env(MJ_APIKEY_PRIVATE)%"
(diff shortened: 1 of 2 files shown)


mailjet/mailjet-bundle/1.0/config/packages/mailjet.yaml
```diff
@@ -0,0 +1,4 @@
+mailjet:
+    api_key:    "%env(MJ_APIKEY_PUBLIC)%"
```
OskarStark: could you change `MJ` to `MAILJET`?

Nyholm: Yeah, It is better to be explicit.

lbassin: You're right! It's fixed :)

lbassin:
Hi @Nyholm, are the changes good for you ?

PR #579, opened 2018-12-06
Add disjfa/translation bundle

disjfa:
| Q             | A
| ------------- | ---
| License       | MIT


disjfa:
Always like it when i push, always get new ways travis is shouting at me.

Made a new translation bundle which is used for the user. You can extend the translations from the base yaml files and cache the customer changes to the database. So when the customer wants to change the translations every day they actually can. Cache afterwards from the database, in a live site.

The checks have now passed, awesome! Cheers.

PR #582, opened 2018-12-08
Fix the travis build when changing multiple recipes

stof:
| Q             | A
| ------------- | ---
| License       | MIT

The different packages need to be passed as separate arguments to `composer require`
#SymfonyConHackday2018


fabpot:

PR #584, opened 2018-12-10
use post-install.txt instead of manifest.json

OskarStark:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/symfony/recipes-contrib/pull/565

cc @makasim

### DIFF
--- a/enqueue/amqp-bunny/0.8/post-install.txt
+++ b/enqueue/amqp-bunny/0.8/post-install.txt
@@ -1,0 +1,5 @@
+<bg=blue;fg=white>               </>
+<bg=blue;fg=white> New transport </>
+<bg=blue;fg=white>               </>
+
+  * the DSN used in ENQUEUE_DSN should start with <fg=green>amqp+bunny://</>.
--- a/enqueue/amqp-bunny/0.8/manifest.json
+++ b/enqueue/amqp-bunny/0.8/manifest.json
@@ -1,5 +1,0 @@
-{
-    "env": {
-        "ENQUEUE_DSN": "amqp+bunny:"
-    }
-}
--- a/enqueue/amqp-ext/0.8/manifest.json
+++ b/enqueue/amqp-ext/0.8/manifest.json
@@ -1,5 +1,0 @@
-{
-    "env": {
-        "ENQUEUE_DSN": "amqp+ext:"
-    }
-}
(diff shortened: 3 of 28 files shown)


enqueue/amqp-bunny/0.8/post-install.txt
```diff
@@ -0,0 +1,5 @@
+<bg=blue;fg=white>               </>
+<bg=blue;fg=white> New transport </>
+<bg=blue;fg=white>               </>
+
+  * the DSN used in ENQUEUE_DSN should start with <fg=green>amqp+bunny://</>.
```
OskarStark: please give me some feedback on the post-install.txt

OskarStark:
Hmm, shall I add empty manifest.json @Nyholm ?

Nyholm:
Thank you. 
I see that you removed the manifest file for amqp-bunny:0.8. Same for other packages. You also have to remove the package for amqp-bunny:0.7 or Flex will automatically use that recipe instead of the 0.8 one.

Nyholm:
> Hmm, shall I add empty manifest.json @Nyholm ?

That seams even weirder... Can I install the bundle 0.9 with amqp-bunny:0.7? Probably, right? If so, I would remove all recipes for adapters.

OskarStark:
> I see that you removed the manifest file for amqp-bunny:0.8. Same for other packages. You also have to remove the package for amqp-bunny:0.7 or Flex will automatically use that recipe instead of the 0.8 one.

good to know 👍 

> That seams even weirder... Can I install the bundle 0.9 with amqp-bunny:0.7? Probably, right? If so, I would remove all recipes for adapters.

Hmm yes, but in this case everybody needs to go to the docs and check how the DSN should start.
would be great if one could only add an `info.txt` for such information.

Nyholm:
Hm.. Yeah, a recipe is never complete without a `manifest.json`. So I would assume that it would ignore the "recipe" in `enqueue/amqp-bunny/0.8`.

> but in this case everybody needs to go to the docs and check how the DSN should start.

True, but that is probably where you find out about the transport anyways, right? Ie https://github.com/php-enqueue/enqueue-dev/blob/master/docs/transport/amqp_bunny.md

OskarStark:
Yes, maybe you are right and we should remove the recipes for the transports

Nyholm:
Sorry Oskar, I missed this PR when merging https://github.com/symfony/recipes-contrib/pull/596#pullrequestreview-190597289.

I'll close this. Please let me know if there is still something in here that should be merged.

OskarStark:
No everything looks fine 👌🏻

PR #587, opened 2018-12-12
add goksagun/scheduler bundle recipie

goksagun:
| Q             | A
| ------------- | ---
| License       | MIT


goksagun/scheduler-bundle/0.1/config/packages/scheduler.yaml
```diff
@@ -0,0 +1,17 @@
...
+#   -    -    -    -    -
+#   |    |    |    |    |
+#   |    |    |    |    |
+#   |    |    |    |    +----- day of week (0 - 7) (Sunday=0 or 7)
+#   |    |    |    +---------- month (1 - 12)
+#   |    |    +--------------- day of month (1 - 31)
+#   |    +-------------------- hour (0 - 23)
+#   +------------------------- min (0 - 59)
+scheduler:
+    enable: true
```
OskarStark: You should only provide non default values here

PR #591, opened 2018-12-28
Added recipe for Buzz 1.0

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

Buzz is currently in beta-2 which will (hopefully) be the final beta version. 

This PR is an updated recipe from version 0.17. There are only minor changes. 

~~The test will fail because we are missing a PSR-17 service.~~


kriswallsmith/buzz/1.0/config/packages/buzz.yaml
```diff
@@ -0,0 +1,54 @@
...
+    Psr\Http\Client\ClientInterface: '@Buzz\Client\BuzzClientInterface'
+
+    # Clients
+    buzz.client.file_get_contents:
+        class: Buzz\Client\FileGetContents
+        arguments:
+            - '@Psr\Http\Message\ResponseFactoryInterface'
+            -
+                timeout: 10
+                allow_redirects: false
```
Nyholm: These options are likely to change and the format is non-trivial. That is why I keep them here.

Nyholm:
@lyrixx could I get your review please?

kriswallsmith/buzz/1.0/config/packages/buzz.yaml
```diff
@@ -0,0 +1,54 @@
...
+    # Tell the auto wire system to use "FileGetContents" client as default client.
+    # Set this to any of the other clients below (e.g. Curl or MultiCurl) to use a different default client.
+    Buzz\Client\BuzzClientInterface: '@buzz.client.file_get_contents'
+    Buzz\Client\BatchClientInterface: '@buzz.client.multi_curl'
+
+    # Register alias for PSR-18
+    Psr\Http\Client\ClientInterface: '@Buzz\Client\BuzzClientInterface'
+
+    # Clients
+    buzz.client.file_get_contents:
```
lyrixx: Why not using the fqcn ?

OskarStark: In this case Most of class: are obsolete too and would decrease the config file size too

Nyholm:
Thank you for the review

kriswallsmith/buzz/1.0/config/packages/buzz.yaml
lyrixx: @nyholm looks like you miss this comments (oh I merged this pr by approving I guess)

Nyholm: Yes, you did. 

I did not use FQCN on these because they should not really be "public". You should type hint to the interface not the concrete implementation. So the only time you should use this service id is when you write your alias.

PR #596, opened 2019-01-08
[WIP] Don't re-add the ENQUEUE_DSN from the bundle - just suggest

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

When using EnqueueBundle, I noticed a DX problem: the EnqueueBundle recipe adds the configuration that uses the `ENQUEUE_DSN` environment variable *and* it adds the `ENQUEUE_DSN` variable to `.env`. Then, when you install an individual transport, it adds a *second* `ENQUEUE_DSN` to `.env`.

Instead of doing that, let's just suggest how to configure this value.


makasim:
> What do you think about this approach? 

I am fine with it.

> Also, in some recipes, like amqp-bunny, there are 2 versions - one with ENQUEUE_DSN= amqp+bunny:// and a newer one with just ENQUEUE_DSN= amqp+bunny:. Can you explain the difference? 

According to [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3) this is the right one `amqp+bunny:`, `//` is added only if `authority` section is defined

weaverryan:
> According to RFC 3986 this is the right one amqp+bunny:, // is added only if authority section is defined

I see it - thank you! 

I have a few other notes / questions:

A) I noticed that `amqp-bunny`, `amqp-ext` and `amqp-lib` all (in their documentation) say that they can be configured using the `amqp:` protocol. How does Enqueue know which to use? I've updated the recipes to always use the more explicit format (e.g. `amqp+bunny:`) - wanted to make sure that's the preferred path.

B) I merged some recipes that had multiple versions into just one recipe. I did this (e.g. `aqmp-bunny`) when the only difference was showing with versus without the authority part. The new post-install message shows with the authority in those cases, because those transports most commonly need an authority.

C) I added the `resolve:` prefix to ENQUEUE_DSN in the bundle recipe so we can use, for example, `%kernel.project_dir%` in the environment variable (useful for the `fs` transport)

D) I also added the `PUBSUB_EMULATOR_HOST` env var to `.env` for the `gps` transport, as it appears to be needed.

I manually tested all of these recipes - things look good. Would love a review and +1 @makasim if you can. I'm smoothing out the install process so I can show Enqueue in a tutorial :).

I believe the test failures are incorrect.

enqueue/amqp-bunny/0.7/manifest.json
```diff
@@ -1,5 +1 @@
-{
-    "env": {
-        "ENQUEUE_DSN": "amqp+bunny://"
```
makasim: Enqueue 0.7 version would not work without `//`. It was fixed since 0.8.

enqueue/amqp-bunny/0.7/post-install.txt
```diff
@@ -0,0 +1,9 @@
+<bg=yellow;fg=black>                             </>
+<bg=yellow;fg=black> Using amqp-bunny in Enqueue </>
+<bg=yellow;fg=black>                             </>
+
+  * Configure <comment>ENQUEUE_DSN=amqp+bunny://user:pass@example.com:10000/%2f</comment>
```
makasim: `amqp+bunny:` adds default options for host, port, user and password. The most commonly used. So I am not sure we should provide more details other than that. 

If you still want to make it more descriptive I'd suggest 

```
ENQUEUE_DSN=amqp+bunny://guest:guest@localhost:5672/%2f
```

makasim:
> I noticed that amqp-bunny, amqp-ext and amqp-lib all (in their documentation) say that they can be configured using the amqp: protocol. How does Enqueue know which to use? I've updated the recipes to always use the more explicit format (e.g. amqp+bunny:) - wanted to make sure that's the preferred path.

Enqueue connection factory would try libraries in [this order](https://github.com/php-enqueue/enqueue-dev/blob/8b114cb1da316dee879ed3beaa3c29ac3a6c6047/pkg/enqueue/Resources.php#L90-L104): bunny, ext, lib

> B) I merged some recipes that had multiple versions into just one recipe. I did this (e.g. aqmp-bunny) when the only difference was showing with versus without the authority part. The new post-install message shows with the authority in those cases, because those transports most commonly need an authority.

👍 

> C) I added the resolve: prefix to ENQUEUE_DSN in the bundle recipe so we can use, for example, %kernel.project_dir% in the environment variable (useful for the fs transport)

👍 

> D) I also added the PUBSUB_EMULATOR_HOST env var to .env for the gps transport, as it appears to be needed.

👍

enqueue/amqp-bunny/0.7/post-install.txt
weaverryan: Ah, ok cool. I've changed to use the more descriptive suggestion then. This also solves the problem with the 0.7 version requiring the `//` - the example now shows WITH `//`, so this recipe is now compat with all versions.

weaverryan:
Thanks for the answers / review. This should be all good now!

weaverryan:
@nyholm I think this can be merged - but maybe the CI failure prevents it? Is there something I can do to help that?

enqueue/enqueue-bundle/0.9/config/packages/enqueue.yaml
```diff
@@ -1,4 +1,4 @@
 enqueue:
     default:
-        transport: '%env(ENQUEUE_DSN)%'
+        transport: '%env(resolve:ENQUEUE_DSN)%'
```
OskarStark: Can you explain why to use the resolve here?

Nyholm: We use `resolve` to be able to look at the DNS in the EnqueueExtension to figure out what transport is used. 

(Im not 100% that this is the case but I assume that's why)

OskarStark: Thank you, I was asking because I already use it without the resolve and it works.
Using amqp+lib and sqs

makasim: This has nothing to do with EnqueueExtension. resolve is used to resolve any parameters found in the DSN, for example `amqp://%amqp_user%:%amqp_pass%@localhost:5672`

OskarStark: You are right 👌🏻

PR #607, opened 2019-01-30
[doctrine_mongodb] add a metadata cache in prod

Lctrs:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Hm. Im leaning towards a 👎  here. But Im too ignorant with doctrine and mongodb. 

Is this really the 90% use case here? 

Ping @alcaeus, can I get your opinion? Is this a good "standard" symfony configuration?

Lctrs:
In Doctrine, the metadata cache refers to the cache of the computed entities informations (such as columns types, etc.) from annotations, xml or yaml files.

In production mode, they don't change. So using the default in memory cache lead to recomputing the metadatas on every request, thus slowing down the application.

The ORM recipe already comes with well configured cache in prod. This PR brings the same thing to the ODM recipe.

PR #610, opened 2019-02-03
Add recipe for FOSJSRoutingBundle >2.3, to fix deprecation on Symfony >4.1

tobias-93:
See https://github.com/FriendsOfSymfony/FOSJsRoutingBundle/issues/327, we added a new routing.xml with the new syntax to keep BC. New installations should automatically use the new xml file which is done in this recipe.

| Q             | A
| ------------- | ---
| License       | MIT


lex0r:
@tobias-93  I have Symfony 3.4.31 (with Flex) and installed the bundle, but the new routing file uses Symfony 4.x syntax so I had to manually change to use routing.xml. Is that an issue or I just misunderstand the compatibility requirements?

PR #614, opened 2019-02-08
Add ConnectHolland/CookieConsentBundle

matthijsch:
| Q             | A
| ------------- | ---
| License       | MIT


connectholland/cookie-consent-bundle/0.4/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "bundles": {
+        "ConnectHolland\\CookieConsentBundle\\CHCookieConsentBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "composer-scripts": {
+        "assets:install --symlink --relative %PUBLIC_DIR%": "symfony-cmd"
```
OskarStark: Is this really needed?
What if the user does not want to use symlink?

PR #620, opened 2019-02-22
Added recipes for KunstmaanCMS

Devolicious:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/kunstmaan/dashboard-bundle/5.2/config/packages/kunstmaan_dashboard.yaml
+++ b/kunstmaan/dashboard-bundle/5.2/config/packages/kunstmaan_dashboard.yaml
@@ -1,0 +1,9 @@
+# Configurate the dashboard bundle google api parameters to enable the google widget
+# More info at https://kunstmaanbundlescms.readthedocs.io/en/latest/cookbook/google-analytics-dashboard/
+#kunstmaan_dashboard:
+#    google_analytics:
+#        api:
+#            client_id: ''
+#            client_secret: ''
+#            dev_key: ''
+#            app_name: ''
--- a/kunstmaan/dashboard-bundle/5.2/config/packages/kunstmaan_dashboard.yaml
+++ b/kunstmaan/dashboard-bundle/5.2/config/packages/kunstmaan_dashboard.yaml
@@ -1,9 +1,9 @@
 # Configurate the dashboard bundle google api parameters to enable the google widget
 # More info at https://kunstmaanbundlescms.readthedocs.io/en/latest/cookbook/google-analytics-dashboard/
-#kunstmaan_dashboard:
-#    google_analytics:
-#        api:
-#            client_id: ''
-#            client_secret: ''
-#            dev_key: ''
-#            app_name: ''
+kunstmaan_dashboard:
+    google_analytics:
+        api:
+            client_id: '%env(GOOGLE_API_CLIENT_ID)%'
+            client_secret: '%env(GOOGLE_API_CLIENT_SECRET)%'
+            dev_key: '%env(GOOGLE_API_DEV_KEY)%'
+            app_name: '%env(GOOGLE_API_APP_NAME)%'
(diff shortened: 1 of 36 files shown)


acrobat:
@Nyholm CI and bot are happy, can you have a look and merge if you approve these recipes? Thanks!

kunstmaan/dashboard-bundle/5.2/config/packages/kunstmaan_dashboard.yaml
```diff
@@ -0,0 +1,9 @@
+# Configurate the dashboard bundle google api parameters to enable the google widget
+# More info at https://kunstmaanbundlescms.readthedocs.io/en/latest/cookbook/google-analytics-dashboard/
+#kunstmaan_dashboard:
+#    google_analytics:
+#        api:
+#            client_id: ''
+#            client_secret: ''
```
Nyholm: You should use environment variables or remove this file

kunstmaan/menu-bundle/5.2/config/packages/kunstmaan_menu.yaml
```diff
@@ -0,0 +1,6 @@
+#kunstmaan_menu:
+#    menus:
+#        - main_navigation
```
Nyholm: I think you should remove or uncomment this

kunstmaan/redirect-bundle/5.2/config/routes/kunstmaan_redirect.yaml
```diff
@@ -0,0 +1,6 @@
+KunstmaanRedirectBundle:
+    resource: "@KunstmaanRedirectBundle/Resources/config/routing.yml"
+    # If your site is not multi language, remove the "{_locale}/" part of prefix and requirements
+    prefix:   /{_locale}/
+    requirements:
+        _locale: "%requiredlocales%"
```
Nyholm: Where is this parameter defined?

acrobat: This is a parameter defined in the admin-bundle, which is a dependency of all bundles using this parameter. See [KunstmaanAdminBundle::KunstmaanAdminExtension](https://github.com/Kunstmaan/KunstmaanBundlesCMS/blob/422c988b093f1d9016a669864cc8a2408ced5146/src/Kunstmaan/AdminBundle/DependencyInjection/KunstmaanAdminExtension.php#L208-L209)

kunstmaan/menu-bundle/5.2/config/packages/kunstmaan_menu.yaml
acrobat: This is more to give the user a example config which they can uncomment and change to their needs. The names will show up in the interface, so the menu name will likely be changed by the user. We added this like the liip/imagine-bundle adds an [example config](https://github.com/symfony/recipes-contrib/blob/bce79fc5f07e850c0d53d13fcbd5978d740a6a6f/liip/imagine-bundle/1.8/config/packages/liip_imagine.yaml)

kunstmaan/dashboard-bundle/5.2/config/packages/kunstmaan_dashboard.yaml
acrobat: @Devolicious Do we remove this files and extend our documention so it's clear what the user should do, or do we add env vars for the google analytics config? I'm more in favor of removing it

Devolicious: @acrobat I would add the env vars. We use it all the time so it's more convenient if we do it that way. I know that those vars are 99% of the time not really bound to a specific environment but then the option will be there and if the user wants to replace with fixed vars that is also possible.

Devolicious:
@Nyholm can you please review again? We have made changes and are now using env vars as you suggested and the `requiredlocales` parameter has been changed to the `kunstmaan_admin.required_locales` parameter that you need to set in the config of that bundle. We left the other remark as is because it is example config that should be defined by the user and not by us like @acrobat stated.

PR #622, opened 2019-02-25
Added sonata-project/timeline-bundle

core23:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Could you address why Travis is red?

core23:
We rely on the sonata / fos user bundle, which does not have a recipe. So this is not possible for now :(

PR #634, opened 2019-03-12
commercetools bundle v0.4

nikossvnk:
| Q             | A
| ------------- | ---
| License       | MIT

recipe for http://packagist.org/packages/commercetools/symfony-bundle


commercetools/symfony-bundle/0.4/config/packages/security.yaml
```diff
@@ -0,0 +1,22 @@
+security:
+    providers:
+        ctp:
+            id: Commercetools\Symfony\CustomerBundle\Security\User\UserProvider
+    access_control:
+        - { path: /user/, roles: ROLE_USER }
+    encoders:
+        Symfony\Component\Security\Core\User\User: plaintext
+        Commercetools\Symfony\CustomerBundle\Security\User\User: plaintext
+    firewalls:
```
Nyholm: Do we really need these security rules?

Nyholm: Sorry for being short. 
I meant: The user will probably have a firewall setup already. But Im happy with you adding one new encoder.

commercetools/symfony-bundle/0.4/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+    "bundles": {
+        "Commercetools\\Symfony\\CtpBundle\\CtpBundle": ["all"],
+        "Commercetools\\Symfony\\CustomerBundle\\CustomerBundle": ["all"],
+        "Commercetools\\Symfony\\SetupBundle\\SetupBundle": ["dev"],
+        "Commercetools\\Symfony\\StateBundle\\StateBundle": ["dev"]
```
Nyholm: Are all these bundle living in this package?

commercetools/symfony-bundle/0.4/config/packages/security.yaml
nikossvnk: Well to be honest, I wasn't sure what the behavior of flex would be in case there is already a security.yaml file. ie if it was going to overwrite the file (?). So, I completely removed security.yaml now

commercetools/symfony-bundle/0.4/manifest.json
nikossvnk: yes

nikossvnk:
@Nyholm could you please re-review this PR after addressing your comments/suggestions?

PR #640, opened 2019-03-27
* Add a recipe for coka/pagination-bundle 4.0

CedrickOka:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/coka/api-bundle/4.0/config/packages/oka_api.yaml
+++ b/coka/api-bundle/4.0/config/packages/oka_api.yaml
@@ -1,0 +1,14 @@
+oka_api:
+    cors:
+        default:
+            allow_credentials: true
+            expose_headers: ['Accept-Encoding']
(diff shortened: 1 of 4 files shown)


CedrickOka:
Hi @Nyholm, 
I understood that it needed a different folder by major version of a bundle in a recipe; and in this version the default configuration was changed. Exemple the configuration value `oka_api.security_behaviors.password_updater` is equal to `false` instead of `true` in the previous versions.

CedrickOka:
Hi @Nyholm, 
I understood that it needed a different folder by major version of a bundle in a recipe; and in this version the default configuration was changed. Exemple the configuration value `oka_api.security_behaviors.password_updater` is equal to `false` instead of `true` in the previous versions.

Nyholm:
Pro tio: 

PR #644, opened 2019-03-29
Crayner Authenticate Bundle Recipe

crayner:
| Q             | A
| ------------- | ---
| License       | MIT

## Features
* Highest Available Encoder
* Manage Lock on Login Failures
* Password Validation
* Rotate Passwords


crayner/authenticate-bundle/1.0/config/packages/crayner_authentication.yaml
```diff
@@ -0,0 +1,46 @@
+crayner_authenticate:
+    user_class: Crayner\Authenticate\Entity\User
+    mailer_available: true
+    highest_available_encoder:
+        # Argon2i Options
+        memory_cost: 16384
```
Nyholm: It seams like you added all the default config in the recipe. You should remove this and make a reference to your docs. 

You should only keep the things that are most likely to be changed by the user. Like the `rotate_password.change_every`.

crayner: All removed.  I do see all of the settings like 'rotate_password.change_every', but understand why you would not see them that way.  Documentation already refers to all of these settings.  Thanks.

Nyholm:
Could you check why Travis is upset?

crayner:
Now passes on Symfony 4+, fails on Symfony <4, as per specification.

maxhelias:
@crayner you can define your `Configuration.php ` compatible like this : 
https://github.com/doctrine/DoctrineBundle/blob/f5c10ecbc2e199c8deeff3aedfe3ee778d44c50a/DependencyInjection/Configuration.php#L41

And travis will be happy :)

PR #649, opened 2019-04-03
Add gheb/configuration-converter-bundle

GregoireHebert:
| Q             | A
| ------------- | ---
| License       | MIT

Hello good fellows, this PR is part of a bundle that will allow you to switch from annotation configuration to xml or yaml format. At the moment dedicated to api-platform, but will be enlarged to serialization groups and doctrine mapping in the futur.


GregoireHebert:
@

PR #652, opened 2019-04-06
Remove DoctrineCacheBundle from recipes-contrib

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

This is installed automatically with the orm-pack. We never need to use this configuration though. Having this recipe here means that we will get the question "Do you want to execute this recipe?" every time we install the orm-pack. Which is clearly not what we want. 

As a symfony developer we configure the symfony/cache instead. 

#eu-fossa


Nyholm:
I reopened the PR and updated the title and description. Please re-review

@alcaeus

PR #653, opened 2019-04-06
[LiipImagine] Remove comments not needed

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

Files like this does not bring any value. 

#eu-fossa


Nyholm:
Thank you

PR #654, opened 2019-04-09
Add flysystem-bundle recipe

tgalopin:
| Q             | A
| ------------- | ---
| License       | MIT

Package: https://github.com/thephpleague/flysystem-bundle

### DIFF
--- a/league/flysystem-bundle/1.0/config/packages/flysystem.yaml
+++ b/league/flysystem-bundle/1.0/config/packages/flysystem.yaml
@@ -1,0 +1,7 @@
+# Read the documentation at https://github.com/thephpleague/flysystem-bundle/blob/master/docs/1-getting-started.md
+flysystem:
+    storages:
+        default.storage:
+            adapter: 'local'
+            options:
+                directory: '%kernel.project_dir%/%VAR_DIR%/storage/default'
--- a/league/flysystem-bundle/1.0/manifest.json
+++ b/league/flysystem-bundle/1.0/manifest.json
@@ -1,0 +1,9 @@
+{
+    "bundles": {
+        "League\\FlysystemBundle\\FlysystemBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "var/storage/": "%VAR_DIR%/storage/"
+    }
+}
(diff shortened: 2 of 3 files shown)


league/flysystem-bundle/1.0/config/packages/flysystem.yaml
```diff
@@ -0,0 +1,6 @@
+flysystem:
+    storages:
+        default.storage:
+            adapter: 'local'
+            options:
+                directory: '%kernel.project_dir%/storage'
```
Nyholm: If this is `'%kernel.project_dir%/var/storage'` there is a good change that folder is writable already.

league/flysystem-bundle/1.0/storage/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
Nyholm: No content needed if moved to `var/storage`

league/flysystem-bundle/1.0/config/packages/flysystem.yaml
tgalopin: Isn't there a habit of considering the var directory temporary and removable? I like the idea, but I'm not sure what people expect from var.

tgalopin:
I moved the directory but kept the gitignore content just in case.

tgalopin:
The bundle does not support Symfony 3.4, I'm not sure what to do about Travis :/ .

tgalopin:
I fixed the issue by introducing a env variable to disabled 3.4 check if needed.

@nicolas-grekas would be nice to have your review :) !

nicolas-grekas:
LGTM, @Nyholm OK to you?

league/flysystem-bundle/1.0/var/storage/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
Nyholm: Wouldn't this directory be created automatically since it is in the "var" folder?

league/flysystem-bundle/1.0/config/packages/flysystem.yaml
```diff
@@ -0,0 +1,6 @@
+flysystem:
```
Nyholm: If you want to, add a comment with a link to the docs

.travis.yml
```diff
@@ -14,6 +14,7 @@ matrix:
           env: SKELETON_VERSION="^4.0"
 
 before_install:
+    - export PACKAGES_ONLY_SF4="league/flysystem-bundle"
```
Nyholm: We should revert the changes in .travis.yml

It is okey that the build partially fails.

league/flysystem-bundle/1.0/var/storage/.gitignore
tgalopin: Not sure what you mean, I don't think it will.

PR #655, opened 2019-04-11
Remove recipe for FOSUserBundle

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

It was merge prematurely. Reverting #270

PR #270, opened 2018-02-01
Add recipe for FOSUserBundle

tattali:
| Q             | A
| ------------- | ---
| License       | MIT

PR #656, opened 2019-04-11
WIP: Reintroduce FOSUserBundle

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

Initially added in #270 and reverted in #655

### DIFF
--- a/friendsofsymfony/user-bundle/2.1/config/packages/fos_user.yaml
+++ b/friendsofsymfony/user-bundle/2.1/config/packages/fos_user.yaml
@@ -1,0 +1,10 @@
+fos_user:
+    db_driver: no_driver # valid values are 'orm', 'mongodb' and 'couchdb'
+    user_class: App\Entity\User
+    firewall_name: main
+    service:
+        mailer: fos_user.mailer.noop
+    from_email:
+        address: "%env(MAILER_SENDER_ADDRESS)%"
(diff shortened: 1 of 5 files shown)


maxhelias:
Blocked by FriendsOfSymfony/FOSUserBundle#2708

friendsofsymfony/user-bundle/2.1/config/packages/fos_user.yaml
```diff
@@ -0,0 +1,10 @@
+fos_user:
+    db_driver: no_driver # valid values are 'orm', 'mongodb' and 'couchdb'
```
Nyholm: This is not yet supported.

Nyholm:
We are waiting for https://github.com/FriendsOfSymfony/FOSUserBundle/pull/2708

juanmiguelbesada:
I think we have a problem here.

One side, we don't want to pre-config the db_driver to orm.

But at the other side, we are includin a User entity that uses ORM annotations, that requires doctrine/orm.

I think we have two options right now.

One, set default db_driver to orm, and configure a the recipe according.

The other, create a new meta-package, fosuserbundle-orm, that requires fosuserbundle and doctrine-bundle, and create a new recipe for this new meta-package that can be configured to use orm as db_driver.

What do you think?

PR #664, opened 2019-05-10
Support mukadi/wordpress-bundle v1.2.0

mbo2olivier:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Thank you. I would be happy to approve this. 

You **could** use the "copy-from-package" feature and move some of your code to your package instead of having it here in the recipe. That would make it simpler for you to version your code and to push new updates. 

Tell me what you want to do.

mbo2olivier:
> 
> 
> Thank you. I would be happy to approve this.
> 
> You **could** use the "copy-from-package" feature and move some of your code to your package instead of having it here in the recipe. That would make it simpler for you to version your code and to push new updates.
> 
> Tell me what you want to do.

Okay, it's not a bad idea. I can move all files into the package except the config ones ? what do you think about that ?

Nyholm:
That sounds great!

Nyholm:
Friendly ping

PR #675, opened 2019-05-31
Allow to define the default locale as environment variable

meritoo:
| Q             | A
| ------------- | ---
| License       | MIT

As mentioned in title, new environment variable that stores default locale

### DIFF
--- a/meritoo/common-bundle/0.1/manifest.json
+++ b/meritoo/common-bundle/0.1/manifest.json
@@ -6,6 +6,10 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
+        "MERITOO_DEFAULT_LOCALE": "en",
         "MERITOO_FORM_NOVALIDATE": "false"
+    },
+    "container": {
+        "locale": "%env(MERITOO_DEFAULT_LOCALE)%"
     }
 }


meritoo/common-bundle/0.1/manifest.json
```diff
@@ -6,6 +6,10 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
+        "MERITOO_DEFAULT_LOCALE": "en",
         "MERITOO_FORM_NOVALIDATE": "false"
+    },
+    "container": {
+        "locale": "%env(MERITOO_DEFAULT_LOCALE)%"
```
Nyholm: This recipe does never use the `%locale%` parameter. Why do you suggest adding this line?

meritoo: It's required to give a possibility to set value of `locale`, container's parameter, with environment variable `MERITOO_DEFAULT_LOCALE` defined in `.env`.

The `MERITOO_DEFAULT_LOCALE` variable has `en` as default value. So after installation of `meritoo/common-bundle` package and its recipe you have nothing to do.

In next step you can use `%locale%` parameter in your project, e.g.:

```yaml
framework:
    default_locale: '%locale%'
```

Nyholm: I undertand. My question is: Why do you need to set the `locale` parameter?

meritoo: To not using hardcoded string as value of the default locale:

```yaml
# config/packages/translation.yaml

framework:
    default_locale: en
```

and set value with environment variable:

```yaml
# config/services.yaml

parameters:
    locale: '%env(MERITOO_DEFAULT_LOCALE)%'
```

```yaml
# config/packages/translation.yaml

framework:
    default_locale: '%locale%'
```

Nyholm: Yes. 

I might be expressing myself incorrectly. I'll try again =)

I dont see anywhere in the recipe that you are using a parameter called "locale". Why do you need that parameter? 

I do understand why the framework bundle needs it (as you show in examples). That should go to the recipe of the framework bundle (it already is).

meritoo: I Wass looking for "locale" in `symfony/framework-bundle`'s recipe, but there is no such parameter. Take a look, please:

- https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/manifest.json#L14
- https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/config/packages/framework.yaml#L2

Could you tell where it is?

Nyholm: I was wrong. There is no parameter. It is defined here: https://github.com/symfony/recipes/blob/master/symfony/translation/3.3/config/packages/translation.yaml

Please answer my question: Why do **you** need that parameter?

meritoo: Sure :) Main reason: to have `locale` defined as environment variable out of the box after installation of the `meritoo/common-bundle` package.

Nyholm: That is not a good enough reason Im afraid. You should define it only if you use it. 

If you really really really really want an environment variable like that, I would be okey if you renamed it `meritoo_locale`.

fabpot: But then, the question is: Why do you need the locale to be an env var? Using an env var means that you want to be able to change it *dynamically* without recompiling the cache. I do'nt see the use case and that won't work anyway as some parts of the cache do embed the local.

meritoo: Well, the `locale` as environmental variable allows developers working at project to define it as they need. Code of project will be unchanged, in this case.

Cache is not the main reason, actually. Placing all customizable parameters in one place, in the `.env` file, allows to make it easy to change.

fabpot: That's the wrong usage for `.env`. Define a parameter instead.

PR #677, opened 2019-06-11
Cover Service

LinkerReknil:
| Q             | A
| ------------- | ---
| License       | MIT


reknil/lp-cover/1.3/config/packages/lp_section.yaml
```diff
@@ -0,0 +1,3 @@
+twig:
+    globals:
+        myvar: 'XXX'
```
maxhelias: You can define a env var like this PR #230, if you want

Nyholm:
I dont undersand the purpose of this recipe. It looks like you just are adding a twig global variable. 

Could you elaborate?

PR #682, opened 2019-06-13
Add notfloran/mjml-bundle

notFloran:
| Q             | A
| ------------- | ---
| License       | MIT


notFloran:
If I do that, it will replace the packages.json or merge the content?

maxhelias:
If the file already exists in the project it will not do anything. However, if it does not exist, it will be copied

PR #692, opened 2019-06-28
Adding recipe for sroze/messenger-enqueue-transport

weaverryan:
| Q             | A
| ------------- | ---
| License       | MIT

Hi! The package was renamed from the enqueue organization - https://github.com/sroze/messenger-enqueue-transport/commit/222138523c9f30991bca0199f3479531fe8b40b9

I duplicated the old recipe, which should fix: https://github.com/sroze/messenger-enqueue-transport/issues/73

Cheers!


weaverryan:
Hmm... I see now that the library IS listed as a symfony-bundle... so this recipe shouldn't be needed (in theory?) to trigger the `bundles.php` update? Or, perhaps because it's a non-standard location it *is* needed?

PR #693, opened 2019-06-29
Adding Bref logger

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/brefphp/logger

FYI @mnapoli

### DIFF
--- a/bref/logger/1.0/config/packages/bref_logger.yaml
+++ b/bref/logger/1.0/config/packages/bref_logger.yaml
@@ -1,0 +1,6 @@
+services:
+    Bref\Logger\StderrLogger: ~
+
+    # Register aliases
+    logger: '@Bref\Logger\StderrLogger'
+    Psr\Log\LoggerInterface: '@Bref\Logger\StderrLogger'
--- a/bref/logger/1.0/config/packages/bref_logger.yaml
+++ b/bref/logger/1.0/config/packages/bref_logger.yaml
@@ -2,5 +2,5 @@ services:
     Bref\Logger\StderrLogger: ~
 
     # Register aliases
-    logger: '@Bref\Logger\StderrLogger'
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
There's already a very similar logger wired by default in Symfony. I fear this might be confusing.

Nyholm:
That is true.. and you would not get an error that two services are defining the same alias.

I will comment out the aliases. Thank you for the review

nicolas-grekas:
Sorry to be annoying, but I don't see the point of having a recipe at all :)

Nyholm:
The point is to automatically register the service and to make sure you can use it with zero config in your application. 

With this recipe I can write the class: 

```php
use Bref\Logger\StderrLogger;
class Foo {
  public function __construct(StderrLogger $logger) {
     // ...
  }
}
```

I cannot do this without this recipe.

nicolas-grekas:
Sure, but the best practice is to type hint an interface, and there is one here. Hacking autowiring to save writing configuration is something we should discourage IMHO.

Nyholm:
Oh, I see what you are saying now.. Or what you mean: 

In the HttpKernel component there is a [Logger](https://github.com/symfony/symfony/blob/v4.3.2/src/Symfony/Component/HttpKernel/Log/Logger.php) which does pretty much exact the same thing as `bref/logger`. That class is automatically registered as service `logger` with the alias to `Psr\Log\LoggerInterface`. So the `bref/logger` package is never needed in a Symfony application. 

One could of course make the argument that this recipe would be good for people not using the HttpKernel component. But I think that is an edge case. 

Thank you.

mnapoli:
Here is the current documentation for using Symfony with Bref:

Capture d’écran 2019-06-29 à 13 55 53

(see https://bref.sh/docs/frameworks/symfony.html)

As such I don't think bref/logger is necessary in Symfony.

PR #694, opened 2019-07-01
Add recipe for netgen/content-browser

emodric:
| Q             | A
| ------------- | ---
| License       | MIT


emodric:
Hi. The failure is caused by latest version of `ocramius/proxy-manager` in combination with `ignore-platform-reqs` option used by Travis. How to handle this best, since in normal operation (when not ignoring platform requirements) this should not happen?

Thanks!

emodric:
Thank you @nicolas-grekas !

PR #696, opened 2019-07-10
[Redirection.io] Add project_key to the configuration

lyrixx:
| Q             | A
| ------------- | ---
| License       | MIT

Fixes https://github.com/redirectionio/proxy-symfony/issues/2

### DIFF
--- a/redirectionio/proxy-symfony/0.2/config/packages/redirectionio.yaml
+++ b/redirectionio/proxy-symfony/0.2/config/packages/redirectionio.yaml
@@ -1,0 +1,6 @@
+# Documentation: https://github.com/redirectionio/proxy-symfony
+redirection_io:
+    project_key: FIXME
+    connections:
+        my_agent: tcp://127.0.0.1:10301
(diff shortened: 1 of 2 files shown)


redirectionio/proxy-symfony/0.1/config/packages/redirectionio.yaml
```diff
@@ -1,5 +1,6 @@
 # Documentation: https://github.com/redirectionio/proxy-symfony
 redirection_io:
+    project_key: FIXME
```
fabpot: Could it be an env var? That would be more consistent with how we approach the same situation for other projects.

lyrixx: I don't know, but we are using the same project_key for a project. This value will be static (it's the same in dev / preprod / prod). The end user configure each instances via the website.

So adding a new env var will add an unnecessary complexity to me.

WDYT?

lyrixx:
I updated the PR: Now the recipies target v0.2.0

PR #697, opened 2019-07-11
Tweak default cache conf to rely on Symfony Cache and only in prod

Lctrs:
| Q             | A
| ------------- | ---
| License       | MIT

Also add cache for metadatas


Lctrs:
cc @ElectricMaxxx @dbu

PR #721, opened 2019-08-03
Added rule to rewrite url with encoded anchor

LaurentMarquet:
Don't know if it should be included in the recipe but I was fighting to have a rule to rewrite https://example.com%23anchor to https://example.com#anchor so I think it could be

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/apache-pack/1.0/public/.htaccess
+++ b/symfony/apache-pack/1.0/public/.htaccess
@@ -53,6 +53,9 @@ DirectoryIndex index.php
     RewriteCond %{REQUEST_FILENAME} -f
     RewriteRule ^ - [L]
 
+    # Rewrites link with anchor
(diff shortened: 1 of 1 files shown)


Nyholm:
Thank you. 

I can understand that you chased for that config for a time. 
I dont think it belongs in a recipe since it is not related to symfony. But I think you should write a blog post about it to tell the world.

LaurentMarquet:
Ok no problem. For the blog post as, it's written here, google may find it ;)
So, for information, here are the three rules I've added to my own htacess, as "=" and "?" are sometimes coded too
```
    RewriteRule ^(.*)#(.*)$ %{REQUEST_URI} [R=301,L,NE]
    RewriteRule ^(.*)\?(.*)$ %{REQUEST_URI} [R=301,L,NE]
    RewriteRule ^(.*)=(.*)$ %{REQUEST_URI} [R=301,L,NE]
```

PR #728, opened 2019-08-12
Add oauth2-bundle contrib recipe

Allypost:
| Q             | A
| ------------- | ---
| License       | MIT


Allypost:
The build fails because of the Symfony security-bundle package or rather because of a problem with the way interfaces are loaded in PHP 7.4 (https://github.com/symfony/symfony/issues/32395 and https://github.com/symfony/symfony/issues/32995).

@nicolas-grekas @nikic
We're not sure what to do with this error since the CI uses a 7.4snapshot image.
Any help is appreciated.

Nyholm:
Thank you @Allypost for this PR and for your work. 

I've tested this PR locally and it looks great: 

Screenshot 2019-08-13 at 19 18 45

PR #731, opened 2019-08-15
Add a Domain Event Bundle

peter-gribanov:
| Q             | A
| ------------- | ---
| License       | MIT

Bundle to create the domain layer of your [Domain-driven design (DDD)](https://en.wikipedia.org/wiki/Domain-driven_design) application.
This [Symfony](https://symfony.com/) bundle is a wrapper for [gpslab/domain-event](https://github.com/gpslab/domain-event), look it for more details.


peter-gribanov:
> If i understand correctly, the bundle will behave the same if we put gpslab_domain_event ~.

Almost so. The difference in the parameter `publish_on_flush`.
I honestly don't really understand the idea configurations in ​​flex. If the bundle can work with default configurations, should i create a config? Or maybe i should define only the parameters recommended for change?

Nyholm:
> If the bundle can work with default configurations, should i create a config?

No, if so, there is no need to create a recipe. Flex can install your package without a recipe. 

Like Doctrine bundle, you cannot use it without a DSN to a database, that is why it needs a recipe. 

> Or maybe i should define only the parameters recommended for change?

If you have a highly opinionated config value (like `cache_ttl` or `default_langauge` etc) you may include it in the recipe. 

It is more about being "simple to install and configure" but not "show me all your options".

peter-gribanov:
@Nyholm i change config. Please review.

Nyholm:
Thank you. Im happy to merge. 

If you want to, you can include a link in to your readme/documentation in the yaml file

peter-gribanov:
@Nyholm thanks for idea :+1:

PR #739, opened 2019-08-27
Add recipe for mrjoops/symfony-service

mrjoops:
| Q             | A
| ------------- | ---
| License       | MIT


mrjoops:
Travis fail with the following error:

>   [LogicException]                         
>
>  File "/.phpcs.xml.dist" does not exist! 

I don't know why it fails since the file effectively exists at the root of [mrjoops/symfony-service](https://github.com/mrjoops/symfony-service) !

mrjoops:
Ok, that was because of the metapackage type in the composer config, fixed!

mrjoops:
Actually there are multiple version folders because the target package version was incrementing then, but I agree with you, do you want me to remove the older folders?

The package is used to pre-configure multiple microservice projects and have gently defaults, as well as some more opinionated. It is supposed to be installed at the beginning of the project lifetime.
Another way to achieve the same goal is to create a project template, however I tend to prefer the recipe style which is more light on the dependencies.

Nyholm:
> do you want me to remove the older folders?

Yes please. 

> It is supposed to be installed at the beginning of the project lifetime

That is something we cannot guarantee. 

Im afraid that the recipes are not build to be used as a replacement of a project template or skeleton.

Nyholm:
Im closing this due inactivity.

PR #740, opened 2019-08-30
Adding alias to recipe.

ktherage:
| Q             | A
| ------------- | ---
| License       | MIT

Adding alias to this recipe.


fabpot:
Recipes in the contrib repo cannot have aliases.

PR #743, opened 2019-09-03
Add becklyn/assets-bundle

apfelbox:
| Q             | A
| ------------- | ---
| License       | MIT

Adding the recipe for https://packagist.org/packages/becklyn/assets-bundle

### DIFF
--- a/becklyn/assets-bundle/2.4/config/routes/becklyn_assets.yaml
+++ b/becklyn/assets-bundle/2.4/config/routes/becklyn_assets.yaml
@@ -1,0 +1,3 @@
+_import.becklyn_assets:
+    resource: .
+    type: 'becklyn-assets'
--- a/becklyn/assets-bundle/2.4/manifest.json
+++ b/becklyn/assets-bundle/2.4/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Becklyn\\AssetsBundle\\BecklynAssetsBundle": ["all"]
+    },
+    "copy-from-recipe": {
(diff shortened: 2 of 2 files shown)


apfelbox:
This one should be ready. It is supposed to fail on ^3.4, as we only support 4.2+

apfelbox:
Sorry for bothering you, but is there something missing here? @fabpot @Nyholm

becklyn/assets-bundle/2.4/config/routes/becklyn_assets.yaml
```diff
@@ -0,0 +1,3 @@
+_import.becklyn_assets:
+    resource: .
+    type: 'becklyn-assets'
```
Nyholm: I dont quite understand this. Would you elaborate?

apfelbox: Sure!

That is the matching configuration for our route loader: https://github.com/Becklyn/AssetsBundle/blob/master/src/RouteLoader/AssetsRouteLoader.php

The path of where the assets in your app should be at is configurable in the bundle.

This is required as the assets bundle can either provide the assets directly via a controller (there we need the path to configure a route dynamically – which is the link above and why this custom route loader is required) or we can statically dump the assets and serve them directly via the web server (in prod – here we need the path to know where to write the files to).

tl;dr as we have a configurable part in a route, we create the route dynamically.

Nyholm: Cool. Thank you

PR #754, opened 2019-09-15
Dockfony/dockfony

MatthewBooth:
| Q             | A
| ------------- | ---
| License       | MIT

Added some new Environment Variables that are required for some recently added functionality, and touched up the post-install message.

### DIFF
--- a/dockfony/dockfony/5.0/manifest.json
+++ b/dockfony/dockfony/5.0/manifest.json
@@ -7,7 +7,9 @@
     "env": {
         "DB_DATABASE": "symfony",
         "DB_USERNAME": "symfony",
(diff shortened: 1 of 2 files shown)


Nyholm:
Where do you use these variables?

MatthewBooth:
@Nyholm

Partly here: https://github.com/dockfony/dockfony/commit/2b3bc692bc04e42aa5241217d437514277c50677

And then you would also use the host inside your Symfony config.

I'm hoping these will be the last PR I need for this recipe for a while 😊

Nyholm:
Cool. I assumed they would be there =)

Thank you

PR #758, opened 2019-09-25
Added some aliases only for faster recipe install

racastellanosm:
| Q             | A
| ------------- | ---
| License       | MIT

Only adding some aliases for Elastica Bundle recipe, for faster recipe install without knowledge of compelte package path.

Proposed aliases: 
- "elastica" 
- "elasttica-bundle"


racastellanosm:
Contri repos, doesn't allow aliases. My bad!

PR #761, opened 2019-09-27
Update coka recipes and add new recipes

CedrickOka:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/coka/file-bundle/3.0/config/packages/oka_file.yaml
+++ b/coka/file-bundle/3.0/config/packages/oka_file.yaml
@@ -4,4 +4,3 @@ oka_file:
         root_path: '%env(OKAFILE_STORAGE_ROOT_PATH)%'
         webserver:
             host: '%env(OKAFILE_STORAGE_WEB_SERVER_HOST)%'
-
--- a/coka/file-bundle/3.0/config/packages/oka_file.yaml
+++ b/coka/file-bundle/3.0/config/packages/oka_file.yaml
@@ -1,6 +1,6 @@
 oka_file:
     db_driver: orm
     storage:
-        root_path: '%env(OKAFILE_STORAGE_ROOT_PATH)%'
+        root_path: '%env(resolve:OKAFILE_STORAGE_ROOT_PATH)%'
         webserver:
-            host: '%env(OKAFILE_STORAGE_WEB_SERVER_HOST)%'
+            host: '%env(resolve:OKAFILE_STORAGE_WEB_SERVER_HOST)%'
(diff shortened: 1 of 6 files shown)


Nyholm:
thank you. Could you move the parameters to the manifest file?

CedrickOka:
Hi,

@Nyholm Can you help me by looking at my pull request?

coka/file-bundle/3.0/config/packages/oka_file.yaml
```diff
@@ -4,4 +4,3 @@ oka_file:
         root_path: '%env(OKAFILE_STORAGE_ROOT_PATH)%'
         webserver:
             host: '%env(OKAFILE_STORAGE_WEB_SERVER_HOST)%'
-
```
Nyholm: Why were these lines changed?

coka/rest-request-validator-bundle/1.0/config/packages/oka_rest_request_validator.yaml
```diff
@@ -0,0 +1,3 @@
+oka_rest_request_validator:
+    exception:
+        enabled: true
```
Nyholm: Is this the same as the default value?

coka/rate-limit-bundle/1.0/config/packages/oka_rate_limit.yaml
```diff
@@ -0,0 +1,3 @@
+oka_rate_limit:
+    configs:
+        - {path: ''}
```
Nyholm: I like this config, because me as user are probably going to change that. 

:+1:

coka/file-bundle/3.0/config/packages/oka_file.yaml
CedrickOka: Hi,
These lines have changed to allow you to use environment variables to configure where your files are stored and little like the doctrine URL database

coka/rest-request-validator-bundle/1.0/config/packages/oka_rest_request_validator.yaml
CedrickOka: Hi,
yes it is the same thing as the default variable but I put it more in mind to indicate what needs to be changed according to these needs

coka/rate-limit-bundle/1.0/config/packages/oka_rate_limit.yaml
CedrickOka: Hi,
Cool, thanks

coka/file-bundle/3.0/config/packages/oka_file.yaml
Nyholm: If I look at the diff. I see that you just removed an empty line in this file. 

Generally you should not modify other files or fix CS. I just asked if there is any particular reason for this change or if it just was “unrelated CS”.

coka/rest-request-validator-bundle/1.0/config/packages/oka_rest_request_validator.yaml
Nyholm: Hm... okey.

coka/file-bundle/3.0/config/packages/oka_file.yaml
CedrickOka: Well understood thank you

PR #763, opened 2019-10-01
Add biig/domain recipe

Nek-:
It is useful because by default some features are not enabled and it's
not always obvious they are not.

| Q             | A
| ------------- | ---
| License       | MIT


Nek-:
@Nyholm we had something wrong in the package (serializer required while it was not a dependency, it's fixed!).

Nyholm:
Thank you. The recipe looks good now. 

But it seams that you added the exact same version of the recipe for 1.0 and 2.0. Is that correct?

If so, You can remove 2.0 and flex would use the 1.0 recipe when installing 2.0. 

(But I dont thing any new user will install 1.0, right? So maybe deprecate that version and just focus on 2.0?)

Nek-:
@Nyholm v1 is compatible with Sf 3.3+, v2 Sf 4.3+ (because the event dispatcher changed).

The policy is to maintain both versions (for LTS 3.4).

I dropped duplication. Thanks :) .

PR #766, opened 2019-10-05
imper86/allegro-api-bundle and /supervisor-bundle recipes

imper86:
| Q             | A
| ------------- | ---
| License       | MIT


imper86:
added second recipe, for imper86/supervisor-bundle

PR #767, opened 2019-10-06
Add Timiki\RpcServerBundle and Timiki\RpcClientBundle

netesin:
| Q             | A
| ------------- | ---
| License       | MIT


timiki/rpc-server-bundle/4.0/config/packages/rpc_server.yaml
```diff
@@ -0,0 +1,33 @@
...
+    #
+    # Default mapping
+    #   mapping: '%kernel.root_dir%/Method'
+    #
+    # Multi dir mapping
+    #   mapping:
+    #   - '%kernel.root_dir%/Method1'
+    #   - '%kernel.root_dir%/Method2'
+    #
+    # Multi handler|dir mapping
```
Nyholm: This long comment is not needed, right? It should be in the bundles docs.

netesin: @Nyholm you right about that. Bundle doc contains info about it ) i fix it

PR #780, opened 2019-11-05
add networking/init-cms-bundle recipe

ychadwick:
| Q             | A
| ------------- | ---
| License       | MIT


ychadwick:
How is it looking with merging this pull request? It would be great to have the recipe up and running.

networking/init-cms-bundle/4.0/config/packages/dev/networking_init_cms.yaml
```diff
@@ -0,0 +1,4 @@
+networking_init_cms:
+    cache:
+        activate: false
+
```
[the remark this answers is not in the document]
ychadwick: tabs removed

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
```diff
@@ -0,0 +1,101 @@
...
+    - { resource: "@NetworkingInitCmsBundle/Resources/config/cms/sonata_media.yaml" }
+    - { resource: "@NetworkingInitCmsBundle/Resources/config/cms/sonata_notification.yaml" }
+    - { resource: "@NetworkingInitCmsBundle/Resources/config/cms/sonata_user.yaml" }
+    - { resource: "@NetworkingInitCmsBundle/Resources/config/cms/stof_doctrine_extensions.yaml" }
+    - { resource: "@NetworkingInitCmsBundle/Resources/config/cms/symfony_cmf_routing_extra.yaml" }
+# Default security configuration
+#    - { resource: "@NetworkingInitCmsBundle/Resources/config/cms/security.yaml" }
+
+# Set templating to twig. FOSUserBundle requires templating to be setup
+framework:
```
Nyholm: Please done change other bundle's configuration

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
```diff
@@ -0,0 +1,101 @@
...
+        directories:
+            NetworkingSonataMediaBundle:
+                namespace_prefix: "Sonata\\MediaBundle"
+                path: "@NetworkingInitCmsBundle/Resources/config/serializer"
+            app:
+                namespace_prefix: "App"
+                path: "%kernel.project_dir%/config/serializer"
+
+# Translation bundle configuration, swap environment variables if mange_locales_only option set to false
+lexik_translation:
```
Nyholm: Please remove

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+
+networking_init_cms:
+    # Set a cookie in the fronted end for language. If disabled, the URL will be prefixed with the locale
+    allow_locale_cookie: true
+
+    # Simple php based cache for caching the rendered HTML in the frontend
+    cache:
+        activate: true #default false
+        cache_time: 86400 #default value
```
Nyholm: A recipe is not a place where you dump all default values. 

Please remove

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
```diff
@@ -0,0 +1,101 @@
...
+        - { name: 'Text' , class: 'Networking\InitCmsBundle\Entity\Text'}
+        - { name: 'Gallery Viewer' , class: 'Networking\InitCmsBundle\Entity\GalleryView'}
+
+    # Classes to be used for the page and user entities
+    class:
+        page: "App\\Entity\\Page"
+        user: "App\\Entity\\User"
+
+    # Default template settings to get you started
+    templates:
```
Nyholm: This is not needed in the recipe

ychadwick: This is needed, the Entity is copied into the users Entity folder, which if not done, means there is no user entity to setup the CMS.

This is a CMS, and requires Page and User entities in order for the setup to run

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
ychadwick: This is not a dump of default values, these are the initial values set, which the user can change. It is nice to have it visible, rather than only being in the configuration.php file.

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
ychadwick: This configuration is needed in order fo the CMS to work. The idea is for us to add the configuration so that the user does not need to configure all the individual bundles themselves. 

It is vital that it is done this way.

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
ychadwick: The same thing here applies, if there are no default locale set the pages can not be created when running the install command for the cms. Unfortunately many of the bundles that we use do not yet have a recipe. We want to able for a user to get the CMS up and running with as little configuration and file copying as possible.

Also, our company is one of the only group of people who use this CMS. It is developed for us to quickly set up a CMS for our customers.

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
Nyholm: Sorry, When I said "default values" I meant "initial values".

ychadwick:
Basically if we did not incorporate the configs for the other bundles, the website would not run. The CMS is made of many other bundles as well, which if not configured correctly will not run at all.

As you can see from the post-install.txt we point out that there are many configurations imported, and that the user should check, update and extend them where necessary.

After that, the user can run the install command, and the website will run.

If we were to remove the config files and configuration as suggested, then there would be nothing but fatal errors.

For example, the framework templating variable is needed as the fosUserBundle requires it, and they do not have this set in their recipe (there is no recipe). However the FosCkeditor bundle adds a twig configuration. According to your logic, they should not do that?

Also we need to define a cache pool for our application.

As you can see here below, we add the framework configuration so that the user doesn't need to figure out why FosUserBundle is not working.

```
# Set templating to twig. FOSUserBundle requires templating to be setup
framework:
    templating:      { engines: ['twig'] }
    cache:
        pools:
            page.cache:
                adapter: cache.app #Setup page cache using template from cache.app
```

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
Nyholm: Oh, Im misstaken. I read the comment as "Default template settings".. as this is default. I should have read it as "here is some bootstrap settings". 

This is fine.

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
Nyholm: There are some issues will modifying other bundles configuration. 

- But what if someone got this bundle already installed? Then you have two places where you defined `lexik_translation`'s default locale
-LexikTranslation bundle's recipe will also run. This will create multiple configurations for the same bundle.

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
ychadwick: As mentioned, we could leave the initial values away (they are anyway in the configuration.php file), but it is important that the users see it. We could leave all the configuation setting away for our particular bundle, there are initial values set for almost every service and parameter in our code. But, without a view of what and where to customise it, you are making the user experience, in my humble opinion, pretty crappy.

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
ychadwick: There is not LexikTranslationBundle recipe, that is exactly the point.

ychadwick: Also, as mentioned in our post-instal.txt we do point out that there may be cross over in the configuration, which is why we state that the user should check first before proceeding.

That is mentioned in the very first point.

```
    1. Check that all files imported via the networking_init_cms.yaml file are not overwritten with a similar
       configuration file in your /config/packages folder. !IMPORTANT the doctrine.yaml and security.yaml must be either
       deleted, or update/extended with the configuration as proposed in the doctrine.yaml and security.yaml files
       included with the CMS.
```

Nyholm:
I understand what you are saying and why you need to modify other bundle's config. 

I do accept recipes that ie adds cache pools to framework bundle or form themes to Twig bundle. 
I will not accept PRs that changes config (like the templating engine or Lexik's locale). 

Creating massive bundles like this are much harder to setup and create recipes for. That is why we never manage to do one for ForUserBundle as example.

ychadwick:
But the fosckeditor busked recipe adds a form theme, to name just one example. It seems a little inconsistent to me. 

We make all the configuration transparent for the user. In this way nothing is hidden in the bundle.

My tests have shown it to be easy a straightforward to set up, not harder. Harder would be copying and pasting configuration all over the place, the recipe eliminates this complexity whilst maintaining transparency.

ychadwick:
Of course we can just remove the configuration items, and inform the user that she/he either needs to import them from our bundle or configure them accordingly

Nyholm:
> But the fosckeditor busked recipe adds a form theme, to name just one example. It seems a little inconsistent to me.

Here are the rules of when I accept modifying other bundles config: 
- I do accept recipes that ie adds cache pools to framework bundle or form themes to Twig bundle.
- I will not accept PRs that changes config (like the templating engine or Lexik's locale).

> Of course we can just remove the configuration items, and inform the user that she/he either needs to import them from our bundle or configure them accordingly

I agree. That would be the best way to move forward if you want this PR to be merged.

ychadwick:
Thank you for the clarification, my mistake, I misread your email, I thought you said, that you do „not“ accept recipes that add cache pools and form themes.

Right, so as long as I add, and not change configuration, all is good.

I will submit a change tomorrow.

Thank you for your time and feedback.

ychadwick:
How can I prevent testing for symfony 5.0? This bundle is only ready for symfony 3.4 and 4.x versions

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+    #Simple php based cache for caching the rendered HTML in the frontend
+    cache:
+        activate: true #enable for product, false in dev config
+    #    cache_time: 86400 #default value
+    #    cache_service: 'Networking\InitCmsBundle\Cache\PageCache' #default value
+
+    #Show admin toolbar when in the frontend
+    #admin_toolbar:
+    #    toolbar: true #Default is activated
```
Nyholm: Could you remove these config in comments? 

Rather give a link to your readme where you explain and show all possible config options.

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+    #Language set up, label is used in the default frontend language switcher menu
+    #languages:
+    #    - {label: 'Deutsch', locale: 'de'}
+    #    - {label: 'English', locale: 'en'}
+
+    #Add content types here
+    #content_types:
+    #    - { name: 'Text' , class: 'Networking\InitCmsBundle\Entity\Text'}
+    #    - { name: 'Gallery Viewer' , class: 'Networking\InitCmsBundle\Entity\GalleryView'}
```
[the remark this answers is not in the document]
ychadwick: Ok, no problem. Done

networking/init-cms-bundle/4.0/config/packages/networking_init_cms.yaml
ychadwick: Ok, no problem. Done

PR #785, opened 2019-11-25
Add PaylineBundle

lolautruche:
| Q             | A
| ------------- | ---
| License       | MIT


lolautruche/payline-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+    "bundles": {
+        "Lolautruche\\PaylineBundle\\LolautruchePaylineBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "PAYLINE_MERCHANT_ID": "your_payline_merchant_id",
+        "PAYLINE_ACCESS_KEY": "your_payline_access_key",
```
nicolas-grekas: New practice here: I think env var for secrets should be commented, in order to allow managing secrets with the new `secrets:*` commands
`"#PAYLINE_ACCESS_KEY": "your_payline_access_key",`

lolautruche/payline-bundle/2.0/post-install.txt
```diff
@@ -0,0 +1,9 @@
+<bg=blue;fg=white>                                                 </>
+<bg=blue;fg=white> You just installed LolautruchePaylineBundle!    </>
+<bg=blue;fg=white> What's next?                                    </>
+<bg=blue;fg=white>                                                 </>
+
+  * <fg=blue>Configure</> Payline environment variables and settings in <comment>config/packages/payline.yaml</comment>.
```
nicolas-grekas: but the yaml file references env vars - so we should tell ppl to have a look at `.env` files instead isn't it?

lolautruche/payline-bundle/2.0/manifest.json
lolautruche: Won't this make it unavailable then? If so, that would generate an exception as these information are mandatory

lolautruche:
Review ping @Nyholm :-)

lolautruche/payline-bundle/2.0/manifest.json
nicolas-grekas: One will get an exception when accessing the service while the env var is not set - but a random string is not better, wdyt? At compile time, this should make no difference as no env vars should be read then.
Can you give it a try?

lolautruche:
🎉

PR #787, opened 2019-11-27
Add the fork of phone-number-bundle recipe

Nek-:
After lack of maintainance of the main bundle, we decided to fork the bundle, this recipe will help new users.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/odolbeau/phone-number-bundle/3.0/config/packages/misd_phone_number.yaml
+++ b/odolbeau/phone-number-bundle/3.0/config/packages/misd_phone_number.yaml
@@ -1,0 +1,6 @@
+# To persist libphonenumber\PhoneNumber objects, add the Misd\PhoneNumberBundle\Doctrine\DBAL\Types\PhoneNumberType mapping to your application's config.
+# This requires: doctrine/doctrine-bundle
+#doctrine:
+#    dbal:
+#        types:
+#            phone_number: Misd\PhoneNumberBundle\Doctrine\DBAL\Types\PhoneNumberType
--- a/odolbeau/phone-number-bundle/3.0/post-install.txt
+++ b/odolbeau/phone-number-bundle/3.0/post-install.txt
@@ -1,0 +1,9 @@
+<bg=green;fg=white></>
+<bg=green;fg=white> Uncomment the configuration provided for doctrine
+<bg=green;fg=white> and now can use the phone bundle as described here: </>
+<bg=green;fg=white> https://github.com/odolbeau/phone-number-bundle#usage </>
+<bg=green;fg=white> The Doctrine configuration has been already done.</>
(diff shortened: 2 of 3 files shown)


odolbeau/phone-number-bundle/3.0/config/packages/misd_phone_number.yaml
```diff
@@ -0,0 +1,6 @@
+# To persist libphonenumber\PhoneNumber objects, add the Misd\PhoneNumberBundle\Doctrine\DBAL\Types\PhoneNumberType mapping to your application's config.
+# This requires: doctrine/doctrine-bundle
+doctrine:
+    dbal:
+        types:
+            phone_number: Misd\PhoneNumberBundle\Doctrine\DBAL\Types\PhoneNumberType
```
Taluu: this should be commented imo, as doctrine is not always installed and is optionnal.

odolbeau/phone-number-bundle/3.0/post-install.txt
```diff
@@ -0,0 +1,8 @@
+<bg=green;fg=white></>
+<bg=green;fg=white> You now can use the phone bundle as described here: </>
+<bg=green;fg=white> https://github.com/odolbeau/phone-number-bundle#usage </>
+<bg=green;fg=white> The Doctrine configuration has been already done.</>
+<bg=green;fg=white> You now can use `phone_number` as an ORM-Type: </>
+
+  /** @ORM\Column(type="phone_number") */
+  private $phoneNumber;
```
Taluu: you shouldn't need this file

Nek-: It's printed in the console. Which is nice.

Taluu: Maybe but it's kinda useless most of the time, so it's really there if you really have something to say. which is not the case here.

Imagine if you have 10 recipes executed with each of them one message...

maxhelias: Not necessarily, with this RFC we should be able to replay the message symfony/flex#518

jderusse:
Any reason for not registering this recipes for tag `1.3` (or 2.x) to allow people using `libphonenumber-for-php < 8` ?

odolbeau/phone-number-bundle/3.0/post-install.txt
jderusse: agreed with @Taluu recipes are not documentation.
What if I don't use `phone-number-bundle` for ORM, but for Form, or twig ?

maxhelias:
@jderusse Good point, i think we can

PR #798, opened 2019-12-09
Add required parent constructor call.

webdevilopers:
As described in the symfony docs `you *must* call the parent constructor`:
https://symfony.com/doc/current/console/commands_as_services.html

Came from:
https://twitter.com/webdevilopers/status/1203757726314184705


fabpot:
It would be better to move this command to the bundle.

gquemener:
Closed in favor of #799 .

Thanks @webdevilopers

PR #799, opened 2019-12-09
[prooph/event-store-symfony-bundle] Fixes incorrect command declaration

gquemener:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/prooph/event-store-symfony-bundle/0.4/config/packages/prooph_event_store.yaml
+++ b/prooph/event-store-symfony-bundle/0.4/config/packages/prooph_event_store.yaml
@@ -8,3 +8,5 @@ services:
         public: false
 
     Prooph\EventSourcing\EventStoreIntegration\AggregateTranslator: null
+
+    Prooph\EventStore\EventStore: '@app.event_store.default'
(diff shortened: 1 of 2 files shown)


gquemener:
> It would be better to move this command to the bundle. (https://github.com/symfony/recipes-contrib/pull/798#issuecomment-563326402)

This command is not shipped within the bundle because it is not generic enough. Writing a generic command in the bundle would require a certain amount of time that no contributor has the ability to offer at present time. That is why the command is short, specific to the configuration shipped in the recipe and living by the side of the recipe, on this repository.

Does that make sense?

gquemener:
As this recipe won't work if the `prooph/pdo-event-store` package is not required as well, I am suggesting the (re)creation of a `prooph-pack` here: https://github.com/prooph/prooph-pack/pull/2

Nyholm:
Friendly ping

gquemener:
Hello,

I don't have the will, nor the knowledge, to continue this pull request. As a result, you can close it @Nyholm.

Thanks.

Nyholm:
I understand. 

Thank you for getting back to me.

PR #811, opened 2019-12-22
Add recipe for inwx-api-bundle >= v1.0

SebTM:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sebtm/inwx-api-bundle/1.0/config/packages/inwx_api.yaml
+++ b/sebtm/inwx-api-bundle/1.0/config/packages/inwx_api.yaml
@@ -1,0 +1,8 @@
+# Read the documentation: https://github.com/SebTM/inwx-api-bundle
+inwx_api:
+    debug: '%kernel.debug%'
+    environment: development
+    json: true
+    language: en
+    username: '%env(resolve:INWX_USERNAME)%'
+    password: '%env(resolve:INWX_PASSWORD)%'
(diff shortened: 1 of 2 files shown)


sebtm/inwx-api-bundle/1.0/config/packages/inwx_api.yaml
```diff
@@ -0,0 +1,8 @@
+# Read the documentation: https://github.com/SebTM/inwx-api-bundle
+inwx_api:
+    debug: false
+    environment: development
```
Nyholm: You should remove these two and maybe read the default from `%kernel.debug%`, `%kernel.environment%`, right?

SebTM: I understand the first point with removing the not needed defaults, but I think there are cases you want to enable debugging and/or switch to different environment regardless "%kernel.*%" settings. I would suggest to set them to the "%kernel.*" as default but keep them independent changeable.

Would you be fine with this?

SebTM: @Nyholm Ping, would you be fine with this as proposed?

Nyholm: Yes

PR #818, opened 2020-01-03
updates php-translation/symfony-bundle recipe for version 0.10

zghosts:
This PR updates the recipe for php-translation/symfony-bundle to work for version 0.10


zghosts:
Already done nwm

PR #829, opened 2020-01-23
[symfony/apache-pack] Better FollowSymlinks suggestion

fbastien:
`.htaccess` has a commented out `Options FollowSymlinks` suggestion. When uncommented, it overwrites options defined in any prior `Options` directive (for instance in the main Apache configuration). See [Apache doc](https://httpd.apache.org/docs/current/fr/mod/core.html#options).

I think it should rather be `Options +FollowSymlinks` so that it enables that option without disabling others.

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/symfony/apache-pack/1.0/public/.htaccess
+++ b/symfony/apache-pack/1.0/public/.htaccess
@@ -9,7 +9,7 @@ DirectoryIndex index.php
 # feature in your server configuration. Uncomment the following line if you
 # install assets as symlinks or if you experience problems related to symlinks
 # when compiling LESS/Sass/CoffeScript assets.
-# Options FollowSymlinks
+# Options +FollowSymlinks
 
 # Disabling MultiViews prevents unwanted negotiation, e.g. "/index" should not resolve
 # to the front controller "/index.php" but be rewritten to "/index.php/index".


Nyholm:
But there is no other prior options directive here, right? So there will be no difference if there is a plus or not.

fbastien:
@Nyholm Although there is no other prior `Options` directive in that particular `.htaccess` file, directives in this file still override global directives in `httpd.conf` (or similar) file and directives in `.htaccess` files or `<Directory>` sections of ancestor directories.

For instance, the Apache distro I'm using comes with an `Options Indexes FollowSymLinks Includes ExecCGI` directive for the `htdocs` directory (`var/www` equivalent). If `.htaccess` in my Symfony project subdirectory contains `Options FollowSymLinks`, then `FollowSymLinks` becomes the only one enabled option (thus disabling `Indexes`, `Includes` and `ExecCGI`). Whereas if `.htaccess` contains `Options +FollowSymLinks` (notice plus sign), it would enable `FollowSymLinks` if it wasn't previously, but other options stay in the same state (i.e. in that case, `Indexes`, `Includes` and `ExecCGI` stay enabled).

Nyholm:
@Roy-Orbison, could you help us here? I am not 100% about this PR.

Roy-Orbison:
@Nyholm Yes, @fbastien is correct. Omission of the `+` overwrites all other values previously set in `Options` directives. The `+` to add it to the list of effective options.

Hanmac:
FollowSymlinks is needed even when not installing assets as symlinks.

if FollowSymlinks is not enabled, RewriteEngine will make it crash into a 403

PR #834, opened 2020-02-04
Add tienvx/fos-oauth-server-bundle-propel

tienvx:
| Q             | A
| ------------- | ---
| License       | MIT

FOSOAuthServerBundle does not support Symfony 5.0, so does this recipe.


tienvx:
Closed because propel/propel-bundle does not support Symfony 3.4, 4.4 and 5.0

PR #837, opened 2020-02-06
Configure gpslab/geoip2:~2.0

peter-gribanov:
| Q             | A
| ------------- | ---
| License       | MIT

The [`gpslab/geoip2`](https://github.com/gpslab/geoip2) package has a [new major release](https://github.com/gpslab/geoip2/releases/tag/v2.0.0) and needs update the recipe configuration for this release.

Before

```yml
gpslab_geoip:
    cache: '%kernel.cache_dir%/GeoLite2-City.mmdb'
    url: 'https://geolite.maxmind.com/download/geoip/database/GeoLite2-City.mmdb.gz'
    locales: [ 'en' ]
```

After

```yml
gpslab_geoip:
    license: 'XXXXXXXXXXXXXXXX'
    edition: 'GeoLite2-City'
    path: '%kernel.cache_dir%/GeoLite2-City.mmdb'
    url: 'https://download.maxmind.com/app/geoip_download?edition_id=%s&license_key=%s&suffix=tar.gz'
    locales: [ 'en' ]
```

### DIFF
--- a/gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
+++ b/gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
@@ -2,7 +2,7 @@
 # https://github.com/gpslab/geoip2
 gpslab_geoip:
     # You should specify your personal licence key
-    license: '%env(GEOIP_LICENSE)%'
+    license: 'YOUR-LICENSE-KEY'
 
     # Edition ID of database you need
     edition: 'GeoLite2-City'
--- a/gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
+++ b/gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
@@ -1,8 +1,8 @@
 # Read the README file for more detailed configuration information
 # https://github.com/gpslab/geoip2
-#gpslab_geoip:
-#    # You should specify your personal licence key
-#    license: '%env(GEOIP_LICENSE)%'
-#
-#    # Edition ID of database you need
-#    edition: 'GeoLite2-City'
+gpslab_geoip:
+    # You should specify your personal licence key
+    license: '%env(GEOIP_LICENSE)%'
+
+    # Edition ID of database you need
+    edition: 'GeoLite2-City'
--- a/gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
+++ b/gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
@@ -1,0 +1,8 @@
+# Read the README file for more detailed configuration information
+# https://github.com/gpslab/geoip2
+#gpslab_geoip:
+#    # You should specify your personal licence key
+#    license: '%env(GEOIP_LICENSE)%'
+#
+#    # Edition ID of database you need
+#    edition: 'GeoLite2-City'
--- a/gpslab/geoip2/2.0/manifest.json
+++ b/gpslab/geoip2/2.0/manifest.json
@@ -9,6 +9,6 @@
         "geoip2:update": "symfony-cmd"
     },
     "env": {
-        "GEOIP_LICENSE": ""
+        "GEOIP_LICENSE": "YOUR-LICENSE-KEY"
     }
 }
--- a/gpslab/geoip2/2.0/manifest.json
+++ b/gpslab/geoip2/2.0/manifest.json
@@ -1,0 +1,14 @@
+{
+    "bundles": {
+        "GpsLab\\Bundle\\GeoIP2Bundle\\GpsLabGeoIP2Bundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "composer-scripts": {
+        "geoip2:update": "symfony-cmd"
+    },
+    "env": {
+        "GEOIP_LICENSE": ""
+    }
+}
--- a/gpslab/geoip2/2.0/manifest.json
+++ b/gpslab/geoip2/2.0/manifest.json
@@ -7,8 +7,5 @@
     },
     "composer-scripts": {
         "geoip2:update": "symfony-cmd"
-    },
-    "env": {
-        "GEOIP_LICENSE": "YOUR-LICENSE-KEY"
     }
 }


peter-gribanov:
Can you comment on this error?

> Failed to copy "https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-City&license_key=&suffix=tar.gz" to "/tmp/5e3c36ee974bc3.89307117_GeoLite2.tar.gz" because source file could not be opened for reading.

The command for automatically updating the database is added to the `composer-scripts`. This command is run when installing and building a project. This is necessary to have the actual version of the database.

The problem is that to download the database we need a license key that the user can specify only after installing and configuring the package. A recursive dependency is obtained.

peter-gribanov:
rebuild

peter-gribanov:
rebuild

peter-gribanov:
All is ready

peter-gribanov:
@Nyholm ping

gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
```diff
@@ -0,0 +1,8 @@
+# Read the README file for more detailed configuration information
+# https://github.com/gpslab/geoip2
+#gpslab_geoip:
+    # You should specify your personal licence key
+    #license: ''
```
Nyholm: You should comment this and use an environment variable instead.

gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
```diff
@@ -0,0 +1,8 @@
...
+
+    # Edition ID of database you need
+    #edition: 'GeoLite2-City'
```
Nyholm: Does it make sense to uncomment this too?

gpslab/geoip2/2.0/config/packages/gpslab_geoip.yaml
peter-gribanov: The problem is that the user configures these parameters. The values for these parameters are individual for each user ([see docs](https://github.com/gpslab/geoip2#configuration)). It is not possible to fill these parameters with some predefined data. If you leave these parameters empty, an [error](https://travis-ci.org/symfony/recipes-contrib/jobs/646926681?utm_medium=notification&utm_source=github_status) occurs when building the project. I wrote about this problem [earlier](https://github.com/symfony/recipes-contrib/pull/837#issuecomment-582985103).

I see no other way to solve the problem than to add a commented out config.

peter-gribanov: @Nyholm i can add environment variables, but without setting the default values ​​for them, the project will still not be builded.

```json
{
    "env": {
        "GEOIP_LICENSE": "",
        "GEOIP_EDITION": "GeoLite2-City"
    }
}
```

```yml
#gpslab_geoip:
    # You should specify your personal licence key
    #license: '%env(GEOIP_LICENSE)%'

    # Edition ID of database you need
    #edition: '%env(GEOIP_EDITION)%'
```

Nyholm: Hm.. I would really prefer environment variables. 

How come the project cant be built?

----

I usually reject recipes with only config in comments. But this makes a lot of sense to add config and environment variables.

peter-gribanov: > How come the project cant be built?

I wrote about this problem [earlier](https://github.com/symfony/recipes-contrib/pull/837#issuecomment-582985103).

In order to always have the current version of the GeoIP database, we update it with the composer install/update. This is convenient for quick start of development and quick deployment of the project. But at the first installation of the package, there are no required parameters for installing the GeoIP database. *This was not relevant before, but [MaxMind](https://www.maxmind.com/en/home) changed the policy of using his database.*

peter-gribanov: @Nyholm the ideal config should look like this, but it will [not work](https://travis-ci.org/github/symfony/recipes-contrib/jobs/661017457#L326) because the license key is a required option.

```yml
gpslab_geoip:
    license: '%env(GEOIP_LICENSE)%'
    edition: 'GeoLite2-City'
```

```
# ./bin/console debug:config gpslab_geoip

Fatal error: Uncaught InvalidArgumentException: License for downloaded databases "default" is not specified.
```

Nyholm:
I like this version of the recipe.

Could you add logic to the bundle so it does not download database if `empty($license)`? Does that make sense?

peter-gribanov:
> Could you add logic to the bundle so it does not download database if `empty($license)`? Does that make sense?

This is not right. This means that the bundle is configured incorrectly. This is equivalent to the fact that i will configure the Doctrine, but i will not specify the parameters for connecting to the database.

```yml
doctrine:
    dbal:
        url: '' # not defined
        charset: UTF8
        logging: '%kernel.debug%'
        server_version: '5.6'
```

You can set default connection parameters for example for [MongoDB](https://github.com/symfony/recipes-contrib/blob/master/doctrine/mongodb-odm-bundle/3.3/manifest.json#L14) or [Redis](https://github.com/symfony/recipes-contrib/blob/master/snc/redis-bundle/2.0/manifest.json#L10). But this is not possible for DBAL. With GeoIP the same.

And despite this, the default connection for the Redis is [not configured](https://github.com/symfony/recipes-contrib/blob/master/snc/redis-bundle/2.0/config/packages/snc_redis.yaml#L9-L12). Therefore, i propose to describe the connection parameters as a comment.

peter-gribanov:
The [Symfony Config component](https://symfony.com/doc/current/components/config.html) is designed to structure and validate the config. It is very strange to allow the use incorrect config only so that the package can be installed by Symfony Flex. Your proposed behavior will lead to unexpected results for the user, especially if he uses several databases.

```yml
gpslab_geoip:
    default_database: 'city'
    databases:
        asn:
            edition: 'GeoLite2-ASN'
            license: 'ZS8t22ihCJLiofFa'
            url: 'https://proxy.example.com/GeoLite2-ASN.tar.gz'
            path: '%kernel.project_dir%/var/GeoLite2-ASN.mmdb'
            locales: [ 'ru', 'en' ]
        city:
            edition: 'GeoLite2-City'
            license: 'f5eIQ5R9E8ryNIsz'
            url: 'https://proxy.example.com/GeoLite2-City.tar.gz'
            path: '%kernel.project_dir%/var/GeoLite2-City.mmdb'
            locales: [ 'ru', 'en' ]
        country:
            edition: 'GeoLite2-Country'
            url: 'https://proxy.example.com/GeoLite2-Country.tar.gz'
            path: '%kernel.project_dir%/var/GeoLite2-Country.mmdb'
            locales: [ 'ru', 'en' ]
```

If i do as you suggest, the next config will be valid and the application will work fine, but not 3 databases will be registered in the application, but 2. Debugging this problem can be difficult.

I have already thought about this solution and do not consider it acceptable.

Nyholm:
Sure, I understand your concern. 
But config like this will not break it for doctrine. 
```
doctrine:
    dbal:
        url: 'mysql://localhost'
        charset: UTF8
        logging: '%kernel.debug%'
        server_version: '5.6'
```

Could we find a solution were there is no config in comments and also does not break when you run `composer require`?

peter-gribanov:
I will remove the package configuration example. I do not think this is correct and acceptable, but apparently this is the only solution.

Nyholm:
I can think of some more solutions: 

- The `empty($license)` as mentioned above.
- Add a special "your licence" key. If that is set, you dont download the database on composer install/update. 
- Or just "false" as default license. That could disable the database service in your example
- Not to hook into `composer update/install`. But provide a command to download database files. 

Also note: You can detect differences between if a config value is "not set" and "empty".
-------------

Recipes are meant for easy installation with no or minimal configuration. Adding a config file with just a bunch of comments is not really helpful compared to what it could be.

peter-gribanov:
```yml
doctrine:
    dbal:
        url: 'mysql://localhost'
        charset: UTF8
        logging: '%kernel.debug%'
        server_version: '5.6'
```

This is incorrect and will result in error if the user uses the PostgreSQL or Oracle. The user may not even have the `pdo_mysql` extension installed.

> * Not to hook into `composer update/install`. But provide a command to download database files.

There are commands and one of them is called upon an event from the composer. Disabling event processing breaks the expected behavior, as it was in the [previous version](https://github.com/symfony/recipes-contrib/blob/master/gpslab/geoip2/1.1/manifest.json#L9).

> * The `empty($license)` as mentioned above.
> * Add a special "your licence" key. If that is set, you dont download the database on composer install/update.
> * Or just "false" as default license. That could disable the database service in your example

These are synonyms of the same and lead to the problem i [described earlier](https://github.com/symfony/recipes-contrib/pull/837#issuecomment-598096607).

I am against using *dirty hacks*, but if you insist on adding it, i will do it.

PR #840, opened 2020-02-14
Add elao/enum recipe

ogizanagi:
| Q             | A
| ------------- | ---
| License       | MIT


ogizanagi:
Note: the package is mainly a library, the bundle is only provided as a bridge inside the very same package. The flex-server failure regarding this could be ignored, right?

chalasr:
Failure should be ignored indeed, we have similar recipes (e.g. api-platform https://github.com/symfony/recipes/pull/648#issuecomment-532262725).

Nyholm:
Thank you for the PR and the review

PR #868, opened 2020-02-26
Fix symfony-bundles kafka-bundle config. Removed topics option

Dyachenko:
Q | A
-- | --
Bug fix? | yes
New feature? | no
Deprecations? | no
License | MIT

This PR fixes wrong config option "topics".


khaperets:
CR+

PR #869, opened 2020-02-27
Add 3.0 recipe for "trikoder/oauth2-bundle"

spideyfusion:
| Q             | A
| ------------- | ---
| License       | MIT


spideyfusion:
FYI, the `3.0` version of the bundle no longer supports Symfony 3.4.

HypeMC:
@Nyholm Hi, we're having a little trouble merging this one. Not sure what we're doing wrong, if it's cause of the failing pipeline it's as @spideyfusion said, 3.0 version of the bundle no longer supports Symfony 3.4. Any help on how to proceed from here would be greatly appreciated, tnx.

PR #871, opened 2020-03-05
Monofony recipes

loic425:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/monofony/api-bundle/0.1/manifest.json
+++ b/monofony/api-bundle/0.1/manifest.json
@@ -1,0 +1,9 @@
+{
+    "bundles": {
+        "Monofony\\Bundle\\ApiBundle\\MonofonyApiBundle": ["all"]
+    },
+    "copy-from-package": {
+        "Recipe/config/": "%CONFIG_DIR%/",
+        "Recipe/tests/": "tests/"
+    }
+}
--- a/monofony/core-bundle/0.1/manifest.json
+++ b/monofony/core-bundle/0.1/manifest.json
@@ -4,9 +4,9 @@
     },
     "container": {
         "locale": "fr_FR",
-        "email_contact": "contact@app_name.com",
+        "email_contact": "contact@example.com",
         "email_name": "Contact AppName",
-        "email_sender": "no-reply@app_name.com"
+        "email_sender": "no-reply@example.com"
     },
     "copy-from-package": {
         "Recipe/config/": "%CONFIG_DIR%/",
--- a/monofony/core-bundle/0.1/manifest.json
+++ b/monofony/core-bundle/0.1/manifest.json
@@ -3,7 +3,7 @@
         "Monofony\\Bundle\\CoreBundle\\MonofonyCoreBundle": ["all"]
     },
     "container": {
-        "locale": "fr_FR",
+        "locale": "en",
         "email_contact": "contact@example.com",
         "email_name": "Contact AppName",
         "email_sender": "no-reply@example.com"
--- a/monofony/core-bundle/0.1/manifest.json
+++ b/monofony/core-bundle/0.1/manifest.json
@@ -17,5 +17,9 @@
         "Recipe/tests/": "tests/",
         "Recipe/translations/": "translations/",
         "Recipe/config-builder.js": "config-builder.js"
-    }
+    },
+    "gitignore": [
+        "/public/assets",
+        "/public/media"
+    ]
 }
--- a/monofony/core-bundle/0.1/manifest.json
+++ b/monofony/core-bundle/0.1/manifest.json
@@ -1,0 +1,21 @@
+{
+    "bundles": {
+        "Monofony\\Bundle\\CoreBundle\\MonofonyCoreBundle": ["all"]
+    },
+    "container": {
+        "locale": "fr_FR",
+        "email_contact": "contact@app_name.com",
+        "email_name": "Contact AppName",
+        "email_sender": "no-reply@app_name.com"
+    },
+    "copy-from-package": {
+        "Recipe/config/": "%CONFIG_DIR%/",
+        "Recipe/features/": "features/",
+        "Recipe/spec/": "spec/",
+        "Recipe/src/": "%SRC_DIR%/",
+        "Recipe/templates/": "templates/",
+        "Recipe/tests/": "tests/",
+        "Recipe/translations/": "translations/",
+        "Recipe/config-builder.js": "config-builder.js"
+    }
+}
--- a/monofony/fixtures-plugin/0.1/manifest.json
+++ b/monofony/fixtures-plugin/0.1/manifest.json
@@ -1,0 +1,9 @@
+{
+    "bundles": {
+        "Monofony\\Plugin\\FixturesPlugin\\MonofonyFixturesPlugin": ["all"]
+    },
+    "copy-from-package": {
+        "Recipe/config/": "%CONFIG_DIR%/",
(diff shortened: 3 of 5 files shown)


monofony/core-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+    "bundles": {
+        "Monofony\\Bundle\\CoreBundle\\MonofonyCoreBundle": ["all"]
+    },
+    "container": {
+        "locale": "fr_FR",
```
Pierstoval: Why not `en` to fit more apps? 😉

monofony/core-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+        "email_contact": "contact@app_name.com",
+        "email_name": "Contact AppName",
+        "email_sender": "no-reply@app_name.com"
```
Pierstoval: I'd suggest to use another domain, to avoid having issues if one day the `app_name.com` exists 😛 
(by the way, the `app_name.com` domain name is invalid because domain names can contain only alphanumeric and hyphen characters)
```suggestion
        "email_contact": "contact@example.com",
        "email_name": "Contact AppName",
        "email_sender": "no-reply@example.com"
```

loic425:
@Nyholm Could you have a look why this pull-request is not automatically merged here? 1 owner (@smildev) of the modified packages have approved this PR. @lchrusciel is a private member of these packages but approved too. Thank you

monofony/api-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "bundles": {
+        "Monofony\\Bundle\\ApiBundle\\MonofonyApiBundle": ["all"]
```
Nyholm: I dont understand this bundle. It does not have any code in it...

monofony/core-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,21 @@
...
+        "Monofony\\Bundle\\CoreBundle\\MonofonyCoreBundle": ["all"]
+    },
+    "container": {
+        "locale": "en",
+        "email_contact": "contact@example.com",
+        "email_name": "Contact AppName",
+        "email_sender": "no-reply@example.com"
+    },
+    "copy-from-package": {
+        "Recipe/config/": "%CONFIG_DIR%/",
```
Nyholm: Unrelated note: 

The `Recipe` directory in your bundle is still autoloaded with PSR-4. You should update the autoload config in composer.json.

monofony/fixtures-plugin/0.1/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
+    "bundles": {
+        "Monofony\\Plugin\\FixturesPlugin\\MonofonyFixturesPlugin": ["all"]
```
Nyholm: I dont understand this bundle either. 

- It looks like a library because it is not using any "bundle features" like config or service definitions. 
- It got missing dependencies in composer.json
- It contains 2 interfaces and one abstract class. 

So what does this package do?

monofony/front-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "Monofony\\Bundle\\FrontBundle\\MonofonyFrontBundle": ["all"]
```
Nyholm: This does also look like "bundle" only made for copying data to the users application.

loic425:
> Are you sure you want to create all these "bundles" and recipes? It looks like you want to create a package with type `project`
> 
> ```
> composer create-project monofony/monofony
> ```

Monofony is a Symfony starter based on Sylius internals without the E-commerce part.
The previous version was working like this (just a composer create-project and neither bundles nor recipes). But the problem is to keep your project up to date with new features and bug fixes of this starter.
@Pierstoval helped me to find a solution and Flex recipes was part of this solution.

monofony/core-bundle/0.1/manifest.json
loic425: Yes you're right, I will fix that

monofony/api-bundle/0.1/manifest.json
loic425: It will contains some controller actions on recipes for Api platform for basic endpoints (register, login, forgotten password...).
But I think I can just create a library instead of a bundle.

monofony/fixtures-plugin/0.1/manifest.json
loic425: This package adds the Sylius Fixtures bundle with two abstract classes from Sylius Core Bundle (which depends on too much things to use it directly).
It also contains two data fixtures for the 2 first entities configured with this project (administrators and customers/users).
But you're right, I can create a library instead of a bundle too.

monofony/front-bundle/0.1/manifest.json
loic425: There's a dependency injection configuration:
https://github.com/Monofony/FrontBundle/blob/master/DependencyInjection/MonofonyFrontBundleExtension.php

monofony/api-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,9 @@
...
+    },
+    "copy-from-package": {
+        "Recipe/config/": "%CONFIG_DIR%/",
+        "Recipe/tests/": "tests/"
```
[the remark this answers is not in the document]
loic425: This starter promotes BDD/TDD a lot and recipes contains the tests for all the services and features included in recipes.

Nyholm:
> Monofony is a Symfony starter based on Sylius internals without the E-commerce part.

If this is a starter, shouldn't the user use `composer create-project monofony/starter` to setup the project? There you can add whatever code and config you want.

loic425:
> > Monofony is a Symfony starter based on Sylius internals without the E-commerce part.
> 
> If this is a starter, shouldn't the user use `composer create-project monofony/starter` to setup the project? There you can add whatever code and config you want.

@Nyholm 
The reason is simple. On this starter, we propose some first resources (admin users and customers) and some pages like register, login, forgotten password but this code are not part of the Monofony bundles and can completely be customized. On Symfony flex, The Kernel class is proposed on a recipe too, so I don't understand why we can't do that with Monofony project. Moreover, this allows us to propose upgrades for the resources and services and user can synchronize Monofony recipies via `composer sync-recipes monofony/* --force`. 

@Pierstoval @lchrusciel Can you help explain the concept of using Symfony flex for that?

Pierstoval:
Well, I'm not sure about what I can say more than what Flex already is: a provider for default configuration, which Monofony needs, and in order to make upgrade easier, having its config as Flex recipes seems much easier. The `composer create` is good for the default dependencies and bundles, but Flex is here for the config.

Am I missing something @loic425 ?

monofony/fixtures-plugin/0.1/manifest.json
loic425: @Nyholm @lchrusciel
I think it could be renamed to "monofony/fixtures-pack" cause it contains only composer dependencies and configuration files for the 2 default resources. 

@Nyholm FYI I opened a pull request on Sylius fixtures bundle to include these 2 abstract classes directly on the Sylius package. https://github.com/Sylius/SyliusFixturesBundle/pull/29

Nyholm:
I'll close this because I merged #1342

PR #874, opened 2020-03-14
Update post-install.txt

ThomasLandauer:
Minor improvements, some mentioned at https://github.com/Codeception/Codeception/issues/5438

| Q             | A
| ------------- | ---
| License       | MIT


ThomasLandauer:
No, `yml` is correct! I just double-checked this with a brand-new installation:

```bash
symfony new Symfony5
cd Symfony5
composer require "codeception/codeception" --dev
```

The generated file's actual name is `codeception.yml`. But the message says:

> File codeception.yaml created

And this is the mismatch I wanted to fix.

See also https://github.com/Codeception/Codeception/issues/5661

PR #876, opened 2020-03-16
Added HWIOAuthBundle version 1.0 support

stloyd:
| Q             | A
| ------------- | ---
| License       | MIT


stloyd:
@Nyholm In fact that's true... I was thinking if I could force adding required packages (which are in fact a bit optional)... but not luck ;)

PR #878, opened 2020-03-18
Recipe for monsieurbiz/sylius-alert-message-plugin

jacquesbh:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/monsieurbiz/sylius-alert-message-plugin/1.0/config/packages/monsieurbiz_sylius_alert_message_plugin.yaml
+++ b/monsieurbiz/sylius-alert-message-plugin/1.0/config/packages/monsieurbiz_sylius_alert_message_plugin.yaml
@@ -1,0 +1,2 @@
+imports:
+    - { resource: "@MonsieurBizSyliusAlertMessagePlugin/Resources/config/config.yaml" }
--- a/monsieurbiz/sylius-alert-message-plugin/1.0/config/routes/monsieurbiz_sylius_alert_message_plugin.yaml
+++ b/monsieurbiz/sylius-alert-message-plugin/1.0/config/routes/monsieurbiz_sylius_alert_message_plugin.yaml
@@ -1,0 +1,3 @@
+monsieurbiz_sylius_alert_message_admin:
+    resource: "@MonsieurBizSyliusAlertMessagePlugin/Resources/config/routing/admin.yaml"
+    prefix: /admin
--- a/monsieurbiz/sylius-alert-message-plugin/1.0/manifest.json
+++ b/monsieurbiz/sylius-alert-message-plugin/1.0/manifest.json
@@ -1,0 +1,10 @@
+{
+    "bundles": {
+        "MonsieurBiz\\SyliusAlertMessagePlugin\\MonsieurBizSyliusAlertMessagePlugin": [
+            "all"
+        ]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


jacquesbh:
@Nyholm I think this one is kind of like https://github.com/symfony/recipes-contrib/pull/867 but I have no time right now to create the recipe for the missing configuration.

Thank you!

monsieurbiz/sylius-alert-message-plugin/1.0/config/packages/monsieurbiz_sylius_alert_message_plugin.yaml
```diff
@@ -0,0 +1,2 @@
+imports:
+    - { resource: "@MonsieurBizSyliusAlertMessagePlugin/Resources/config/config.yaml" }
```
Nyholm, line 2: This looks strange. What is happening herer?

jacquesbh: What is strange?

The fact that it's not a `Bundle` but a `Plugin`?

See -> https://github.com/symfony/flex/pull/426

Nyholm: I find it strange that you need to add a file that only imports a config file from your plugin. 

Why dont you include that file [here](https://github.com/monsieurbiz/SyliusAlertMessagePlugin/blob/master/src/DependencyInjection/MonsieurBizSyliusAlertMessageExtension.php#L21) just as you do with `services.yaml`?

jacquesbh: Because I use the SyliusResourceBundle. And if I don't do it this way I get the following error:

`There is no extension able to load the configuration for "sylius_resource"…`

So… 🤷‍♂️

Nyholm: You get that error because a bundle cannot modify other bundles config. Like a recipe should not modify other bundles config. You are trying to hide config to other bundles by using this import. 

=)

That is why I think this is weird. 

--------

I see that you are adding to the `grid` collection in the GridBundle and similarly adding to a collection in the Resource bundle. That is fine to have in a recipe.

jacquesbh: It's a merge then! 🥰

jacquesbh: Or did I miss something @Nyholm?

Nyholm:

monsieurbiz/sylius-alert-message-plugin/1.0/config/packages/monsieurbiz_sylius_alert_message_plugin.yaml
Nyholm: It took me a while to be okey with this Sylius thing. Sorry for the wait.

jacquesbh: No harm done! Thank you!

PR #879, opened 2020-03-19
Add tilleuls/migration-bundle

vincentchalamon:
| Q             | A
| ------------- | ---
| License       | MIT
| Related issue | https://github.com/coopTilleuls/CoopTilleulsMigrationBundle/issues/9


tilleuls/migration-bundle/1.1/config/packages/coop_tilleuls_migration.yaml
```diff
@@ -0,0 +1,11 @@
+# Read the documentation: https://github.com/coopTilleuls/CoopTilleulsMigrationBundle/blob/master/Resources/doc/getting_started.md
+coop_tilleuls_migration:
+    legacy_connection_name: legacy
+
+doctrine:
+    dbal:
+        default_connection: default
```
Nyholm, line 7: This property is not needed. That will conflict with the users existing config, right?

PR #882, opened 2020-03-21
Add kerox/twig-image-placeholder-extension

ker0x:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/kerox/twig-image-placeholder-extension/1.0/config/packages/kerox_twig_image_placeholder.yaml
+++ b/kerox/twig-image-placeholder-extension/1.0/config/packages/kerox_twig_image_placeholder.yaml
@@ -1,0 +1,5 @@
+services:
+    Kerox\TwigImagePlaceholder\SvgExtension:
+        autowire: true
+        autoconfigure: true
+
--- a/kerox/twig-image-placeholder-extension/1.0/manifest.json
+++ b/kerox/twig-image-placeholder-extension/1.0/manifest.json
@@ -1,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


kerox/twig-image-placeholder-extension/1.0/config/packages/kerox_twig_image_placeholder.yaml
```diff
@@ -0,0 +1,7 @@
+services:
+    _defaults:
+        public: false
+        autowire: true
+        autoconfigure: true
+
+    Kerox\TwigImagePlaceholder\SvgExtension: null
```
Nyholm, line 7: Why "null"? That looks strange. You mean tilde (`~`) right?

Also, one more thing.. It might just be a personal preference and you dont need to change. I found it strange to have one service  and still use `_defaults`. This file could be replaced with: 

```yaml
services:
  Kerox\TwigImagePlaceholder\SvgExtension:
    autowire: true
    autoconfigure: true
```

This is both less config and more clear.

ker0x: @Nyholm, thank you for your review.

I have actually take example on the `twig_extensions.yaml` configuration file that use `_defaults` and `null`.

```yaml
services:
    _defaults:
        public: false
        autowire: true
        autoconfigure: true

    # Uncomment any lines below to activate that Twig extension
    #Twig\Extensions\ArrayExtension: null
    #Twig\Extensions\DateExtension: null
    #Twig\Extensions\IntlExtension: null
    #Twig\Extensions\TextExtension: null
```

`~` be a way to define null value, I'm fine to use it. For the `_default`, it was in case I decided to add new extension to this library, but currently it's not planed so I changed with your suggestion.

Nyholm: There is technically a difference between tilde and null. But most of the time it does not matter. 

If you want to, make a new PR to update `twi_extensions.yaml` to use tilde.

> For the _default, it was in case I decided to add new extension to this library, but currently it's not planed so I changed with your suggestion.

It's fine either way. It is just a personal preference.

PR #883, opened 2020-03-23
Set "sonata_admin.options.legacy_twig_text_extension" option

phansys:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/sonata-project/admin-bundle/3.64/config/packages/sonata_admin.yaml
+++ b/sonata-project/admin-bundle/3.64/config/packages/sonata_admin.yaml
@@ -1,0 +1,12 @@
+sonata_admin:
+    title: 'Sonata Admin'
+    dashboard:
+        blocks:
+            - { type: sonata.admin.block.admin_list, position: left }
+    options:
+        legacy_twig_text_extension: false
+
+sonata_block:
+    blocks:
+        sonata.admin.block.admin_list:
+            contexts: [admin]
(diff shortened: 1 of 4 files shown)


sonata-project/admin-bundle/3.64/config/packages/sonata_admin.yaml
```diff
@@ -0,0 +1,12 @@
+sonata_admin:
+    title: 'Sonata Admin'
+    dashboard:
+        blocks:
+            - { type: sonata.admin.block.admin_list, position: left }
+    options:
+        legacy_twig_text_extension: false
+
+sonata_block:
```
Nyholm, line 9: Is this a dependency of admin bundle?

Generally, recipes should not modify other bundles config. But sonata_block.blocks is a collection, right? Like framework.cache.pools

phansys: >Is this a dependency of admin bundle?

Yes, this node is provided by "sonata-project/block-bundle"

>Generally, recipes should not modify other bundles config. But sonata_block.blocks is a collection, right? Like framework.cache.pools

I understand. Yes, `sonata_block.blocks` is a collection. Is there a way to prevent collisions with other bundles or previous configuration in a specific node? I mean, something like a merge or similar approach.

Nyholm: No. There is no way (as I know) to prevent that. But adding things to collections is fine.

PR #915, opened 2020-04-22
add enhavo recipes

gseidel:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Thank you for this PR. 

It seams like you added quite a lot of files. Not only config and some PHP but there is javascript and images too. Why are these files needed in the recipe? Why can't they stay in the package?

gseidel:
@Nyholm Thanks for review my PR. I'm sorry for the many files, but enhavo has a lot of bundles already i want to add recipes for them.

Indeed, there are some javascript files, but it's important they stay outside of the package, because they need to be editable during the work with enhavo like other configuration files. I try to keep them as short as possible. The most of the assets code live inside the (node) packages.

The images just used for asset demonstration, but will be also included via webpack.

Nyholm:
Travis is failing and I fear this PR will be open forever. Could you submit one bunde at the time? Maybe start with 2-4 of them so I can learn about Enhavo system and you can learn about what is a good recipe and not. 

I think that is the best way forward.

gseidel:
@Nyholm Thanks again for your feedback. I follow your advice and add only the basic bundles to start with enhavo. I also add more work for a good documentation and minimize the required steps and files.

The current version is only compatible with symfony 4, so the other CI tests will keep fail.

enhavo/dashboard-bundle/0.8/assets/enhavo/dashboard.ts
```diff
@@ -0,0 +1,2 @@
+import Application from "@enhavo/dashboard/Dashboard/DashboardApplication";
+Application.getVueLoader().load(() => import("@enhavo/dashboard/Dashboard/Components/ApplicationComponent.vue"));
```
Nyholm, line 2: Why is this file needed in the recipe?

enhavo/dashboard-bundle/0.8/config/packages/enhavo_dashboard.yaml
```diff
@@ -0,0 +1,2 @@
+imports:
+    - { resource: '@EnhavoDashboardBundle/Resources/config/app/config.yml' }
```
Nyholm, line 2: Shouldn't the bundle import this file?

enhavo/dashboard-bundle/0.8/assets/enhavo/dashboard.ts
gseidel: This is the entry point for the dashboard. It will be used to add js and css or load further assets from other enhavo bundles

enhavo/dashboard-bundle/0.8/config/packages/enhavo_dashboard.yaml
gseidel: A lot of configuration files in enhavo will overwrite other bundle settings. So i think they can't be just "imported" by the bundle. The symfony documentation said, to load these kinds of configuration over ``PrependExtensionInterface``.  That is an ongoing work in enhavo and will be used for ``0.9``. For enhavo ``0.8`` they are still loaded over the configuration

enhavo/dashboard-bundle/0.8/assets/enhavo/dashboard.ts
Nyholm: But you cannot be sure this is added to webpack.. It make little sense. 

Instead of asking the user to include this file in webpack.js, then you may ask them to include the file from your bundle instead.

enhavo/dashboard-bundle/0.8/config/packages/enhavo_dashboard.yaml
Nyholm: > A lot of configuration files in enhavo will overwrite other bundle settings

That is not really a preferred thing.. In fact. I would suggest to try different solution.

gseidel:
I will close that PR to reopen a new one soon with discussed fixes and only one bundle at a time

PR #916, opened 2020-04-23
[jacquesndl/mailer-bundle] add recipe

jacquesndl:
| Q             | A
| ------------- | ---
| License       | MIT


jacquesndl:
Hello,

I tried to create a recipe for my bundle [MailerBundle](https://github.com/jacquesndl/MailerBundle) but the CI fails with skeleton 3.4 while the bundle is available only for skeleton >4.4.
How do I handle this ?
Thanks in advance for your help.

Best regards

PR #923, opened 2020-05-03
Allow jobs to fail on SF 3.4.

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT

There are may people asking why travis is failing and if it is okey to merge the PR without support on 3.4. 

I thought I would make it easier for them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,22 +31,22 @@ install:
 
 before_script:
   - |
-      install_package() {
-          composer require --ignore-platform-reqs $1
-          EXIT_CODE=$?
+    install_package() {
+        composer require --ignore-platform-reqs $1
+        EXIT_CODE=$?
 
-          MESSAGE="";
-          if [[ EXIT_CODE -eq 0 ]]; then
+        MESSAGE="";
+        if [[ EXIT_CODE -eq 0 ]]; then
             exit 0;
-          elif [[ EXIT_CODE -eq 1 ]]; then
+        elif [[ EXIT_CODE -eq 1 ]]; then
             MESSAGE="Could not install package";
-          elif [[ EXIT_CODE -eq 2 ]]; then
+        elif [[ EXIT_CODE -eq 2 ]]; then
             MESSAGE="Could not resolve dependencies.\n# You can ignore this error if your package does not support Symfony $SKELETON_VERSION";
-          fi
+        fi
 
-          echo -e "#\n#\n#\n# $MESSAGE\n#\n#\n#\n";
-          exit $EXIT_CODE;
-     }
-     export -f install_package
+        echo -e "#\n#\n#\n# $MESSAGE\n#\n#\n#\n";
+        exit $EXIT_CODE;
+    }
+    export -f install_package
 
 script: install_package $(echo $PACKAGES)
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,4 +30,17 @@ install:
     - composer config extra.symfony.allow-contrib true
 
 script:
-    - composer req --ignore-platform-reqs $(echo $PACKAGES)
+    - |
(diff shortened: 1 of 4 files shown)


xabbuh:
Could we maybe somehow check in the `3.4` job whether or not the package in question supports Symfony 3.4 and based on that decide whether or not to run the full checks?

Nyholm:
Yeah, that would be super helpful. I'll see if I can improve the script this weekend. 

Good idea!

wouterj:
https://getcomposer.org/doc/03-cli.md#process-exit-codes

Seems like a "Dependency solving error code" has a different exit code (2) from any other errors. So I think you can solve it by allowing exit code `2` in 3.4 builds (but I'm not that much of a linux wizard to know how)

.travis.yml
```diff
@@ -27,4 +27,18 @@ install:
...
+        exit 0;
+      elif [[ EXIT_CODE -eq 1 ]]; then
+        echo "Could not install package";
+        exit 1;
+      elif [[ EXIT_CODE -eq 2 ]]; then
+        echo "Could not resolve dependencies. You can ignore this error if your package does not support Symfony $SKELETON_VERSION";
+        exit 1;
+      fi
+
+      exit $EXIT_CODE;
```
wouterj, line 43: I think the Travis output is more readable if we transform this into a function set in `before_script`, which is then used by the `script`:

```yaml
before_script:
  - |
      install_package() {
          composer req --ignore-platform-reqs $1
          local exit_code = $?
          echo "Exit code: $exit_code"
          if [[ exit_code -eq 0 ]]; then
            exit 0;
          elif [[ exit_code -eq 1 ]]; then
            echo "Could not install package";
            exit 1;
          elif [[ exit_code -eq 2 ]]; then
            echo "Could not resolve dependencies. You can ignore this error if your package does not support Symfony $SKELETON_VERSION";
            exit 1;
          fi
          exit $exit_code;
     }
     export -f install_package

script: install_package $(echo $PACKAGES)
```

Nyholm: Yeah. You are correct. I'll keep updating. 

The script works now: https://travis-ci.org/github/symfony/recipes-contrib/jobs/701979394

Nyholm:
Im done with this PR now. I snuck in an update to async-aws to make sure travis is working properly. 

Screenshot 2020-06-25 at 12 57 57

.travis.yml
```diff
@@ -26,5 +29,24 @@ install:
...
+        EXIT_CODE=$?;
+
+        MESSAGE="";
+        if [[ EXIT_CODE -eq 0 ]]; then
+            exit 0;
+        elif [[ EXIT_CODE -eq 1 ]]; then
+            MESSAGE="Could not install package and configure package.";
+        elif [[ EXIT_CODE -eq 2 ]]; then
+            MESSAGE="Could not resolve dependencies.\n# You can ignore this error if your package does not support Symfony $SKELETON_VERSION";
+        fi
```
wouterj, line 45:
```suggestion
            MESSAGE="Could not resolve dependencies.\n# You can ignore this error if your package does not support Symfony $SKELETON_VERSION";
            exit 0
        fi
```

Shouldn't we make the job green when the package could not be installed?

Nyholm: We are not. We always exit with the same exit code as composer

Nyholm:
Example error because of poor config in recipe

Screenshot 2020-06-25 at 13 29 47

.travis.yml
Nyholm: I dont think we should mark the job as green. Or should we....
I dont want to have 3 green travis job but the package could not be installed .

Nek-:
@wouterj merging this would be greaaat :)

fabpot (approved):

PR #924, opened 2020-05-03
[AsyncAws Bundle] Add commented config for secrets

jderusse:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/async-aws/async-aws-bundle/0.1/config/packages/async_aws.yaml
+++ b/async-aws/async-aws-bundle/0.1/config/packages/async_aws.yaml
@@ -1,9 +1,12 @@
 # Read documentation at https://github.com/async-aws/symfony-bundle
 async_aws:
-    config:
         region: eu-central-1
         accessKeyId: '%env(AWS_KEY)%'
         accessKeySecret: '%env(AWS_SECRET)%'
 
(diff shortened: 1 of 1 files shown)


async-aws/async-aws-bundle/0.1/config/packages/async_aws.yaml
```diff
@@ -1,9 +1,12 @@
 # Read documentation at https://github.com/async-aws/symfony-bundle
 async_aws:
-    config:
```
Nyholm, line 3: This key should not be removed.

async-aws/async-aws-bundle/0.1/config/packages/async_aws.yaml
```diff
@@ -1,9 +1,12 @@
...
         region: eu-central-1
         accessKeyId: '%env(AWS_KEY)%'
         accessKeySecret: '%env(AWS_SECRET)%'
 
     # All installed clients are automatically autowired. Use this section to customize clients
     clients:
+
+    # Use AWS SSM to store secrets or parameters.
+    #secrets:
+        #path: /parameters/my-project
```
Nyholm, line 12: Even though I really like this feature, I cannot accept config in comments. The recipe is not a replacement for documentation.

PR #930, opened 2020-05-11
adding localise bundle recipe

mradhi:
| Q             | A
| ------------- | ---
| License       | MIT


cosavostra/localise-bundle/1.2/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "bundles": {
+        "CosaVostra\\LocaliseBundle\\CosaVostraLocaliseBundle": ["all"]
+    },
+    "env": {
+        "LOCALISE_EXPORT_KEY": "localise_api_key_to_replace"
```
Nyholm, line 6: I dont see that this is used by your bundle's configuration.

cosavostra/localise-bundle/1.2/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+    },
+    "gitignore": [
+        "!/translations/.gitignore",
+        "!/translations/.gitkeep",
```
Nyholm, line 10: `.gitkeep` is not a standard file name.

cosavostra/localise-bundle/1.2/manifest.json
mradhi: @Nyholm Check https://github.com/mradhi/localise-bundle/blob/master/Resources/config/services.xml, I used this env variable in `CosaVostra\LocaliseBundle\Http\Client` service

Nyholm: You are referencing me to your service definition. 
Your bundle dont even have configuration: https://github.com/mradhi/localise-bundle/tree/master/DependencyInjection

The issue with your current setup is that if a user is missing the environment variable, they will not get an exception until runtime.
I suggest you create some configuration to your bundle and make sure to use that env var there. If the env var is missing then, the user will get an error at build time, which is much more preferred.

Nyholm: Friendly ping

Nyholm (changes requested):

Nyholm:
Feel free to make a new PR if you are still interested to finish this.

PR #931, opened 2020-05-11
Add recipe for fresh/centrifugo-bundle

fre5h:
| Q             | A
| ------------- | ---
| License       | MIT

Recipes for my new bundle [https://github.com/fre5h/CentrifugoBundle](https://github.com/fre5h/CentrifugoBundle)


fresh/centrifugo-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,14 @@
...
+        "Fresh\\CentrifugoBundle\\FreshCentrifugoBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "src/": "%SRC_DIR%/"
+    },
+    "env": {
+        "CENTRIFUGO_API_KEY": "%generate(secret)%",
+        "CENTRIFUGO_API_ENDPOINT": "http://localhost:8000/api",
+        "CENTRIFUGO_SECRET": "%generate(secret)%"
```
Nyholm, line 12: I dont see that these environment variables are used by your bundle's config.

fresh/centrifugo-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,14 @@
+{
+    "bundles": {
+        "Fresh\\CentrifugoBundle\\FreshCentrifugoBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/",
+        "src/": "%SRC_DIR%/"
```
Nyholm, line 7: Is this really needed? 

Does your bundle require this exact file path?

fresh/centrifugo-bundle/1.0/manifest.json
fre5h: here
https://github.com/fre5h/CentrifugoBundle/blob/master/Resources/config/services.yaml#L9
https://github.com/fre5h/CentrifugoBundle/blob/master/Resources/config/services.yaml#L25

fresh/centrifugo-bundle/1.0/manifest.json
fre5h: No, it is not required. It is just a small help. It creates a directory `src/Service/Centrifugo/ChannelAuthenticator/` where I recommend to put channel authenticators. I use Centrifugo for a quite long time on several projects. And everywhere there are private channels. And to get access to private channel user should be granted with additional token, so ChannelAuthenticators are quite popular things. I decided to put it to recipe, so new developers would notice this directory in their projects, then take a look in the documentations about private channels https://github.com/fre5h/CentrifugoBundle/blob/master/Resources/docs/authentication.md#create-own-channel-authenticator and so would have an easier setup of my bundle.

Nyholm: Im not convinced this is more helpful than annoying to be honest. 

A user may have decided to put all their authentication related stuff in `src/Auth`. Now this recipe is trying to force a coding standard on that user that isnt really needed.

fresh/centrifugo-bundle/1.0/manifest.json
Nyholm: That is your service definition. I meant your [Configuration](https://github.com/fre5h/CentrifugoBundle/blob/master/DependencyInjection/Configuration.php#L24)

Nyholm: I dont see the point of your package being a bundle if you don't. 

Your package could easily be a library instead.

fre5h: My bundle uses DataCollector and adds info into Symfony Profiler Toolbar https://github.com/fre5h/CentrifugoBundle/blob/master/DataCollector/CentrifugoCollector.php
Also it has a little configuration, which can be edited, and it is loaded by Symfony https://github.com/fre5h/CentrifugoBundle/blob/master/DependencyInjection/Configuration.php
Also it registers few services in Symfony DI https://github.com/fre5h/CentrifugoBundle/blob/master/Resources/config/services.yaml, uses autowiring and autobiding and tagged services for registering custom channel authenticators
And also bundle registers several console commands.

I didn't implement it as library, because I wanted a tight integration with Symfony. I wanted to see all logs in Symfony Profiler and it requires to add logs inside http client for Centrifugo.
@Nyholm

Nyholm: Sorry if Im unclear. 

My point is: If you wanted to use environment variables for instead of config, there would be no use for this package to be a bundle. The "tight integration" could be achieved with a normal library and a recipe. 

The benefit of a bundle (and I do think you should use a bundle for this) is that you can have configuration and build time validation. 

The current state of your bundle has a major drawback. If I dont define the environment variable `CENTRIFUGO_API_ENDPOINT`, then I would get an error in runtime. If you are using proper configuration for this, I would see the error in buildtime. 

I dont think it is a good idea to merge a recipe that define environment variables that are not properly used. That is why I requested changes to this PR.

fre5h: Ok. I understand

Nyholm:
Friendly ping

fre5h:
@Nyholm fixed

Nyholm (approved):
Thank you

PR #935, opened 2020-05-18
Feature/lti bundle recipe

ekkinox:
| Q             | A
| ------------- | ---
| License       | MIT

- Added default config files
- Added instructions for development keys generation


ekkinox:
@Nyholm hello :)

Can it me merged since this bundle does not aim to support SF 3.4 ? see https://travis-ci.org/github/symfony/recipes-contrib/builds/688304131?utm_source=github_status&utm_medium=notification

oat-sa/bundle-lti1p3/2.0/config/secrets/dev/private.key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
Nyholm, line 1: I dont think it is a good idea to have a private key in the recipe. 
I understand it is just for debugging but the term "private" do have a meaning. =)

ekkinox: true :) its just a key pair to make the bundle ready out of the box, only for dev env, with a warning very explicit on bundle install to not use them in prod.

I see 2 options:

Nyholm: I think the latter is better. 

The former will make the key unusable.

ekkinox: yes indeed, fixed :)

PR #937, opened 2020-05-19
recipe for mrsuh/json-validation-bundle

mrsuh:
| Q             | A
| ------------- | ---
| License       | MIT


mrsuh:
Hi.
Thx for review.
It's right. Recipe just adds a config file.

Nyholm:
There is no point of using a recipe then. Flex is smart enough to install this bundle without a recipe and the result will be exactly the same.

PR #938, opened 2020-05-20
Add elfinder:install to composer-scripts

zorn-v:
This allow install/update elfinder assets on install/update packages

| Q             | A
| ------------- | ---
| License       | MIT


helios-ag:
@Nyholm If you have time, it would be nice to merge this patch, thanks!

PR #944, opened 2020-05-24
Add recipe for Jurry RabbitMQ Symfony Bundle

wajdijurry:
| Q             | A
| ------------- | ---
| License       | MIT
| Description | RabbitMQ Symfony Bundle

### DIFF
--- a/jurry/amqp-symfony-bundle/1.0/config/packages/jurry_rabbitmq.yaml
+++ b/jurry/amqp-symfony-bundle/1.0/config/packages/jurry_rabbitmq.yaml
@@ -1,11 +1,5 @@
 amqp_handler:
-#    You can define connection in both ways: as array or as a string
-#    connection: 'tcp://guest:guest@172.18.0.01:5672'
-    connection:
-        host: 172.18.0.1
-        port: 5672
-        user: guest
-        password: guest
+    connection: 'tcp://guest:guest@172.18.0.01:5672'
     queues_properties:
         sync_queue:
             name: stores_sync
--- a/jurry/amqp-symfony-bundle/1.0/config/packages/jurry_rabbitmq.yaml
+++ b/jurry/amqp-symfony-bundle/1.0/config/packages/jurry_rabbitmq.yaml
@@ -1,5 +1,5 @@
 amqp_handler:
-    connection: 'tcp://guest:guest@localhost:5672'
+    connection: '%env'
     queues_properties:
         sync_queue:
             name: stores_sync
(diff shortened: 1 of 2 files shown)


jurry/amqp-symfony-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "bundles": {
+        "Jurry\\RabbitMQ\\AmqpHandlerBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "gitignore": [
+        ".idea/"
```
xabbuh, line 9: IMO this line should be removed.

wajdijurry:
Note: Re-request review button is not working

wajdijurry:
Hi, I followed these steps to test the recipe,
https://flex.symfony.com/r/github.com/symfony/recipes-contrib/944
and it works fine, what's next ?
@xabbuh

jurry/amqp-symfony-bundle/1.0/config/packages/jurry_rabbitmq.yaml
```diff
@@ -0,0 +1,15 @@
+amqp_handler:
+#    You can define connection in both ways: as array or as a string
+#    connection: 'tcp://guest:guest@172.18.0.01:5672'
```
dbrumann, line 3: Maybe use a string dsn instead and pass it in via `.env`?

wajdijurry: Developer can define the connection in both ways, as a string or as an array. This increases the flexibility.

dbrumann: If both are equivalent how does it make it more flexible? Currently "best practice" is to use a single DSN that is set via .env (see Doctrine, Messenger, Mailer), instead of having 4 different parameters (host, port, user, password) that are set independently thus increasing the chance of errors. If a bundle supports a dsn I think for a consistent DX it makes sense to use that instead.

wajdijurry: Ok, will do

wajdijurry: Connection now is handled always as string

jurry/amqp-symfony-bundle/1.0/config/packages/jurry_rabbitmq.yaml
```diff
@@ -0,0 +1,9 @@
+amqp_handler:
+    connection: '%env(RABBIT_MQ_DSN)%'
```
dbrumann, line 2: You might want to add something like a vendor prefix, e.g. `JURRY_RABBITMQ_DSN`, to prevent clashes with other bundles.

wajdijurry: Prefix added

Nyholm:

Nyholm:
Strange..

PR #945, opened 2020-05-25
Public directory can be changed in project

peter-gribanov:
| Q             | A
| ------------- | ---
| License       | MIT

See correct example:
https://github.com/symfony/recipes/blob/master/symfony/framework-bundle/4.2/manifest.json#L25


peter-gribanov:
PS: `liip/imagine-bundle:1.8` is not compatible with Symfony 4.4 and 5

PR #946, opened 2020-05-25
feat(optimus): add recipe for ckrack/optimus-bundle

ckrack:
| Q             | A
| ------------- | ---
| License       | MIT

Add recipe for my bundle for [jenssegers/optimus](https://github.com/jenssegers/optimus)

Would be great to have a post-install-hook for recipes, as optimus could auto-configure with sensible values. Optimus needs some calculated prime numbers for instantiation. These should be different in each installation. To not fail during the composer install, we must use some default values. Of course we can use `post-install.txt` to tell people to exchange these defaults in the config/env. I'd like to add auto-configuration, though. There is no post-install-hook for bundles/recipes, so the only way I came up with is adding a script to `auto-scripts` with a `composer-scripts`-entry.
This script would replace the default values in the config with auto-generated ones. By replacing the defaults, this would only act once, even though the script itself will be invoked on every update/install.

I have two issues I'd like to discuss:
1. Should I use only the config file, without any env vars preferably?
2. Is the described way of auto-configuration via composer-scripts something we should do?


ckrack/optimus-bundle/0.1/config/packages/ckrack_optimus.yaml
```diff
@@ -0,0 +1,13 @@
+ckrack_optimus:
+
+    # Large prime number lower than 2147483647
```
Nyholm, line 3: Is a prime considered "large" if it is smaller than 2^31?

Isn't RSA using a few thousand bits log primes?

ckrack: I think what's meant is, large but lower than 2^31.
Anyway, it's a direct quote from https://github.com/jenssegers/optimus#usage

I'll probably replace the comments in env with an explanation to generate via the spark command.

ckrack:
> 1. Is the described way of auto-configuration via composer-scripts something we should do?
> 
> Why is that needed? Because you want to help the user generating primes?

Yes, the numbers can be auto-generated and are random by default, making each installation kind of unique.
Optimus' constructor throws an Exception when the values are invalid and this will lead to an error when auto-scripts are run after installation. So we need some defaults, but we want the user to change them (or automatically generate them).

Nyholm:
The easiest way is to tell the user what to do in a post-install message.

ckrack:
> The easiest way is to tell the user what to do in a post-install message.

That's already in place. So, ready from my side.

PR #955, opened 2020-06-10
Add recipe for connectholland/timchimp-bundle

reyostallenberg:
| Q             | A
| ------------- | ---
| License       | MIT


connectholland/timechimp-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "ConnectHolland\\TimechimpBundle\\ConnecthollandTimechimpBundle": ["all"]
+    },
+    "env": {
+        "TIMECHIMP_ACCESS_TOKEN": "baerer-token-value"
```
Nyholm, line 6: I dont see that this is used by the bundle's config.

reyostallenberg:
It's used here https://github.com/ConnectHolland/timechimp-bundle/blob/36e6b5209e2b2dc499281c19eaa084f89ad25d35/src/Resources/config/services.yaml#L11 

The rest is auto wired

Nyholm:
The issue with your current setup is that if a user is missing the environment variable, they will not get an exception until runtime. 
I suggest you create some configuration to your bundle and make sure to use that env var there. If the env var is missing then, the user will get an error at build time, which is much more preferred.

reyostallenberg:
@Nyholm then this is more like you prefer, I think.

PR #969, opened 2020-06-22
Add reciepe for skeletton.

ScyLabs:
| Q             | A
| ------------- | ---
| License       | MIT

Add reciepe to scylabs/neptune-installer to create custom skeleton

PR #971, opened 2020-06-25
adding scylabs/neptune-installer recipe

ScyLabs:
| Q             | A
| ------------- | ---
| License       | MIT

Add recipe to scylabs/neptune-installer package.
Package to install base for scylabs/neptune-bundle package.
Sending again following a misunderstanding.

The reciepe will be used to transfer the necessary scripts / css to my scylabs / neptune-installer package.

It's not just a skeleton.


Nyholm:
This is a duplicate of #969

I see that you've loaded a bunch of scripts in https://github.com/ScyLabs/neptune-installer
I understand that you want them copied to a project. 

Flex recipes is not the way to do this.

PR #983, opened 2020-07-14
Add flex recipe for SuluArticleBundle

luca-rath:
| Q             | A
| ------------- | ---
| License       | MIT

Link to package: https://github.com/sulu/SuluArticleBundle

### DIFF
--- a/sulu/article-bundle/2.0/config/packages/sulu_article.yaml
+++ b/sulu/article-bundle/2.0/config/packages/sulu_article.yaml
@@ -1,0 +1,16 @@
+sulu_article:
+    index_name: '%env(resolve:ELASTICSEARCH_INDEX)%'
+    hosts:
+        - '%env(resolve:ELASTICSEARCH_HOST)%'
+
+sulu_route:
+    mappings:
+        Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
+            generator: schema
+            options:
+                route_schema: /articles/{object.getTitle()}
+        Sulu\Bundle\ArticleBundle\Document\ArticlePageDocument:
+            generator: article_page
+            options:
+                route_schema: {translator.trans("page")}-{object.getPageNumber()}
+                parent: {object.getParent().getRoutePath()}
--- a/sulu/article-bundle/2.0/config/packages/sulu_article.yaml
+++ b/sulu/article-bundle/2.0/config/packages/sulu_article.yaml
@@ -2,6 +2,9 @@ sulu_article:
     index_name: '%env(resolve:ELASTICSEARCH_INDEX)%'
     hosts:
         - '%env(resolve:ELASTICSEARCH_HOST)%'
+    types:
+        article:
+            translation_key: "sulu_article.article"
 
 sulu_route:
     mappings:
@@ -14,3 +17,12 @@ sulu_route:
             options:
                 route_schema: '{translator.trans("page")}-{object.getPageNumber()}'
                 parent: '{object.getParent().getRoutePath()}'
+
+ongr_elasticsearch:
+    analysis:
+        tokenizer:
+            pathTokenizer:
+                type: path_hierarchy
+        analyzer:
+            pathAnalyzer:
+                tokenizer: pathTokenizer
--- a/sulu/article-bundle/2.0/config/packages/sulu_article.yaml
+++ b/sulu/article-bundle/2.0/config/packages/sulu_article.yaml
@@ -8,9 +8,9 @@ sulu_route:
         Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
             generator: schema
             options:
-                route_schema: /articles/{object.getTitle()}
+                route_schema: '/articles/{object.getTitle()}'
         Sulu\Bundle\ArticleBundle\Document\ArticlePageDocument:
             generator: article_page
             options:
-                route_schema: {translator.trans("page")}-{object.getPageNumber()}
-                parent: {object.getParent().getRoutePath()}
+                route_schema: '{translator.trans("page")}-{object.getPageNumber()}'
+                parent: '{object.getParent().getRoutePath()}'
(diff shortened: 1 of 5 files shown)


Nyholm:
FYI: A recipe should not modify other bundles config.

sulu/article-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,16 @@
...
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "copy-from-package": {
+        "Resources/doc/default.xml": "%CONFIG_DIR%/templates/articles/default.xml",
+        "Resources/doc/default.html.twig": "templates/articles/default.xml"
+    },
+    "env": {
+        "ELASTICSEARCH_HOST": "127.0.0.1:9200",
+        "ELASTICSEARCH_INDEX": ""
```
alexander-schranz, line 14: should be `su_articles` or something like that.

sulu/article-bundle/2.0/config/packages/sulu_article.yaml
```diff
@@ -0,0 +1,16 @@
...
+    mappings:
+        Sulu\Bundle\ArticleBundle\Document\ArticleDocument:
+            generator: schema
+            options:
+                route_schema: '/articles/{object.getTitle()}'
+        Sulu\Bundle\ArticleBundle\Document\ArticlePageDocument:
+            generator: article_page
+            options:
+                route_schema: '{translator.trans("page")}-{object.getPageNumber()}'
+                parent: '{object.getParent().getRoutePath()}'
```
alexander-schranz, line 16: here should also be the path config for ongr:

see
https://github.com/sulu/SuluArticleBundle/blob/master/Resources/doc/installation.md

luca-rath: > FYI: A recipe should not modify other bundles config.

Probably we have to create an own recipe for `handcraftedinthealps/elasticsearch-bundle`. Or better prepending ongr config in the `ArticleExtension`. WDYT?

alexander-schranz: No it should just inside the `sulu_article.yaml` as it is ongr config for the sulu article bundle.

fabpot:

fabpot:
Validation is ok now.

luca-rath:
Thanks a lot :)

sulu/article-bundle/2.0/config/packages/sulu_article.yaml
alexander-schranz: Can we add the `ongr_elasticsearch:` configuration atleast as commented out in this file so the dev sees what he need to activate for the bundle to work.

sulu/article-bundle/2.0/config/packages/sulu_article.yaml
```diff
@@ -0,0 +1,16 @@
+sulu_article:
+    index_name: '%env(resolve:ELASTICSEARCH_INDEX)%'
+    hosts:
+        - '%env(resolve:ELASTICSEARCH_HOST)%'
```
alexander-schranz, line 4: currently the bundle does not work without configuring a type mapping we should add here:

```
sulu_article:
    types:
        article:
            translation_key: "app.article"
```

sulu/article-bundle/2.0/manifest.json
```diff
@@ -0,0 +1,16 @@
+{
+    "bundles": {
+        "Sulu\\Bundle\\ArticleBundle\\SuluArticleBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "copy-from-package": {
+        "Resources/doc/default.xml": "%CONFIG_DIR%/templates/articles/default.xml",
+        "Resources/doc/default.html.twig": "templates/articles/default.xml"
```
alexander-schranz, line 10: this looks not correct

luca-rath:
ping @alexander-schranz can you review this again?

alexander-schranz (approved):

luca-rath:
> FYI: A recipe should not modify other bundles config.

@Nyholm Unfortunately, the `SuluArticleBundle` doesn't work without the configuration for the `ONGRElasticsearchBundle`. Would the current solution be okay for you?

sulu/article-bundle/2.0/config/packages/sulu_article.yaml
```diff
@@ -0,0 +1,28 @@
...
+        Sulu\Bundle\ArticleBundle\Document\ArticlePageDocument:
+            generator: article_page
+            options:
+                route_schema: '{translator.trans("page")}-{object.getPageNumber()}'
+                parent: '{object.getParent().getRoutePath()}'
+
+ongr_elasticsearch:
+    analysis:
+        tokenizer:
+            pathTokenizer:
```
Nyholm, line 24: The `tokenizer` and `analyzer` key expects a collection of different config. You make up the name `pathTokenizer` and `pathAnalyzer` specifically for this bundle, right? 

If so, this is fine.

https://github.com/ongr-io/ElasticsearchBundle/blob/master/DependencyInjection/Configuration.php#L105

Nyholm (commented):
I just have one question. 

I think this is fine.

sulu/article-bundle/2.0/config/packages/sulu_article.yaml
luca-rath: Yes, we made up the name of the tokenizer and the analyzer by ourselves, it just needs to be the same as configured in the [ArticleViewDocument](https://github.com/sulu/SuluArticleBundle/blob/master/Document/ArticleViewDocument.php#L68). Though we can't change those right now, because it would be a bc break.

Nyholm: Cool cool cool. 

Thank you

Nyholm (approved):

sulu/article-bundle/2.0/config/packages/sulu_article.yaml
luca-rath: thank you 🙂

PR #988, opened 2020-07-27
Add jsor/doctrine-postgis

Saphyel:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm (approved):
Thank you for this PR. 

I dont think it makes sense that your library has a dependency on DoctrineBundle but it makes sense to add this config for this recipe.

PR #995, opened 2020-08-28
Add flex recipe for SuluFormBundle

luca-rath:
| Q             | A
| ------------- | ---
| License       | MIT

Link to package: https://github.com/sulu/SuluFormBundle

### DIFF
--- a/sulu/form-bundle/2.0/config/packages/sulu_form.yaml
+++ b/sulu/form-bundle/2.0/config/packages/sulu_form.yaml
@@ -1,0 +1,9 @@
+sulu_form:
+    mail:
+        from: "%sulu_admin.email%"
+        to: "%sulu_admin.email%"
+        sender: "%sulu_admin.email%"
+
+framework:
+    esi: true
+    fragments: true
--- a/sulu/form-bundle/2.0/config/packages/sulu_form.yaml
+++ b/sulu/form-bundle/2.0/config/packages/sulu_form.yaml
@@ -1,8 +1,8 @@
 sulu_form:
     mail:
-        from: "%sulu_admin.email%"
-        to: "%sulu_admin.email%"
-        sender: "%sulu_admin.email%"
+        from: "%env(resolve:SULU_FORM_EMAIL_FROM)%"
+        to: "%env(resolve:SULU_FORM_EMAIL_TO)%"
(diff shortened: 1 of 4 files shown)


luca-rath:
ping @alexander-schranz

sulu/form-bundle/2.0/config/packages/sulu_form.yaml
```diff
@@ -0,0 +1,9 @@
+sulu_form:
+    mail:
+        from: "%env(SULU_ADMIN_EMAIL)%"
+        to: "%env(SULU_ADMIN_EMAIL)%"
+        sender: "%env(SULU_ADMIN_EMAIL)%"
+
+framework:
+    esi: true
```
alexander-schranz, line 8: Not happy with this but currently needed. Did create an issue for it on our side: https://github.com/sulu/SuluFormBundle/issues/265

alexander-schranz (approved):

sulu/form-bundle/2.0/config/packages/sulu_form.yaml
luca-rath: 👍

alexander-schranz:
@nicolas-grekas the CI is currently failing because in sulu the project need to define which `phpcr/phpcr-implementation` it will use. Not sure if there is any way to detect that. Maybe by checking the "composer.json" of the bundle/library has `jackalope/jackalope-doctrine-dbal` or `jackalope/jackalope-jackrabbit` in its "require-dev" like [here](https://github.com/sulu/SuluFormBundle/blob/fe518a1929099a986a2d19413639b77bcd58b5f2/composer.json#L40)  and then run e.g.:

```bash
composer require jackalope/jackalope-doctrine-dbal --no-update
```

Before running composer install script.

Nyholm:
It is really tricky to find virtual packages. No worries about the CI failures.

sulu/form-bundle/2.0/config/packages/sulu_form.yaml
Nyholm: I agree. Im not happy with this either. 

Thank you for the issue. We should make sure these are removed when https://github.com/sulu/SuluFormBundle/issues/265 is fixed.

Nyholm (approved):
Thank you

PR #996, opened 2020-08-28
Add Klaviyo SDK recipe

Th3Mouk:
| Q             | A
| ------------- | ---
| License       | MIT

A pretty simple recipe that will simplify the package installation

### DIFF
--- a/klaviyo/php-sdk/2.0/config/packages/klaviyo.yaml
+++ b/klaviyo/php-sdk/2.0/config/packages/klaviyo.yaml
@@ -1,0 +1,5 @@
+services:
+    Klaviyo\Klaviyo:
+        arguments:
+            $private_key: '%env(KLAVIYO_PRIVATE_KEY)%'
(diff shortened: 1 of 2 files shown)


klaviyo/php-sdk/2.0/config/packages/klaviyo.yaml
```diff
@@ -0,0 +1,6 @@
+services:
+    Klaviyo\Klaviyo:
+        class: Klaviyo\Klaviyo
```
Nyholm, line 3: This row is not needed when you have the service ID as: "Klaviyo\Klaviyo"

Nyholm (changes requested):
Thank you. Just a small thing

klaviyo/php-sdk/2.0/config/packages/klaviyo.yaml
Th3Mouk: I didn't know, thanks for the tip.

Nyholm (approved):
Thank you

PR #1002, opened 2020-09-03
* Add a recipe for coka/pagination-bundle version 5.0.

CedrickOka:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/coka/pagination-bundle/5.0/config/packages/oka_pagination.yaml
+++ b/coka/pagination-bundle/5.0/config/packages/oka_pagination.yaml
@@ -1,0 +1,4 @@
+oka_pagination:
+    db_driver: orm
+    twig:
+        enabled: false
--- a/coka/pagination-bundle/5.0/manifest.json
+++ b/coka/pagination-bundle/5.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Oka\\PaginationBundle\\OkaPaginationBundle": ["all"]
+    },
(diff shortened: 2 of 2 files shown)


coka/pagination-bundle/5.0/config/packages/oka_pagination.yaml
```diff
@@ -0,0 +1,4 @@
+oka_pagination:
+    db_driver: orm
```
Nyholm, line 2: This looks like the same as the default config. 

If so, flex is smart enough to install this bundle without a recipe.

Nyholm (changes requested):

coka/pagination-bundle/5.0/config/packages/oka_pagination.yaml
CedrickOka: Hi,

No this is not the default configuration. The `oka_pagination.db_driver` configuration must have a supplied value which must be either `orm` or `mongodb` but has no default value. If this value is not configured an exception should be thrown.

Nyholm: Okey. Thank you. 

I think this is a good recipe then =)

Nyholm (approved):
FYI: 

You should consider support SF 5.0. The upgrade path for this bundle is basically just to bump composer.json if I read things correctly.

Nyholm:
Thank you for this PR!

PR #1011, opened 2020-09-25
Add enhavo/app-bundle recipe

gseidel:
| Q             | A
| ------------- | ---
| License       | MIT

Adding a recipe for the enhavo/app-bundle

### DIFF
--- a/enhavo/app-bundle/0.9/assets/services/enhavo/enhavo_app.yaml
+++ b/enhavo/app-bundle/0.9/assets/services/enhavo/enhavo_app.yaml
@@ -1,0 +1,4 @@
+imports:
+    -
+        path: '@enhavo/app/services/*'
+
--- a/enhavo/app-bundle/0.9/manifest.json
+++ b/enhavo/app-bundle/0.9/manifest.json
@@ -7,9 +7,8 @@
     },
     "env": {
         "MAILER_FROM": "no-reply@email.com",
-        "MAILER_TO": "my@email.com",
         "MAILER_NAME": "enhavo",
-        "MAILER_DELIVERY_ADDRESS": "my@email.com"
+        "MAILER_TO": "my@email.com"
     },
     "copy-from-recipe": {
         "config/": "%CONFIG_DIR%/",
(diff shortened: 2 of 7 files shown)


gseidel:
@Nyholm Maybe you can make a review, because you also reviewed my PR before #915

enhavo/app-bundle/0.9/assets/services/enhavo/enhavo_app.yaml
```diff
@@ -0,0 +1,4 @@
+imports:
+    -
+        path: '@enhavo/app/services/*'
```
Nyholm, line 3: You dont need this for a bundle. The bundle will register their own services. See https://github.com/enhavo/EnhavoAppBundle/blob/v0.9.2/DependencyInjection/EnhavoAppExtension.php#L46-L61

enhavo/app-bundle/0.9/manifest.json
```diff
@@ -0,0 +1,25 @@
...
+        "Enhavo\\Bundle\\AppBundle\\EnhavoAppBundle": ["all"]
+    },
+    "container": {
+        "locale": "en"
+    },
+    "env": {
+        "MAILER_FROM": "no-reply@email.com",
+        "MAILER_TO": "my@email.com",
+        "MAILER_NAME": "enhavo",
+        "MAILER_DELIVERY_ADDRESS": "my@email.com"
```
Nyholm, line 12: These variables are never used. You should make sure to add those to `enhavo_app.yaml`

Nyholm (changes requested):
Thank you. It looks good. Just some minor things

enhavo/app-bundle/0.9/assets/services/enhavo/enhavo_app.yaml
gseidel: This isn't a configuration file for `Symfony`. It looks similar, but it is used for a webpack plugin within enhavo. (This is how I avoid using Typescript files in the recipe)

https://github.com/enhavo/dependency-injection#import

Nyholm: Ah, I see. Okey.

Nyholm (approved):
Thank you

PR #1015, opened 2020-10-02
added recipe for dneustadt/csrf-cookie-bundle

dneustadt:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/dneustadt/DneustadtCsrfCookieBundle


Nyholm (changes requested):
We don't need all this config in comments. Rather add a link to your documentation. 

In fact.. the bundle will work exactly the same with or without recipe.

PR #1028, opened 2020-11-01
Add recipe for ferror/phpscan repository

Ferror:
| Q                    | A
| ------------- | ---
| License          | MIT

Sorry for some "empty" commits - I had to trigger a flex-server bot because the re-review option got my 404. Maybe it's some kind of bug, but it could be great if we could re-request a review via the trigger.

Screenshot 2020-11-01 at 17 17 32


Nyholm (approved):
Thank you

PR #1043, opened 2020-11-19
add recipe for sgianushka/generic-bundle and sgianushka/region-bundle

siganushka:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/siganushka/region-bundle/0.1/config/routes/siganushka_region.yaml
+++ b/siganushka/region-bundle/0.1/config/routes/siganushka_region.yaml
@@ -1,0 +1,2 @@
+siganushka_region:
+    resource: "@SiganushkaRegionBundle/Resources/config/routing/routes.xml"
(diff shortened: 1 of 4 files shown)


siganushka/region-bundle/0.1/config/routes/siganushka_region.yaml
```diff
@@ -0,0 +1,2 @@
+siganushka_region:
+    resource: "@SiganushkaRegionBundle/Resources/config/routing/routes.xml"
```
Nyholm, line 2: This path does not exist.

Nyholm (changes requested):

siganushka/region-bundle/0.1/config/routes/siganushka_region.yaml
siganushka: My test has passed, can you provide details?

Nyholm (approved):
Thank you

PR #1056, opened 2020-12-06
Adding logging capabilities to GitWrapper

chr-hertel:
| Q             | A
| ------------- | ---
| License       | MIT

@OskarStark WDYT? This is what i was using before to wire the GitWrapper

edit: arg, requires monolog ... so only a recommendation as comment?

### DIFF
--- a/cpliakas/git-wrapper/3.0/config/packages/cpliakas_git_wrapper.yaml
+++ b/cpliakas/git-wrapper/3.0/config/packages/cpliakas_git_wrapper.yaml
@@ -1,15 +1,15 @@
 services:
     GitWrapper\GitWrapper: null
 
-# To add logging capabilities use monolog and this configuration
-# monolog:
+
+# To add logging capabilities use monolog and this configuration:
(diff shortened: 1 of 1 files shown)


OskarStark:
Uhh yes very nice 👌🏻

OskarStark (approved):

cpliakas/git-wrapper/3.0/config/packages/cpliakas_git_wrapper.yaml
```diff
@@ -1,2 +1,15 @@
 services:
     GitWrapper\GitWrapper: null
+
+# To add logging capabilities use monolog and this configuration
+# monolog:
```
fabpot, line 5: I would remove the space between # and the content so that people can just remove the comment and it works (probably even more important for the other lines to have the right number of spaces).

fabpot (changes requested):

fabpot (approved):

PR #1065, opened 2020-12-13
Add recipe for Sylius Settings Plugin by @monsieurbiz

jacquesbh:
| Q             | A
| ------------- | ---
| License       | MIT

https://github.com/monsieurbiz/SyliusSettingsPlugin

### DIFF
--- a/monsieurbiz/sylius-settings-plugin/1.0/config/routes/monsieurbiz_settings_plugin.yaml
+++ b/monsieurbiz/sylius-settings-plugin/1.0/config/routes/monsieurbiz_settings_plugin.yaml
@@ -1,0 +1,3 @@
+monsieurbiz_sylius_settings_admin:
+    resource: "@MonsieurBizSyliusSettingsPlugin/Resources/config/routes/admin.yaml"
+    prefix: /%sylius_admin.path_name%
--- a/monsieurbiz/sylius-settings-plugin/1.0/config/packages/monsieurbiz_settings_plugin.yaml
+++ b/monsieurbiz/sylius-settings-plugin/1.0/config/packages/monsieurbiz_settings_plugin.yaml
@@ -1,0 +1,2 @@
+imports:
+    - { resource: "@MonsieurBizSyliusSettingsPlugin/Resources/config/config.yaml" }
--- a/monsieurbiz/sylius-settings-plugin/1.0/manifest.json
+++ b/monsieurbiz/sylius-settings-plugin/1.0/manifest.json
@@ -1,0 +1,10 @@
+{
+    "bundles": {
+        "MonsieurBiz\\SyliusSettingsPlugin\\MonsieurBizSyliusSettingsPlugin": [
+            "all"
+        ]
+    },
+    "copy-from-recipe": {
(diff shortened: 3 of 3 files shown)


monsieurbiz/sylius-settings-plugin/1.0/config/routes/monsieurbiz_settings_plugin.yaml
```diff
@@ -0,0 +1,3 @@
+monsieurbiz_sylius_settings_admin:
+    resource: "@MonsieurBizSyliusSettingsPlugin/Resources/config/routes/admin.yaml"
+    prefix: /%sylius_admin.path_name%
```
Nyholm, line 3: Will this really work? 
Excuse my ignorance, but I was sure you could not add parameters like this in the routes.

Nyholm (commented):
Just a small question to confirm

monsieurbiz/sylius-settings-plugin/1.0/config/routes/monsieurbiz_settings_plugin.yaml
jacquesbh: It works pretty well indeed. We have the same in production.

The parameter is defined by Sylius itself here:

https://github.com/Sylius/Sylius/blob/f27be445c11eccd0556d5edbafd60067960ca46c/src/Sylius/Bundle/AdminBundle/Resources/config/app/config.yml#L38

This is actually very convenient to secure the admin path with a random name!

jacquesbh: @Nyholm What do you think?

jacquesbh: ping @Nyholm ? :p

Nyholm:

Nyholm:
Could you tell me why Travis is red?

jacquesbh:
> Could you tell me why Travis is red?

Hey @Nyholm!

I'm pretty sure it has to do with the installation process.
As a Sylius plugin you install it on top of a Sylius installation, which requires the  sylius/ui-bundle.
The ui bundle adds it's own configuration `sylius_ui`: https://github.com/Sylius/SyliusUiBundle/blob/b34241bb76415faff996b27dc123077b8389778d/DependencyInjection/Configuration.php#L24

We check our recipe on a Sylius standard edition here: https://github.com/monsieurbiz/SyliusSettingsPlugin/runs/2415461391?check_suite_focus=true
You can get the details in the step "Install plugin".

My understanding is that there is no auto configuration for the sylius/ui-bundle so it's not declared in the `bundles.php`, therefor there is no way a default symfony app would work with our declaration of some new blocks.

Maybe we could require the sylius/ui-bundle, that could solve the issue?
But it's inside the sylius/sylius… https://github.com/Sylius/Sylius/blob/c68ccd7fc7bda5f4dfac279f15bf669b23c37100/composer.json#L189

Nyholm (approved):
Thank you

PR #1069, opened 2020-12-18
added systemsdk/easy-log-bundle

dimadeush:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/systemsdk/easy-log-bundle/1.10/config/packages/dev/systemsdk_easy_log.yaml
+++ b/systemsdk/easy-log-bundle/1.10/config/packages/dev/systemsdk_easy_log.yaml
@@ -1,0 +1,3 @@
+easy_log:
+    log_path: '%kernel.logs_dir%/%kernel.environment%-readable.log'
+    ignored_routes: ['_wdt', '_profiler']
--- a/systemsdk/easy-log-bundle/1.10/config/packages/test/systemsdk_easy_log.yaml
+++ b/systemsdk/easy-log-bundle/1.10/config/packages/test/systemsdk_easy_log.yaml
@@ -1,0 +1,3 @@
+easy_log:
+    log_path: '%kernel.logs_dir%/%kernel.environment%-readable.log'
+    ignored_routes: ['_wdt', '_profiler']
--- a/systemsdk/easy-log-bundle/1.10/manifest.json
+++ b/systemsdk/easy-log-bundle/1.10/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "Systemsdk\\Bundle\\EasyLogBundle\\EasyLogBundle": ["dev", "test"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


systemsdk/easy-log-bundle/1.10/config/packages/dev/easy_log.yaml
```diff
@@ -0,0 +1,7 @@
+easy_log:
+    log_path: '%kernel.logs_dir%/%kernel.environment%-readable.log'
+    # Use below if you're running application in container and want to log in stdout
+    #log_path: 'php://stdout'
+    max_line_length: 120
+    prefix_length: 2
+    ignored_routes: ['_wdt', '_profiler']
```
Nyholm, line 7: You dont need to show all your config. Just keep sensible defaults.

systemsdk/easy-log-bundle/1.10/config/packages/test/easy_log.yaml
```diff
@@ -0,0 +1,7 @@
+easy_log:
```
Nyholm, line 1: Same here

Nyholm (commented):
Whenever a user install this package, they will get a big error message. (Like on travis)

Could we add the monolog handler config?

systemsdk/easy-log-bundle/1.10/config/packages/dev/easy_log.yaml
dimadeush: It is already sensible defaults (we usually don't need to log debug info)

systemsdk/easy-log-bundle/1.10/config/packages/test/easy_log.yaml
dimadeush: It is already sensible defaults (we usually don't need to log debug info)

dimadeush:
> Whenever a user install this package, they will get a big error message. (Like on travis)
> 
> Could we add the monolog handler config?

Yes, we can but there is some issue with it - Sometimes some projects can have edited default monolog config and it means that prepared monolog config here will not work as expected. 

So I think here I have two cases: 

1)delete checking monolog config in EasyLogExtension (row 40 https://github.com/systemsdk/easy-log-bundle/blob/master/src/DependencyInjection/EasyLogExtension.php) and that's exactly same error what you can see in travis CI here  

2)can try to add monolog config but in such case there can be some issues in case project have custom monolog config.

Do you have better ideas?

dimadeush:
@fabpot @Nyholm  Selected case 1 mentioned above. Now CI passed, please review once again.

systemsdk/easy-log-bundle/1.10/config/packages/dev/easy_log.yaml
Nyholm: I suggest you keep: 

```yaml
easy_log:
    log_path: '%kernel.logs_dir%/%kernel.environment%-readable.log'
    ignored_routes: ['_wdt', '_profiler']
 ```

The rest can be removed. There is no need to show all your config. 

Feel free to add a link to your documentation if you want to.

**Review by Nyholm — CHANGES_REQUESTED** [review]
Thank you. I just want you to address my comments and I'll be happy to merge.

**Comment by dimadeush** [comment]
> Thank you. I just want you to address my comments and I'll be happy to merge.

Hello, done. Thank you in advance.

File: systemsdk/easy-log-bundle/1.10/config/packages/dev/easy_log.yaml [hunk 3]
```diff
@@ -0,0 +1,5 @@
+easy_log:
+    log_path: '%kernel.logs_dir%/%kernel.environment%-readable.log'
+    # Use below if you're running application in container and want to log in stdout
+    #log_path: 'php://stdout'
```
Nyholm, line 4: I suggest you put this comment in your bundles documentation. The recipe is not a replacement for your documentation. 

Feel free to add a link to your docs.

Nyholm (changes requested):

systemsdk/easy-log-bundle/1.10/config/packages/dev/easy_log.yaml
dimadeush: Ok, done.

Nyholm (commented):
Im happy with this. 
Just one more thing. Could you rename the files to `systemsdk_easy_log.yaml`?

dimadeush:
> Im happy with this.
> Just one more thing. Could you rename the files to `systemsdk_easy_log.yaml`?

Done

Nyholm (approved):
Thank you

PR #1077, opened 2021-01-12
Adding iungo-web/img-compressor-cli

mateusfmello:
| Q             | A
| ------------- | ---
| License       | MIT


iungo-web/img-compressor-cli/0.1/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+	"copy-from-package": {
+		"bin/compressor": ""
```
Nyholm, line 3: Shouldn't this be handled by composer? See https://getcomposer.org/doc/articles/vendor-binaries.md

iungo-web/img-compressor-cli/0.1/manifest.json
```diff
@@ -0,0 +1,13 @@
...
+	},
+	"env": {
+		"API_KEY": "123ABC1",
+		"DIRETORIO_RAIZ": "seu/diretorio/aqui/"
+	},
+	"gitignore": [
+		".env.local",
+		"/vendor/"
+	]
```
Nyholm, line 12: This is not needed

iungo-web/img-compressor-cli/0.1/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+	"copy-from-package": {
+		"bin/compressor": ""
+	},
+	"env": {
+		"API_KEY": "123ABC1",
+		"DIRETORIO_RAIZ": "seu/diretorio/aqui/"
+	},
```
Nyholm, line 8: I dont see that your recipe is using these.

Nyholm (changes requested):
Thank you for the PR.

iungo-web/img-compressor-cli/0.1/manifest.json
mateusfmello: The recipe does not use it, it will only create these lines in the .env file for the compression system to use. 

[https://github.com/iungoweb/img-compressor-cli/blob/main/src/Compressor.php#L49](https://github.com/iungoweb/img-compressor-cli/blob/main/src/Compressor.php#L49)

iungo-web/img-compressor-cli/0.1/manifest.json
mateusfmello: I also thought about doing it this way, but the point is that this system is something to use alone through the command line, so I thought it was more appropriate to put the executable file at the root of the project.

mateusfmello:
@Nyholm Hello how are you? Is there something missing for the PR to be merged?

iungo-web/img-compressor-cli/0.1/manifest.json
Nyholm: But if the user explicitly says that they want all the binaries installed by composer in a specific binary directory, I think that should be respected.

iungo-web/img-compressor-cli/0.1/manifest.json
Nyholm: Sure, lets keep them.

Nyholm (changes requested):
Im fine thank you. 

Im happy with adding a recipe with just environment variable for this package.

mateusfmello:
> Im fine thank you.
> 
> Im happy with adding a recipe with just environment variable for this package.

Okay I will remove the copy statement from the binary, I will review this in the future, I am new to this issue of creating recipes for Symfony.

Any suggestions for content about creating recipes for Symfony?

Nyholm:
Friendly ping to finish this. 

You can read more about recipes here: https://github.com/symfony/recipes

Nyholm:
Another ping

Issue #1076, opened 2021-01-11
Bug in automatic merging process?


ThomasLandauer:
https://github.com/symfony/recipes-contrib/blob/master/README.rst says that a PR gets automatically merged if it's approved by:

> The user is a member of the organizations (publicly visible) for all modified packages.

Isn't this the case at https://github.com/symfony/recipes-contrib/pull/1075#pullrequestreview-565419301 ?

fabpot:
This has been desactivated for a while now.

PR #1078, opened 2021-01-12
Removing outdated information on automatic merge process

ThomasLandauer:
See https://github.com/symfony/recipes-contrib/issues/1076#issuecomment-758502196

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,22 @@ For more information about contributing a recipe, read the `documentation`_ on
 the main repository.
 
 Unlike for official recipes, the pull requests for new contrib recipes are
-managed by the community.
+managed by the community. Pull requests are reviewed by the Symfony Bot and
+automatically merged when the following conditions are met (in that order):
+
+* The pull request title does not contain "WIP";
+
+* The Symfony Bot approved the pull request and reported no validation errors;
+
+* Someone (not the Symfony bot nor the pull request author) approved the pull
+  request;
+
+* The pull request author or the reviewer is a "trusted user". A "trusted user"
+  being one of the following:
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -19,25 +19,7 @@ For more information about contributing a recipe, read the `documentation`_ on
...
-
-* The pull request author or the reviewer is a "trusted user". A "trusted user"
-  being one of the following:
-
-  * A Symfony Core Merger;
-
-  * The user is the Github owner for all modified packages;
-
-  * The user is a member of the organizations (publicly visible) for all
-    modified packages.
```
fabpot, line 40: Actually, everything is still true except for the last item which should read: "The pull request author or the reviewer is a Symfony Core Merger"

fabpot (changes requested):

README.rst
ThomasLandauer: This is already the first point ;-) So in fact, the third needs to be deleted - that's what I did now.

fabpot: Why keep a list with one item?

ThomasLandauer: *Two* items (GitHub's display here is a little confusing):
  * a Symfony Core Merger;
  * the Github owner for all modified packages.

fabpot: The second one is not true anymore; that's why I proposed "The pull request author or the reviewer is a Symfony Core Merger"

ThomasLandauer: Sorry, didn't understand what you meant by "last item" ;-) Fixed now - hopefully :-)

fabpot (approved):

PR #1079, opened 2021-01-22
Add recipe for tilleuls/url-signer-bundle

alanpoulain:
| Q             | A
| ------------- | ---
| License       | MIT

Bundle can be found here: https://github.com/coopTilleuls/UrlSignerBundle

### DIFF
--- a/tilleuls/url-signer-bundle/1.0/config/packages/url_signer.yaml
+++ b/tilleuls/url-signer-bundle/1.0/config/packages/url_signer.yaml
@@ -1,0 +1,2 @@
+coop_tilleuls_url_signer:
+    signature_key: '%env(string:SIGNATURE_KEY)%'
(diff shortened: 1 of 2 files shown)


tilleuls/url-signer-bundle/1.0/config/packages/url_signer.yaml
```diff
@@ -0,0 +1,2 @@
+coop_tilleuls_url_signer:
+    signature_key: 'change_me'
```
dunglas, line 2: You should maybe define an environment variable?

dunglas (changes requested):

tilleuls/url-signer-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
```
dunglas, line 2: You should add the env var to the manifest too. So Flex will automatically add it to `.env`.

dunglas (changes requested):

dunglas (approved):

PR #1083, opened 2021-01-29
Add recipe for artox-lab/clarc-message-bus-bundle

Svarok:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm (approved):
Thank you. 

It seams like you've added some default config. That is generally not ideal, but I believe that this config will most likely be changed by 50% of the users. 

Im happy with this PR.

PR #1085, opened 2021-02-01
Update storage directory for PayumBundle

pierredup:
| Q             | A
| ------------- | ---
| License       | MIT

The current recipe for [PayumBundle](https://github.com/Payum/PayumBundle) uses the old directory structure of Symfony < 4. This update changes the location of the storage to use `kernel.project_dir` instead of `kernel.root_dir`, and also uses the `var` folder for storage instead of `app/Resources`.
This ensures compatibility with all the latest versions of Symfony.


Nyholm (approved):
Thank you

PR #1087, opened 2021-02-06
Add recipe for Annotation Cache Bundle

danieltoader:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/emag-tech-labs/annotation-cache-bundle/5.0/config/packages/emag_annotation_cache.yaml
+++ b/emag-tech-labs/annotation-cache-bundle/5.0/config/packages/emag_annotation_cache.yaml
@@ -1,0 +1,4 @@
+annotation_cache:
+    provider: cache.app
+    ignore_namespaces:
+        - 'App\Entity'
--- a/emag-tech-labs/annotation-cache-bundle/5.0/manifest.json
+++ b/emag-tech-labs/annotation-cache-bundle/5.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "EmagTechLabs\\AnnotationCacheBundle\\AnnotationCacheBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


danieltoader:
@Nyholm Hello, I don't know what the procedure is to get this PR merged, could you please help or point me in the right direction?
Thank you!

emag-tech-labs/annotation-cache-bundle/5.0/config/packages/emag_annotation_cache.yaml
```diff
@@ -0,0 +1,3 @@
+annotation_cache:
+    provider: {}
```
Nyholm, line 2: I see that this bundle does not work without a provider. Let's add `cache.app` here as a default.

emag-tech-labs/annotation-cache-bundle/5.0/config/packages/emag_annotation_cache.yaml
```diff
@@ -0,0 +1,3 @@
...
+    ignore_namespaces: []
```
Nyholm, line 3: We can remove this. Let's add this to the docs instead. 

Alternatively, add `App\Entity` here. They are already cached by doctrine.

Nyholm (changes requested):
Thank you for the PR. 

Unrelated: Have you done a test with Blackfire to make sure this is actually an improvement? 
Also, I made a similar bundle a few years ago. See https://github.com/Happyr/annotation-warmer

emag-tech-labs/annotation-cache-bundle/5.0/config/packages/emag_annotation_cache.yaml
danieltoader: Added 'App\Entity'

emag-tech-labs/annotation-cache-bundle/5.0/config/packages/emag_annotation_cache.yaml
danieltoader: Added 'cache.app' as the default value

danieltoader:
@Nyholm After reading your comment I realized that the description was not that accurate so I updated the readme.md with more details about the Bundle and how it can be used.
Basically this package can be used to add cache to any method with the `@Cache` annotation. You can use any implementation that uses the PSR-6 standard as a provider and it will store the response of the method there.

This is helpful because you do not need to write the classic: search in cache, if found return, if not found run function code, store the result in cache and then return the value for each method you need to use cache for. That will all be done by the bundle which generates a proxy class that extends the original class and overwrites the methods with the `@Cache` annotation and adds the logic from above.

Nyholm:

Nyholm (approved):
Okey. That is a bit different. 

Thank you

PR #1091, opened 2021-02-11
Add new recipe ecphp/api-gw-authentication-bundle

drupol:
| Q             | A
| ------------- | ---
| License       | MIT

The bundle is for Symfony >= 5. See the repo: https://github.com/ecphp/api-gw-authentication-bundle

### DIFF
--- a/ecphp/api-gw-authentication-bundle/1.0/manifest.json
+++ b/ecphp/api-gw-authentication-bundle/1.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "EcPhp\\ApiGwAuthenticationBundle\\ApiGwAuthenticationBundle": ["all"]
+    },
+    "copy-from-package": {
+        "Resources/config/": "%CONFIG_DIR%/"
+    }
+}
--- a/ecphp/api-gw-authentication-bundle/1.0/manifest.json
+++ b/ecphp/api-gw-authentication-bundle/1.0/manifest.json
@@ -2,7 +2,7 @@
     "bundles": {
         "EcPhp\\ApiGwAuthenticationBundle\\ApiGwAuthenticationBundle": ["all"]
     },
-    "copy-from-package": {
-        "Resources/config/": "%CONFIG_DIR%/"
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
     }
 }
--- a/ecphp/api-gw-authentication-bundle/1.0/config/packages/dev/api_gw_authentication.yaml
+++ b/ecphp/api-gw-authentication-bundle/1.0/config/packages/dev/api_gw_authentication.yaml
@@ -1,0 +1,3 @@
+api_gw_authentication:
+    defaults:
+        env: acceptance # Available values are: acceptance, intra, production, user
--- a/ecphp/api-gw-authentication-bundle/1.0/config/routes/dev/api_gw_authentication.yaml
+++ b/ecphp/api-gw-authentication-bundle/1.0/config/routes/dev/api_gw_authentication.yaml
@@ -1,0 +1,3 @@
+api_gw_authentication_bundle:
+    resource: '@ApiGwAuthenticationBundle/Resources/config/routes/routes.php'
+    prefix: /api


sstok (approved):

ecphp/api-gw-authentication-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "EcPhp\\ApiGwAuthenticationBundle\\ApiGwAuthenticationBundle": ["all"]
+    },
+    "copy-from-package": {
+        "Resources/config/": "%CONFIG_DIR%/"
+    }
```
Nyholm, line 7: I dont think adding recipes like this is a good idea. 

See the best practices in the readme: https://github.com/symfony/recipes-contrib#best-practices

Nyholm (changes requested):

ecphp/api-gw-authentication-bundle/1.0/manifest.json
drupol: I understand you just wrote those new best-practices after seeing my PR (_thanks for sending me the link of [your PR](https://github.com/symfony/recipes-contrib/pull/1093) as kindly requested on Slack by the way_).

All the previous PR for my other ECPHP packages were accepted straight, by you: https://github.com/symfony/recipes-contrib/pulls?q=is%3Apr+is%3Aclosed+ecphp

Today [a new Best Practice paragraph](https://github.com/symfony/recipes-contrib/pull/1093), I read:

> The symfony/recipes-contrib repository should contain the config for packages. Using "copy-from-package Configurator" for routes and config is not allowed. That would make the recipes impossible to maintain and to assure their quality.

Now, if I provide the same PR with the config files that I automatically copy, what would it change?
Do you really think that just by reading those files you can correctly or better assert/maintain/evaluate the quality of a bundle?
The idea of having such a recipe is to facilitate the installation of configuration files, nothing else, how could you assert the quality of a bundle just by reading those files?

Thanks.

Nyholm: > I understand you just wrote those new best-practices after seeing my PR

That is correct that I just wrote them. But they have been discussed since last year. 

Im sorry that I missed sending you the link. 

> All the previous PR for my other ECPHP packages were accepted straight, by you: https://github.com/symfony/recipes-contrib/pulls?q=is%3Apr+is%3Aclosed+ecphp

That is correct. Recipes like ECPHP was a part of the reason why I thought we needed some better guidance. 

> Now, if I provide the same PR with the config files that I automatically copy, what would it change?

Yes

> Do you really think that just by reading those files you can correctly or better assert/maintain/evaluate the quality of a bundle?

Yes

> The idea of having such a recipe is to facilitate the installation of configuration files, nothing else, how could you assert the quality of a bundle just by reading those files?

I can be sure that this recipe does not violate any of the other quality rules.

drupol: After discussing this matter with @nicolas-grekas, I decided to submit the request changes, even if I do not agree with this new best-practice rule.

ecphp/api-gw-authentication-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "EcPhp\\ApiGwAuthenticationBundle\\ApiGwAuthenticationBundle": ["all"]
```
Nyholm, line 3: I see that you enable the bundle for all environments but the config is only for `dev`, is this really correct?

drupol: Yes definitely correct.

The use should be able to use some routes only on the `dev` environment.

This is why it is like that.

ecphp/api-gw-authentication-bundle/1.0/manifest.json
Nyholm: Thank you

ecphp/api-gw-authentication-bundle/1.0/manifest.json
Nyholm: Thank you. Im fine with that. 

Im sure everything is correct. But could you help me understand why the routes is needed only in `dev` but the bundle needs to be active in prod without routes?

drupol: Sure I can explain.

The package is about JWT authentication. We have to authenticate requests coming from a corporate European Commission service. However, some developers doesn't have the possibility to generate tokens, it can only be generate by that service.

For example, myself. When I developed the bundle for my colleagues, I had no "easy" way to generate a real token for testing.
Also, my colleagues needs to generate fake tokens so they can play with their api and their tests.

This is why that routes exists on `dev` environment only. To be able to generate fake tokens and introspect them.

There is also a note about that in the [documentation](https://ecphp-api-gw-authentication-bundle.readthedocs.io/en/latest/pages/installation.html#step-2).

drupol: That said, now that I think of it.
As we obviously do not provide the secret key, even if those routes are enabled on `prod`, it would not be an issue.

Nyholm: Cool, then I understand. 

I assume that there are some commands too for generating tokens for tests. 

Im happy to merge this PR if you are ready.

drupol: Thanks, you can merge it safely :-)

Nyholm (approved):
Thank you for the PR.

drupol:
Thanks!

PR #1094, opened 2021-02-12
Add routmoute/routmoute-discord-bundle

routmoute:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm (changes requested):
I see that you added two recipes, one for 1.0 and one for 1.1. They are a bit different...

Do you expect any new installations to install 1.0? If not, I think we can remove that recipe.

routmoute:
Yes, we can :) ok.

Nyholm (approved):
Im happy with this. Can you rebase this to make sure the new CI jobs are happy?

Nyholm:
Thank you

PR #1099, opened 2021-02-23
Lorem-ipsum-bundle recipe test

BooleanType:
| Q             | A
| ------------- | ---
| License       | MIT


OskarStark:
No files changed 🧐

BooleanType:
> No files changed 🧐

Yeah, that's strange for me, 'cause I pushed them hour ago. Now re-pushed.

BooleanType:
Also I don't get, why I still see "There are some validation errors", when click on "View deployment" link. I've fixed everything and all checks have passed.

image

boolean-type/lorem-ipsum-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,12 @@
+{
+    "bundles": {
+        "KnpU\\LoremIpsumBundle\\KnpULoremIpsumBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "KNPU_LOREM_IPSUM_FOO": "hello",
+        "KNPU_LOREM_IPSUM_BAR": "world"
```
Nyholm, line 10: What are these environment variables for?

BooleanType: It was only for test: I just wanted to see, if variables are set after bundle installation.
I will handle it soon if needed.

BooleanType: Ok, I've decided not to change bundle code to use those variables. There is possible for user to use them instead. For example, in subscriber, which allows to change data returned by API:

<pre>
public static function getSubscribedEvents()
{
    return [
        KnpULoremIpsumEvents::FILTER_API => 'onFilterApi',
    ];
}

public function onFilterApi(FilterApiResponseEvent $event)
{
    $data = $event->getData();
    $data['message'] = 'Have a magical day!';
    $data['hello_world'] = $_ENV['KNPU_LOREM_IPSUM_FOO'] . ', ' . $_ENV['KNPU_LOREM_IPSUM_BAR'];
    $event->setData($data);
}
</pre>

Nyholm:
This recipe just add default values (and not even the one promoted by the bundle). 

Flex is smart enough to install this bundle without a recipe.

PR #1105, opened 2021-03-12
[CentrifugoBundle] Move config to packages directory

fre5h:
| Q             | A
| ------------- | ---
| License       | MIT

Found that config file is in the root of `config` directory instead of `config\packages`


Nyholm (approved):
Oops. Thank you

PR #1118, opened 2021-04-12
Update FluxSESyliusPayumStripePlugin

Prometee:
| Q             | A
| ------------- | ---
| License       | MIT

Update the recipe to handle the future 2.0 version.

### DIFF
--- a/flux-se/sylius-payum-stripe-plugin/2.0/config/packages/flux_se_sylius_payum_stripe.yaml
+++ b/flux-se/sylius-payum-stripe-plugin/2.0/config/packages/flux_se_sylius_payum_stripe.yaml
@@ -1,0 +1,11 @@
+imports:
+    - { resource: "@FluxSESyliusPayumStripePlugin/Resources/config/config.yaml" }
+
+flux_se_sylius_payum_stripe:
+    line_item_image:
+        # This is the imagine filter use to get the image displayed on Stripe Checkout Session
+        # (default: the filter uses into `@ShopBundle/Product/Show/_images.html.twig`)
+        imagine_filter: 'sylius_shop_product_original'
+        # Fallback image used when no image is set on a product and also when you test this plugin
+        # from a private web server (ex: from localhost)
+        fallback_image: 'https://placehold.it/400x300'
--- a/flux-se/sylius-payum-stripe-plugin/2.0/manifest.json
+++ b/flux-se/sylius-payum-stripe-plugin/2.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "FluxSE\\SyliusPayumStripePlugin\\FluxSESyliusPayumStripePlugin": ["all"]
+    },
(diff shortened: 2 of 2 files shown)


Prometee:
@Nyholm I know you are not a fan of config imports but we are forced to configure it like this :

- We can make a `CompilerPass` but this will add complexity for the developers wanting to customize our plugins
- We can add the config directly to the right `config/packages/*.yaml` file but in some casses we need to overwrite what Sylius as already configured altering for example the `config/packages/_sylius.yaml` file. Also our Sylius plugin require this config and in almost all cases It should not be changed.

Do you see another way of doing what we need ?

flux-se/sylius-payum-stripe-plugin/2.0/config/packages/flux_se_sylius_payum_stripe.yaml
```diff
@@ -0,0 +1,11 @@
+imports:
+    - { resource: "@FluxSESyliusPayumStripePlugin/Resources/config/config.yaml" }
```
Nyholm, line 2: But you are using a bundle... Why don't you import this file yourself?

Just next to https://github.com/FLUX-SE/SyliusPayumStripePlugin/blob/master/src/DependencyInjection/FluxSESyliusPayumStripeExtension.php#L49

```php
$loader->load('config.yaml');
```

Nyholm (changes requested):

Nyholm:
> We can add the config directly to the right config/packages/*.yaml file but in some casses we need to overwrite what Sylius as already configured altering for example the config/packages/_sylius.yaml file. Also our Sylius plugin require this config and in almost all cases It should not be changed.

Oh, So it is because the bundle's config depend on eachother... That is a horrible situation to be in. hm..

Nyholm (approved):
I guess this is a design thing with Sylius itself. 

Let's be pragmatic. 

Thank you

flux-se/sylius-payum-stripe-plugin/2.0/config/packages/flux_se_sylius_payum_stripe.yaml
Prometee: @Nyholm I guess you already my comment : https://github.com/symfony/recipes-contrib/pull/1118#issuecomment-817681098

It's to give users of this Sylius plugin a chance to modify for example the `winzou_state_machine` at their convenience, because sometimes  it can not correspond to the processes he needs. `winzou_state_machine` has a big tree and it's very often hard to disable or simply add something to it, that's why most of the time Sylius plugin let this configuration open to modification for the final users. Regarding the `twig`  config it can be removed from the config file and added to a `CompilerPass`.

Prometee:
@Nyholm Thank you, yes I simply follow what Sylius is doing too, but from a dev perspective it's better to allow this to be able to bend  the config to our needs (only for Sylius projects).

PR #364, opened 2018-05-16
Added recipe for FOSOAuthServerBundle

patrickbussmann:
| Q             | A
| ------------- | ---
| License       | MIT

Hopefully this can be merged soon to support FOSOAuthServerBundle 🙂

PR #1119, opened 2021-04-12
Add recipe for FOSOAuthServerBundle

MatTheCat:
| Q             | A
| ------------- | ---
| License       | MIT

Follow-up of https://github.com/symfony/recipes-contrib/pull/364

Fixes https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/issues/501 and https://github.com/FriendsOfSymfony/FOSOAuthServerBundle/issues/638

It is my first recipe so I may have missed some things.

FTR entities properties are `protected` because their parent’s are.


MatTheCat:
If I understand the error message correctly I cannot create a recipe for this package as it has dependencies based on the value of `db_driver` which is mandatory.

Could someone confirm?

Nyholm:
> If I understand the error message correctly I cannot create a recipe for this package as it has dependencies based on the value of db_driver which is mandatory.

It is fine. It is just Travis that wont run.

Nyholm (approved):
Thank you

Issue #1127, opened 2021-04-24
Bref Symfony Bridge recipe


ker0x:
Hello

After some [discussion](https://github.com/brefphp/symfony-bridge/issues/35) with @mnapoli and @t-richard, we would like to propose a recipe for [brefphp/symfony-bridge](https://github.com/brefphp/symfony-bridge).

Currently the recipe will focus on doing 3 things:

- Adding the .serverless folder to .gitignore
- Creating a serverless.yml already configured for a simple Symfony application
- Updating src/Kernel.php with the necessary changes

However, before opening a PR we have some questions in suspend. 

The library is not a Symfony bundle, so we were wondering if a recipe would be accepted for it?

Next, to work the bridge has to replace the default `Kernel.php` created by Symfony with a new one extended it to [override the cache and logs directories](https://github.com/brefphp/symfony-bridge/blob/master/src/BrefKernel.php#L9). Is it something a recipe is allowed to do or is it only permitted by offcial recipes (or totally forbidden) ?

Finally, the recipe will create a `serverless.yml` configuration file at the root of the project. Does this file should be store on the library side or recipe side?

fabpot:
We accept recipes for any packages available on Packagist (we do have plenty on non-bundles already).

Replacing an existing file is not an option as it would make debugging/upgrading/... much harder. We should find another way. Maybe the new runtime component can help here?

Generating files is one of the things a recipe can do, so I don't any issue here.

mnapoli:
@fabpot thank you for the answer!

> Replacing an existing file is not an option as it would make debugging/upgrading/... much harder. We should find another way. Maybe the new runtime component can help here?

Nyholm is working hard on that, the runtime component could be useful in some ways, but AFAIK not directly for this problem.

On Lambda, logs & cache need to be moved elsewhere (because the filesystem is read-only except for `/tmp`). The only way to do that is by editing `Kernel.php` (unless there's another way?).

Worst case scenario: we ask the user to do the change manually, which isn't the end of the world. But I'd love to know if there's an alternative that could be automated.

OskarStark:
What about adding a ServerlessKernel.php file hey to the Kernel and tell the user to change one line in index.php ?

mnapoli:
@OskarStark that is the same amount of changes as telling the user to extend Bref's modified Kernel.

t-richard:
> What about adding a ServerlessKernel.php file next to the Kernel and tell the user to change one line in index.php ?

[...]

PR #1129, opened 2021-04-27
Add recipe for bref/symfony-bridge

ker0x:
| Q             | A
| ------------- | ---
| License       | MIT

Following #1127, this is the first draft of the recipe.

Ping @t-richard, @mnapoli

### DIFF
--- a/bref/symfony-bridge/0.1/manifest.json
+++ b/bref/symfony-bridge/0.1/manifest.json
@@ -1,6 +1,6 @@
 {
     "copy-from-recipe": {
-        "serverless.yaml": "serverless.yaml"
+        "serverless.yaml": "serverless.yml"
     },
     "gitignore": [
         "/.serverless/"
--- a/bref/symfony-bridge/0.1/manifest.json
+++ b/bref/symfony-bridge/0.1/manifest.json
@@ -1,0 +1,8 @@
+{
+    "copy-from-recipe": {
+        "serverless.yaml": "serverless.yaml"
+    },
+    "gitignore": [
+        "/.serverless/"
+    ]
+}
--- a/bref/symfony-bridge/0.1/post-install.txt
+++ b/bref/symfony-bridge/0.1/post-install.txt
@@ -1,0 +1,8 @@
+<bg=blue;fg=white>                     </>
+<bg=blue;fg=white> Bref Symfony Bridge </>
+<bg=blue;fg=white>                     </>
+
+  * <fg=blue>Read</> the full documentation to properly configure your
+    Symfony application at <comment>https://bref.sh/docs/frameworks/symfony.html</>
+
+  * Bref documentation: <comment>https://bref.sh/docs/</>
--- a/bref/symfony-bridge/0.1/serverless.yaml
+++ b/bref/symfony-bridge/0.1/serverless.yaml
@@ -1,0 +1,46 @@
+# Read the documentation at https://www.serverless.com/framework/docs/providers/aws/guide/serverless.yml/
+service: symfony
+
+provider:
+    name: aws
+    # The AWS region in which to deploy (us-east-1 is the default)
+    region: us-east-1
+    # The stage of the application, e.g. dev, production, staging… ('dev' is the default)
+    stage: dev
+    runtime: provided.al2
+    environment:
+        # Symfony environment variables
+        APP_ENV: prod
+
+plugins:
+    - ./vendor/bref/bref
+
+functions:
+    # This function run the Symfony application
+    website:
+        handler: public/index.php
+        timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
+        layers:
+            - ${bref:layer.php-80-fpm}
+        events:
+            - httpApi: '*'
+    # This function lets us run console commands in Lambda
+    console:
+        handler: bin/console
+        timeout: 120 # in seconds
+        layers:
+            - ${bref:layer.php-80} # PHP
+            - ${bref:layer.console} # The "console" layer
+
+package:
+    patterns:
+        # Excluded files and folders for deployment
+        - '!assets/**'
+        - '!node_modules/**'
+        - '!public/build/**'
+        - '!tests/**'
+        - '!var/**'
+        - '!.env'
+        # If you want to include files and folders that are part of excluded folders,
+        # add them at the end
+        - 'var/cache/prod/**'
--- a/bref/symfony-bridge/0.1/serverless.yaml
+++ b/bref/symfony-bridge/0.1/serverless.yaml
@@ -24,7 +24,7 @@ functions:
             - ${bref:layer.php-80-fpm}
         events:
             - httpApi: '*'
-    # This function lets us run console commands in Lambda
+    # This function let us run console commands in Lambda
     console:
         handler: bin/console
         timeout: 120 # in seconds
--- a/bref/symfony-bridge/0.1/serverless.yaml
+++ b/bref/symfony-bridge/0.1/serverless.yaml
@@ -16,8 +16,8 @@ plugins:
     - ./vendor/bref/bref
 
 functions:
-    # This function run the Symfony application
-    website:
+    # This function runs the Symfony website/API
+    web:
         handler: public/index.php
         timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
         layers:
@@ -40,7 +40,6 @@ package:
         - '!public/build/**'
         - '!tests/**'
         - '!var/**'
-        - '!.env'
         # If you want to include files and folders that are part of excluded folders,
         # add them at the end
         - 'var/cache/prod/**'


bref/symfony-bridge/0.1/serverless.yaml
```diff
@@ -0,0 +1,46 @@
...
+    # This function lets us run console commands in Lambda
+    console:
+        handler: bin/console
+        timeout: 120 # in seconds
+        layers:
+            - ${bref:layer.php-80} # PHP
+            - ${bref:layer.console} # The "console" layer
+
+package:
+    patterns:
```
t-richard, line 36: I would be rather have `package.exlude` and `package.include` over `package.patterns` as I find it clearer and easier to use and understand when not familiar with Serverless framework.

bref/symfony-bridge/0.1/serverless.yaml
```diff
@@ -0,0 +1,46 @@
...
+        # Excluded files and folders for deployment
+        - '!assets/**'
+        - '!node_modules/**'
+        - '!public/build/**'
+        - '!tests/**'
+        - '!var/**'
+        - '!.env'
```
t-richard, line 43: This seems like optimization to me and could be confusing for new users and might even break at runtime. 

IIRC when you don't have a `.env` file present, it requires you to provide a `.env.local.php` file to work which users might not be aware of.

It's documented in the bridge but may be worth not doing by default.

bref/symfony-bridge/0.1/post-install.txt
```diff
@@ -0,0 +1,8 @@
+<bg=blue;fg=white>                     </>
+<bg=blue;fg=white> Bref Symfony Bridge </>
+<bg=blue;fg=white>                     </>
+
+  * <fg=blue>Read</> the full documentation to properly configure your
+    Symfony application at <comment>https://bref.sh/docs/frameworks/symfony.html</>
```
t-richard, line 6: The bridge is not documented there.

We should document it first in the offcial docs as the bridge is starting to get more attention lately and is working well in several projects of mine. 

The other solution is to point to the bridge's README file

t-richard (changes requested):

bref/symfony-bridge/0.1/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "copy-from-recipe": {
+        "serverless.yaml": "serverless.yaml"
```
mnapoli, line 3: Should be `serverless.yml`

bref/symfony-bridge/0.1/post-install.txt
mnapoli: The official documentation will be there though, once the Symfony Bridge becomes stable. So that link makes sense (the README should be empty eventually).

bref/symfony-bridge/0.1/serverless.yaml
```diff
@@ -0,0 +1,46 @@
...
+    runtime: provided.al2
+    environment:
+        # Symfony environment variables
+        APP_ENV: prod
+
+plugins:
+    - ./vendor/bref/bref
+
+functions:
+    # This function run the Symfony application
```
mnapoli, line 19:
```suggestion
    # This function runs the Symfony website/API
```

(application is very vague, that could include the console and queue workers, so I thought clarifying should help)

bref/symfony-bridge/0.1/serverless.yaml
```diff
@@ -0,0 +1,46 @@
...
+    website:
```
mnapoli, line 20:
```suggestion
    web:
```

Inspired from the Laravel config (https://github.com/brefphp/laravel-bridge/blob/master/config/serverless.yml#L22), that leaves room for APIs.

(BTW I realize that you copied the official Bref documentation :) I'm just taking the opportunity to improve what can be improved)

bref/symfony-bridge/0.1/serverless.yaml
mnapoli: `package.patterns` is the new approach, `package.exlude` (and include) is now deprecated. Because of that, I think that makes sense to go with patterns here (I should probably update the Bref documentation at some point too).

t-richard: Ok, I wasn't aware of the deprecation and just read the reasons behind that choice. 

While I find the new notation confusing, it totally make sense to provide an up-to-date and deprecation free file by default.

bref/symfony-bridge/0.1/manifest.json
t-richard: Both are valid for Serverless and the choice to use `.yaml` may be because of the [Validation](https://github.com/symfony/recipes#validation) applied to PRs here 

> YAML files suffix must be `.yaml`, not `.yml`

But the standard for Serverles Framework is indeed `.yml`. Maybe this could be used as a "workaround"

```suggestion
        "serverless.yml": "serverless.yaml"
```

t-richard: BTW it's already accepted for `docker-compose.yml` if that may help and I think we are in the same situation because it's not Symfony config we are talking about but standard in a third-party tool

See for example : https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/manifest.json#L24

bref/symfony-bridge/0.1/serverless.yaml
mnapoli: > While I find the new notation confusing

I agree on that 😅

bref/symfony-bridge/0.1/manifest.json
ker0x: Yes, i used the `.yaml` extension to pass validation!

@t-richard wouldn't it be rather
```suggestion
        "serverless.yaml": "serverless.yml"
```
?

The other solution would be to store the `serverless.yml` into the package and use `copy-from-package` instead.

t-richard: Yes it would be `"serverless.yaml": "serverless.yml"` :facepalm: 

IMHO it makes more sense to have all here

bref/symfony-bridge/0.1/serverless.yaml
```diff
@@ -0,0 +1,45 @@
...
+    patterns:
+        # Excluded files and folders for deployment
+        - '!assets/**'
+        - '!node_modules/**'
+        - '!public/build/**'
+        - '!tests/**'
+        - '!var/**'
+        # If you want to include files and folders that are part of excluded folders,
+        # add them at the end
+        - 'var/cache/prod/**'
```
mnapoli, line 45: Oh wait one last thing, should we exclude `var/cache` entirely?

Imagine one person is deploying from their machine: we don't want to deploy the dev cache either.

ker0x: I am not sure to understand, dev cache is already excluded with `!var/**`, only the prod cache will be included in the deployement.

mnapoli: 🤦  my bad, sorry I'm completely mistaken ^^

mnapoli (approved):
All good to merge for me 👍 

Thank you!

bref/symfony-bridge/0.1/serverless.yaml
```diff
@@ -0,0 +1,45 @@
...
+functions:
+    # This function runs the Symfony website/API
+    web:
+        handler: public/index.php
+        timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
+        layers:
+            - ${bref:layer.php-80-fpm}
+        events:
+            - httpApi: '*'
+    # This function lets us run console commands in Lambda
```
OskarStark, line 27:
```suggestion
    # This function let us run console commands in Lambda
```
?

bref/symfony-bridge/0.1/serverless.yaml
```diff
@@ -0,0 +1,45 @@
...
+plugins:
+    - ./vendor/bref/bref
+
+functions:
+    # This function runs the Symfony website/API
+    web:
+        handler: public/index.php
+        timeout: 28 # in seconds (API Gateway has a timeout of 29 seconds)
+        layers:
+            - ${bref:layer.php-80-fpm}
```
OskarStark, line 24: Based on
https://github.com/brefphp/symfony-bridge/blob/6156395329df3bb34a9a17c5e440ca8ab76c173f/composer.json#L22

you can use the bridge with PHP 7.3, 7.4, but everything in the recipe contains php 8 🤔

Nyholm (approved):
Thank you

bref/symfony-bridge/0.1/serverless.yaml
t-richard: Can't see the problem with that :thinking: 

IMHO We're just providing a decent default (7.3 is EOL and 8.0 is now supported by a fair amount of packages)

If they want, users can replace this line with a previous version this way and it will work

```suggestion
            - ${bref:layer.php-74-fpm}
```

t-richard:
Thank you @ker0x :+1:

PR #1133, opened 2021-04-30
Add oskarstark/no-floc-listener

OskarStark:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/oskarstark/no-floc-listener/0.1/config/packages/oskarstark_symfony_no_floc_listener.yaml
+++ b/oskarstark/no-floc-listener/0.1/config/packages/oskarstark_symfony_no_floc_listener.yaml
@@ -1,0 +1,3 @@
+services:
+    OskarStark\Symfony\EventListener\NoFlocListener:
+        autoconfigure: true
--- a/oskarstark/no-floc-listener/0.1/manifest.json
+++ b/oskarstark/no-floc-listener/0.1/manifest.json
@@ -1,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}


oskarstark/no-floc-listener/0.1/config/packages/oskarstark_symfony_no_floc_listener.yaml
```diff
@@ -0,0 +1,7 @@
+services:
+    _defaults:
+        autowire: true
```
nicolas-grekas, line 3: if there is only one service, this can be added to the service itself instead of `_defaults`

OskarStark: I think I can remove it entirely, because it has no class members

oskarstark/no-floc-listener/0.1/config/packages/oskarstark_symfony_no_floc_listener.yaml
```diff
@@ -0,0 +1,4 @@
+services:
+    OskarStark\Symfony\EventListener\NoFlocListener:
```
Nyholm, line 2: I dont understand. 

This is an `EventSubscriberInterface`, why do you tag it as an `event_listener`?

It would be enough to remove the tag an add `autowire: true`

Nyholm (changes requested):

oskarstark/no-floc-listener/0.1/config/packages/oskarstark_symfony_no_floc_listener.yaml
OskarStark: Indeed, I mixed sth. up while copy pasting... 🤔

Nyholm (approved):
Thank you

PR #1143, opened 2021-05-07
Update FluxSESyliusPayumStripePlugin

Prometee:
| Q             | A
| ------------- | ---
| License       | MIT

Add missing default configurations with related comments.


Nyholm (changes requested):
We don't need to add default config.

I would be more happy to remove default config. =)

PR #1144, opened 2021-05-07
[PHP_CodeSniffer] Update php_codesniffer recipe to use PSR12.

jenkoian:
| Q             | A
| ------------- | ---
| License       | MIT

Wasn't sure about the version bump but _think_ this is right...

* Added `3.6` (as that's the latest version of `php_codesniffer`) due to 'BC break' of changing the coding standard.
* Symlinked manifest file to `3.0` version (as didn't change)

I realise the vast majority of people using symfony will use php-cs-fixer, but for those who do wish to use `php_codesniffer` the Symfony standard isn't easily supported (there are [community efforts](https://github.com/djoos/Symfony-coding-standard) I believe). However, for simplicity and to be in line with the previous version I think it makes sense to just keep this to PSR12.

### DIFF
--- a/squizlabs/php_codesniffer/3.6/manifest.json
+++ b/squizlabs/php_codesniffer/3.6/manifest.json
@@ -1,1 +1,9 @@
-../3.0/manifest.json
+{
+    "copy-from-recipe": {
+        "phpcs.xml.dist": "phpcs.xml.dist"
+    },
+    "gitignore": [
+        "/.phpcs-cache",
+        "/phpcs.xml"
+    ]
+}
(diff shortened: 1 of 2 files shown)


squizlabs/php_codesniffer/3.6/manifest.json
```diff
@@ -0,0 +1 @@
+../3.0/manifest.json
```
Nyholm, line 1: Let's copy this file instead. It will be easier to read and maintain.

jenkoian: Done.

Nyholm (approved):
Awesome. Thank you

PR #1145, opened 2021-05-08
kunstmaan/admin-bundle: Disable exception logging by default

acrobat:
| Q             | A
| ------------- | ---
| License       | MIT

The upcoming 5.9 version allows to disable the exception logging module and this is a more sensible default config. The 5.9 version is not yet released, but this PR prepare the recipe already

### DIFF
--- a/kunstmaan/admin-bundle/5.9/config/packages/kunstmaan_admin.yaml
+++ b/kunstmaan/admin-bundle/5.9/config/packages/kunstmaan_admin.yaml
@@ -1,0 +1,7 @@
+kunstmaan_admin:
+    multi_language: true
+    required_locales: nl|fr|en
+    default_locale: en
+    admin_locales: ['en']
+    website_title: 'MyProject' #Rename this to your website name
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Please rebase to run checks again.

nicolas-grekas:
To make the create-project check pass, you need to:
1. add `composer config minimum-stability dev` in `.github/workflows/create-project.yml` just after the existing `composer config` lines
2. update the branch-alias of your bundle to `5.9-dev`

acrobat:
@nicolas-grekas done! But wouldn't that change it for all projects?

Also now fails the validation check, can this be ignored?

nicolas-grekas:
I made the change for all packages in 94f3802
Can you please rebase one more time?

acrobat:
@nicolas-grekas Done! The 4.4 project build works correctly now, 5.x is expected to fail as that bundle is not compatible with sf5 yet.

nicolas-grekas (approved):

PR #1151, opened 2021-05-12
Run Travis on PHP 8 and drop SF 3.4

Yozhef:
The problem when the project uses symfony/framework-bundle.
composer require --ignore-platform-reqs install  psr/cache 2.0, his support only PHP 8, but in Travis use PHP 7.4.
1) Remove --ignore-platform-reqs 
or 
2) Use in build PHP 8.0

PHP 8.0 now many libraries does not use this version, because I think good idea remove --ignore-platform-reqs

| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,15 +8,10 @@ cache:
 matrix:
     fast_finish: true
     include:
-        - php: 7.4
-          env: SKELETON_VERSION="^3.4"
-        - php: 7.4
+        - php: 8.0
           env: SKELETON_VERSION="^4.4"
-        - php: 7.4
+        - php: 8.0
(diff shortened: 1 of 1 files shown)


Nyholm:
I think `--ignore-platform-reqs` was added to "support" all kinds of extensions. 

I would prefer to update to PHP8. It is mostly new packages that are added anyways. 

Btw, can you also remove the SF 3.4 job?

Yozhef:
@Nyholm yes of course

Yozhef:
@Nyholm tell me please I understood you correctly?
image

Nyholm (approved):
Great. I agree with this.

PR #1154, opened 2021-05-19
Add recipes for Jane AutoMapper v7 release

Korbeil:
| Q             | A
| ------------- | ---
| License       | MIT

This PR adds all needed recipes for Jane v7 release including following components or bundle:
- `jane-php/automapper-bundle`, will require to add the bundle to the bundle list;

### DIFF
--- a/jane-php/automapper-bundle/7.0/manifest.json
+++ b/jane-php/automapper-bundle/7.0/manifest.json
@@ -1,0 +1,5 @@
+{
+    "bundles": {
+        "Jane\\Bundle\\AutoMapperBundle\\JaneAutoMapperBundle": ["all"]
+    }
+}
--- a/jane-php/json-schema/6.0/post-install.txt
+++ b/jane-php/json-schema/6.0/post-install.txt
@@ -6,7 +6,8 @@
     1. Configure <comment>config/jane/json_schema.php</> with your specification details
     2. Run <comment>bin/jane-json-schema-generate</>
     3. Add <comment>generated/</> directory to your composer autoload definition (eg. <comment>"MyApp\\Library\\Generated\\": "generated/"</>)
-    4. Open <comment>config/packages/jane.yaml</> and replace <comment>MyApp\Library\Generated\</> namespace with your generated namespace.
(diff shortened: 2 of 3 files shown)


jane-php/json-schema/7.0/bin/jane-json-schema-generate
```diff
@@ -0,0 +1,15 @@
...
+
+require __DIR__ . '/../vendor/autoload.php';
+
+use Jane\Component\JsonSchema\Console\Command\GenerateCommand;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\NullOutput;
+use Jane\Component\JsonSchema\Console\Loader\ConfigLoader;
+use Jane\Component\JsonSchema\Console\Loader\SchemaLoader;
+
+$command = new GenerateCommand(new ConfigLoader(), new SchemaLoader());
```
nicolas-grekas, line 12: why make it special and not a regular Symfony command registered by the bundle?

nicolas-grekas: This is unexpected to me, ppl are used to run commands through the `bin/console` script, so if that's possible, that's better indeed

Korbeil: Since that library isn't a bundle it was not made that way, but I'll add a bundle to handle that :+1:

Korbeil: I removed the json-schema / open-api library updates, let's focus on the AutoMapper bundle, the rest will come later when I have a bundle for each of them :)

jane-php/automapper-bundle/7.0/config/packages/automapper.yaml
```diff
@@ -0,0 +1,2 @@
+jane_auto_mapper:
+    normalizer: false
```
nicolas-grekas, line 2: can't we drop this file and have this be the default?

Korbeil:
The recipe isn't needed anymore.

PR #1155, opened 2021-06-03
Add jawira/case-converter-twig

jawira:
| Q             | A
| ------------- | ---
| License       | MIT

This recipe installs the filters provided by https://github.com/jawira/case-converter-twig


Nyholm (approved):
Thank you

PR #1165, opened 2021-06-18
nedac/sylius-temporarily-out-of-stock-plugin

4c0n:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/nedac/sylius-temporarily-out-of-stock-plugin/1.0/config/nedac_sylius_temporarily_out_of_stock_plugin.yaml
+++ b/nedac/sylius-temporarily-out-of-stock-plugin/1.0/config/nedac_sylius_temporarily_out_of_stock_plugin.yaml
@@ -1,0 +1,5 @@
+doctrine:
+    orm:
+        dql:
+            string_functions:
+                FIRST: Nedac\SyliusTemporarilyOutOfStockPlugin\Doctrine\Dql\FirstFunction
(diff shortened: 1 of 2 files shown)


vvasiloi (approved):

4c0n:
@Nyholm Would you mind having a look at this when you have time? Thanks :smiley:

nedac/sylius-temporarily-out-of-stock-plugin/1.0/config/nedac_sylius_temporarily_out_of_stock_plugin.yaml
```diff
@@ -0,0 +1,5 @@
+doctrine:
+    orm:
+        dql:
+            string_functions:
+                FIRST: Nedac\SyliusTemporarilyOutOfStockPlugin\Doctrine\Dql\FirstFunction
```
Nyholm, line 5: I must say that it is a little strange that the ”out of stock bundle” requires special dbal syntax. Is this really needed?

4c0n: Yeah I can imagine you do find it a little strange, it's because it's not just products that are queried in for example the search results, but also the product variants and they need to be ordered on which one has the most stock available in a subquery. Unfortunately however the "limiting" of results in doctrine is done in the collection (on the query results) and not in the query itself, so that's why this is necessary.
I'd prefer not to have to do this and some other things, but Doctrine doesn't provide a better solution and are not the most cooperative devs in my experience. There are some major design flaws in the current version that they don't seem at all interested in fixing for example: https://github.com/doctrine/dbal/pull/4103 which is a shame really.

Nyholm (approved):
Thank you

4c0n:
Thanks @Nyholm !

PR #1166, opened 2021-06-24
Fix compatibility with Webpack-Encore in bref/symfony-bridge

t-richard:
| Q             | A
| ------------- | ---
| License       | MIT

At the moment the whole `public/build` directory is excluded. 

This means the `entrypoints.json` and `manifest.json` files are also excluded but are required by Symfony to resolve filenames correctly and load Encore entrypoints in Twig.

Without this, people would get a 500 error when deploying on Lambda.


mnapoli (approved):
🙏  thank you!

weaverryan (approved):

PR #1177, opened 2021-07-13
Add nbgrp/env-bundle 1.0

a-menshchikov:
| Q             | A
| ------------- | ---
| License       | MIT


Nyholm:
Thank you for this PR. 
I am a bit thorn. I usually don't allow recipes that only have default config. And I always suggest that one should link to the readme instead of adding config in comments. 

Recipes like this are useful when 90% of the users will change the default config. Ie, the number of items to show in a PaginationBundle etc. Is that the scenario with this bundle too?

a-menshchikov:
@Nyholm thank you for your comment.
All env-bundle processors will be disabled by default, therefore IMO 100% of the bundle users will change the default config.
I added link to the bundle's readme into config. I would like to keep commented csv processor config (lines 6:8) and inline comments, because it allows to speed up bundle configuration for advanced users.

Nyholm (approved):
Okey. It seams fair. 

Thank you

PR #1187, opened 2021-08-04
added recipe for SvcVersioningBundle

Sven-Ve:
| Q             | A
| ------------- | ---
| License       | MIT

Thank you for merging :-)


Nyholm (commented):
Why did you add `false` defaults in the recipe but `true` in your bundle config?

https://github.com/Sven-Ve/svc-versioning-bundle/blob/v1.4.0/src/DependencyInjection/Configuration.php

fabpot (approved):

PR #1189, opened 2021-08-08
add recipe for pentatrion/vite-bundle

lhapaipai:
| Q             | A
| ------------- | ---
| License       | MIT

Hi,
This Pull Request adds the recipe for the [pentatrion/vite-bundle](https://github.com/lhapaipai/vite-bundle).


Nyholm (commented):
Hey. 

I suggest to put the css/js/ts in your package and use `copy-from-package`. I also think we should remove the `yaml` file because it does not provide any config, it is just comments.

lhapaipai:
Hi @Nyholm, thanks for your review,
do you think i should move too the `vite.config.js` and `package.json` files into my package ?

Nyholm:
Yes. That sounds like a good idea. Also app.js and app.css

Nyholm (approved):
Thank you

PR #1191, opened 2021-08-10
Woleet naming convention

germainlecourtoissloop:
| Q             | A
| ------------- | ---
| License       | MIT

Naming convention from the Woleet CTO


Nyholm (changes requested):
Thank you.

There is no need for the recipe for 1.0. The recipe will only be used for new installs and nobody will install 1.0 over 1.2. 

I also think you should use environment variables for the API token and callback secret.

germainlecourtoissloop:
OK thanks @Nyholm 

So, I remove recipe 1.0 and add %env system ?

Nyholm:
Yes please

germainlecourtoissloop:
Done :)

Nyholm (approved):
Thank you

PR #1263, opened 2021-08-18
Add datana-gmbh/datapool-api

OskarStark:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/datana-gmbh/datapool-api

### DIFF
--- a/datana-gmbh/datapool-api/1.0/config/packages/datana_datapool_api.yaml
+++ b/datana-gmbh/datapool-api/1.0/config/packages/datana_datapool_api.yaml
@@ -1,0 +1,9 @@
+services:
+    _defaults:
+        autowire: true
+
+    Datana\Datapool\Api\DatapoolClient:
+        arguments:
+            - '%env(DATAPOOL_API_BASE_URI)%'
+            - '%env(DATAPOOL_API_USERNAME)%'
+            - '%env(DATAPOOL_API_PASSWORD)%'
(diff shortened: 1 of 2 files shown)


Nyholm (approved):
Thank you. I agree with this. 

@nicolas-grekas the CI is giving me a false positive. 

> "~" should be replaced with "null"

I dont think this is true for service definitions.

datana-gmbh/datapool-api/1.0/config/packages/datana_datapool_api.yaml
```diff
@@ -0,0 +1,30 @@
...
+            - '%env(DATAPOOL_API_USERNAME)%'
+            - '%env(DATAPOOL_API_PASSWORD)%'
+
+    Datana\Datapool\Api\AktenEventLogApi: ~
+    Datana\Datapool\Api\NullAktenEventLogApi: ~
+
+    Datana\Datapool\Api\AktenzeichenApi: ~
+    Datana\Datapool\Api\FakeAktenzeichenApi: ~
+
+when@prod:
```
Nyholm, line 17: This is not needed. 

Currently there is no service for `Datana\Datapool\Api\AktenzeichenApiInterface` defined by default. I suggest to make "prod" environment the default one.

OskarStark: No, in this case if someone would add a `stage` environment, it is not matched by `when@dev` and `when@test`, so the result would be, that the prod one is used, which could end up in write operations to production system of the service.... Thats what I want to avoid!

OskarStark:
@Nyholm can we please merge this?

OskarStark:
Friendly ping

PR #1269, opened 2021-08-31
add recipe for dunglas/digital-ocean-bundle

dunglas:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/dunglas/digital-ocean-bundle


dunglas:
The error is a false positive. This bundle needs a `psr/http-client-implementation` and a `psr/http-factory-implementation` implementation.

The command to install it with HttpClient is `composer require dunglas/digital-ocean-bundle symfony/http-client nyholm/psr7 guzzlehttp/promises`, but any other compatible client can work.

chalasr (approved):

OskarStark (approved):

PR #1275, opened 2021-09-10
Add a recipe for php-http/httplug-bundle version 1.23

do-logic:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/php-http/httplug-bundle


Nyholm (changes requested):
This is the exact same recipe as earlier version. We dont need to update the recipe for each version. 

(It is a great bundle though)

PR #1279, opened 2021-09-16
remove cmf recipes as the project is not maintained anymore

dbu:
Q | A
-- | --
License | MIT

as discussed with @javiereguiluz , we will not only remove the cmf doc from symfony.com, but also stop maintaining most cmf components as there is not enough interest in them.


fabpot (approved):

javiereguiluz (approved):
Thanks David!

PR #1283, opened 2021-09-19
Remove bref/symfony-bridge

Nyholm:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/bref/symfony-bridge

This recipe is replaced by https://github.com/symfony/recipes/pull/1001


fabpot (approved):

t-richard:
@Nyholm I find this confusing, in https://github.com/brefphp/bref/pull/1032 you say

> I also didnt want to remove docs how to use Bref with non-runtime symfony applications. That is why I added a new page.

But removing this and moving files to the recipe for `runtime/bref` makes it harder for people using the FPM layer because they are on older symfony versions or don't want to use the runtime for some reason.

Removing this would force us to rollback changes to the Symfony part of Bref docs to put back Serverless framework examples that were auto-generated by the recipe.

I understand that the Runtime is the way to go now and I would probably use it on my projects but this seems like a step backward to me. 

If this is merged, then my take would be to rework bref docs first to use the runtime instead of FPM. If this is not possible, I would instead keep this recipe as-is and add only files related to `runtime/bref` in its recipe.

In any case, I would like this PR to wait so that we can have a proper discussion in the bref community and adapt bref docs if necessary.

Nyholm:
Oh. You are correct. I don’t think about that. Hm. Maybe we should close this PR. Let’s leave it open for a while to see how the discussion goes. 

Thank you.

fabpot:
Closing for now then.
Feel free to reopen.

PR #1288, opened 2021-10-19
Add recipe for rikudou/psr6-dynamo-db-bundle

t-richard:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/rikudou/psr6-dynamo-db-bundle

This bundle provides a PSR-6 cache adapter for Symfony + a Session handler both storing data in AWS DynamoDB.

I enabled it only in `prod` because it would otherwise require a local DynamoDB emulator (probably using Docker) or a real DynamoDB table for local development which is not ideal.

Also, `AWS_REGION` is commented because
* it could already be defined if using another AWS-related package 
* it might be predefined in an AWS environment (eg. variables exists in AWS Lambda).

@RikudouSage would love your thoughts on this before going further.


t-richard:
The CI failure is because the bundle is not compatible with Symfony 4.4 and the CI tries to install it on a 4.4 skeleton project. 

Not sure if that's the intended behaviour.

t-richard:
Is there anything blocking the merge of this PR?

Nyholm (approved):
Thank yuo

PR #1307, opened 2021-11-10
added 4.4 config file

ychadwick:
| Q             | A
| ------------- | ---
| License       | MIT


ychadwick:
How can I define which PHP version should be used in the Test? Is it even possible, or does the bundle have to working in php 8?

Nyholm (approved):
Thank you. 

There is no way to define PHP versions Im afraid.

PR #1317, opened 2021-12-09
`FollowSymlinks` to be enabled at any times

Hanmac:
RewriteRule needs FollowSymlinks to be enabled at any times

| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | 

without the Option you get:
> AH00670: Options FollowSymLinks and SymLinksIfOwnerMatch are both off, so the RewriteRule directive is also forbidden due to its similar ability to circumvent directory restrictions

### DIFF
--- a/symfony/apache-pack/1.0/public/.htaccess
+++ b/symfony/apache-pack/1.0/public/.htaccess
@@ -18,6 +18,10 @@ DirectoryIndex index.php
 </IfModule>
 
 <IfModule mod_rewrite.c>
+    # This Option needs to be enabled for RewriteRule, otherwise it will show an error like
+    # 'Options FollowSymLinks or SymLinksIfOwnerMatch is off which implies that RewriteRule directive is forbidden'
+    Options +FollowSymlinks
+
     RewriteEngine On
 
     # Determine the RewriteBase automatically and set it as environment variable.


symfony/apache-pack/1.0/public/.htaccess
```diff
@@ -18,6 +18,10 @@ DirectoryIndex index.php
 </IfModule>
 
 <IfModule mod_rewrite.c>
+    # This Option needs to be enabled for RewriteRule, otherwise it will show an error like
+    # 'Options FollowSymLinks or SymLinksIfOwnerMatch is off which implies that RewriteRule directive is forbidden'
+    Options +FollowSymlinks
+
```
Nyholm, line 24: This is added as a comment just a few lines above. 

I suggest to not change the defaults and let the user uncomment line 12 instead.

OskarStark: @fabpot WDYT?

fabpot: Last time I used Apache was probably more than 15 years ago, so I have no idea.
I would tend to trust people using Apache :)

OskarStark: Same here

OskarStark (approved):

OskarStark:
@Hanmac can you please rebase, so the auto merge can kick in? or @fabpot please klick the merge button 😄

PR #1324, opened 2021-12-20
Recipes/easy block bundle

arnaud-ritti:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [https://packagist.org/packages/agence-adeliom/easy-block-bundle](https://packagist.org/packages/agence-adeliom/easy-block-bundle)


fabpot:

PR #1331, opened 2021-12-20
Recipes/easy page bundle

arnaud-ritti:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [https://packagist.org/packages/agence-adeliom/easy-page-bundle](https://packagist.org/packages/agence-adeliom/easy-page-bundle)

### DIFF
--- a/agence-adeliom/easy-page-bundle/1.0/config/packages/easy_page.yaml
+++ b/agence-adeliom/easy-page-bundle/1.0/config/packages/easy_page.yaml
@@ -1,0 +1,19 @@
+easy_page:
+  page_class: App\Entity\EasyPage\Page
+  page_repository: App\Repository\EasyPage\PageRepository
+  page_controller: App\Controller\EasyPage\PageController
+
+  layouts:
+    front:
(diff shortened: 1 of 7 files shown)


agence-adeliom/easy-page-bundle/1.0/config/packages/easy_page.yaml
```diff
@@ -0,0 +1,19 @@
...
+            resource: '@EasyPage/default_layout.html.twig'
+            pattern: ''
+            assets_css:
+                - 'https://cdn.jsdelivr.net/npm/bootstrap@5.1.0/dist/css/bootstrap.min.css'
+            assets_js:
+                - 'https://cdn.jsdelivr.net/npm/bootstrap@5.1.0/dist/js/bootstrap.bundle.min.js'
+
+
+services:
+        App\Controller\Page\PageController:
```
Snowbaha, line 17: Hello, it should be `App\Controller\EasyPage\PageController` ?

fabpot:

Closing for now as we cannot merge it as is. Feel free to reopen if/when the bundles can generate the PHP code.

PR #1332, opened 2021-12-20
Recipes/easy redirect bundle

arnaud-ritti:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [https://packagist.org/packages/agence-adeliom/easy-redirect-bundle](https://packagist.org/packages/agence-adeliom/easy-redirect-bundle)

### DIFF
--- a/agence-adeliom/easy-redirect-bundle/1.0/src/Entity/EasyRedirect/NotFound.php
+++ b/agence-adeliom/easy-redirect-bundle/1.0/src/Entity/EasyRedirect/NotFound.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace App\Entity\EasyRedirect;
+
+use Adeliom\EasyRedirectBundle\Entity\NotFound as BaseNotFound;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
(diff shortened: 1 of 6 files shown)


agence-adeliom/easy-redirect-bundle/1.0/src/Entity/EasyRedirect/NotFound.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace App\Entity\EasyRedirect;
+
+use Adeliom\EasyRedirectBundle\Entity\NotFound as BaseNotFound;
+use Doctrine\ORM\Mapping as ORM;
+
+/**
+ * @ORM\Entity()
+ * @ORM\Table(name="easy_redirect__not_found")
```
OskarStark, line 10: What about using attributes?

OskarStark (commented):
Are theses PRs still a thing @arnaud-ritti ?

fabpot:

Closing for now as we cannot merge it as is. Feel free to reopen if/when the bundles can generate the PHP code.

PR #1344, opened 2022-01-26
added pre-commit recipe

mustafa-natiq:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/xip-solid-foundation/pre-commit


Nyholm (changes requested):
I see you add config for Dama, PHP-cs-fixer and PHPUnit. 

Recipes should not modify other packages config.

fabpot:
Closing as there is no more activity.

PR #1349, opened 2022-02-02
[sylius/resource-bundle] Sylius configuration for routes with attributes' feature

loic425:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | 

Only the routes/sylius_resource.yaml file has been added. the other files are just "copied".

### DIFF
--- a/sylius/resource-bundle/1.9/manifest.json
+++ b/sylius/resource-bundle/1.9/manifest.json
@@ -1,0 +1,11 @@
+{
+    "container": {
+        "locale": "en"
+    },
+    "bundles": {
+        "Sylius\\Bundle\\ResourceBundle\\SyliusResourceBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    }
+}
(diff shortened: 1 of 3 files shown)


sylius/resource-bundle/1.9/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "container": {
+        "locale": "en"
```
Nyholm, line 3: Why is this needed?

Nyholm (changes requested):
Thank you. Just one comment.

sylius/resource-bundle/1.9/manifest.json
vvasiloi: Legacy. The bundle still uses this parameter as the default locale for translations instead of using configuration. 🙈

loic425: @vvasiloi Maybe we can fix it before RC release. But I think we have to mention it on UPGRADE file, cause it can break the configuration.

vvasiloi: Yes, we could deprecate it. Not sure how to configure it though. Should we define new config, or use the config of Symfony Translations? Right now it assumes that parameter is used to configure Symfony Translations.

vvasiloi: @loic425 we're waiting for https://github.com/Sylius/SyliusResourceBundle/pull/374, right?

loic425: Yes!

fabpot:
Closing as there is no more activity and the PR cannot be merged as is. Feel free to reopen it if you like.

PR #1355, opened 2022-02-14
Added recipes for coka/doctrine-encrypt-bundle

CedrickOka:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |

### DIFF
--- a/coka/doctrine-encrypt-bundle/1.0/manifest.json
+++ b/coka/doctrine-encrypt-bundle/1.0/manifest.json
@@ -1,0 +1,11 @@
+{
+    "bundles": {
+        "Oka\\Doctrine\\EncryptBundle\\OkaDoctrineEncryptBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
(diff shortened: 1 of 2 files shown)


coka/doctrine-encrypt-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "bundles": {
+        "Oka\\Doctrine\\EncryptBundle\\OkaDoctrineEncryptBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "OKA_DOCTRINE_ENCRYPT_SECRET": "%generate(secret)%%generate(secret)%"
```
Nyholm, line 9: Why do you need two secrets? 
I think this would be okey with just one `%generate(secret)%`

Nyholm (dismissed):
Thank you. Just one small change

coka/doctrine-encrypt-bundle/1.0/manifest.json
CedrickOka: HI

In the doc I read that the function generate random alphanumeric sequence of 16 bytes I wanted to have some on 32 bytes

fabpot (approved):

PR #1368, opened 2022-03-10
Remove create_project_php parameter

vvasiloi:
It was removed from the callable workflow https://github.com/symfony/recipes/commit/07cc58ffd13fe05406b7dd23b9e5034c6dfd82fa
It currently breaks the workflow: https://github.com/symfony/recipes-contrib/actions/runs/1962132102/workflow


nicolas-grekas:
Thanks

PR #1371, opened 2022-03-17
Adding `logOutputFile: '/dev/null'`

ThomasLandauer:
See https://github.com/dompdf/dompdf/issues/2810#issuecomment-1070927239 for the reason and https://github.com/dompdf/dompdf/issues/1709#issuecomment-466717959 for the solution

| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://repo.packagist.org/packages/nucleos/dompdf-bundle

### DIFF
--- a/nucleos/dompdf-bundle/3.1/config/packages/nucleos_dompdf.yaml
+++ b/nucleos/dompdf-bundle/3.1/config/packages/nucleos_dompdf.yaml
@@ -1,4 +1,4 @@
 nucleos_dompdf:
     defaults:
         chroot: '%kernel.project_dir%/public/assets' # See https://github.com/dompdf/dompdf/wiki/Usage#security-restrictions-for-local-files
-        logOutputFile: '/dev/null' # Prevent creating `/tmp/log.htm` which causes problems in some server configurations
+        logOutputFile: '' # Prevent creating `/tmp/log.htm` which causes problems in some server configurations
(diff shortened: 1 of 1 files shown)


vvasiloi:
Hey @ThomasLandauer, if this options is passed to [Dompdf](https://github.com/dompdf/dompdf), then I think it's better to set it's value to `null` (`logOutputFile: ~`) or another _falsy_ value to disable the logging altogether. Also, `/dev/null` won't work on Windows. 😅

ThomasLandauer:
@vvasiloi Thanks, I changed it to empty string, according to https://github.com/dompdf/dompdf/issues/2810#issuecomment-1070981789
P.S.: It's meant to solve a Linux-only problem (file permission), so I didn't really care about Windows ;-)

vvasiloi:
> @vvasiloi Thanks, I changed it to empty string, according to [dompdf/dompdf#2810 (comment)](https://github.com/dompdf/dompdf/issues/2810#issuecomment-1070981789) P.S.: It's meant to solve a Linux-only problem (file permission), so I didn't really care about Windows ;-)

I understand. I was just pointing that while this fixes an issue on Linux, it introduces another one on Windows.
I know very few (maybe even no one) would deploy this to a Windows server, but a lot still develop on Windows and that's where this would have failed.

vvasiloi (approved):
Thanks @ThomasLandauer!

PR #1372, opened 2022-03-18
update imported resources for bundle config

ychadwick:
| Q             | A
| ------------- | ---
| License       | MIT


ychadwick:
This project is currently only running on symfony 4.4 and PHP 7.4, for this reason the "Run updated recipes" check fails.
I have previously asked if it is possible to specify which versions a recipe is available for, so that it is only tested against those parameters. This is, as far as I understand, not possible.
Please get back to me if there are any questions.

ychadwick:
I still did not hear back from you, and for some reason it looks as if our recipe has been deleted from the repository, can someone please give me an update on the status?

PR #1385, opened 2022-04-02
add phpzlc 2.0

CJayHe:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |


CJayHe:
My newly released version currently only supports version 5.4, not 6. I have no way to pass automated detection. I don't quite understand.

PR #1386, opened 2022-04-04
Add sulu form bundle 2.4 recipe

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [sulu/form-bundle](github.com/sulu/suluformbundle)

The 2.4 not longer requires enabling of the esi and fragments configuration. So the following was removed: https://github.com/symfony/recipes-contrib/blob/7645718b41e85fdadea21149322cbabf9c2d4840/sulu/form-bundle/2.2/config/packages/sulu_form.yaml#L9-L11


fabpot (approved):

alexander-schranz:
@fabpot Thx :)

PR #1390, opened 2022-04-11
Add a recipe for gesdinet/jwt-refresh-token-bundle

mbabker:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [`gesdinet/jwt-refresh-token-bundle`](https://packagist.org/packages/gesdinet/jwt-refresh-token-bundle)

Note: Similar to #1119, the Q&A build can't pass because the default config requires optional dependencies that can't be installed for testing here.

### DIFF
--- a/gesdinet/jwt-refresh-token-bundle/1.0/src/Entity/RefreshToken.php
+++ b/gesdinet/jwt-refresh-token-bundle/1.0/src/Entity/RefreshToken.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Gesdinet\JWTRefreshTokenBundle\Entity\RefreshToken as BaseRefreshToken;
+
+/**
+ * @ORM\Entity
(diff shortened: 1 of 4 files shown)


gesdinet/jwt-refresh-token-bundle/1.0/src/Entity/RefreshToken.php
```diff
@@ -0,0 +1,13 @@
...
+
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Gesdinet\JWTRefreshTokenBundle\Entity\RefreshToken as BaseRefreshToken;
+
+/**
+ * @ORM\Entity
+ */
+class RefreshToken extends BaseRefreshToken
```
gharlan, line 11: When I use it like this, the bundle tries to use the table `refresh_token` without `s`. Is this table name change intended?
I've added `@ORM\Table("refresh_tokens")` to keep the current name.

chalasr (approved):

PR #1396, opened 2022-04-22
Use when@env for jms serializer configuration

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | jms/serializer-bundle

/cc @goetas 

Symfony switch to use `when@env` everywhere in there core bundle configuraitons.

This will update jms serializer to use als `when@prod` and `when@dev` instead of splitting it into seperate files.

### DIFF
--- a/jms/serializer-bundle/4.0/config/packages/jms_serializer.yaml
+++ b/jms/serializer-bundle/4.0/config/packages/jms_serializer.yaml
@@ -1,0 +1,30 @@
+jms_serializer:
+    visitors:
+        xml_serialization:
+            format_output: '%kernel.debug%'
+#    metadata:
+#        auto_detection: false
+#        directories:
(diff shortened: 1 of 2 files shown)


goetas:
if this is the new standard for defining multi env configs, looks good to me :+1:

goetas (approved):

Johnmeurt (commented):
The `when` keyword was introduced in symfony 5.3. But the version 4.0 of [JMSSerializerBundle](https://github.com/schmittjoh/JMSSerializerBundle) is still compatible with symfony 4.4 application. This should not be merged before the end of life of symfony 4.4

fabpot:
4.4 reaches its EOL now.

fabpot (approved):

PR #1398, opened 2022-04-22
Allow `workflow_dispatch` for update workflow

OskarStark:
| Q             | A
| ------------- | ---
| License       | MIT

Until we know, why the workflow is not dispatched automatically, we should be able to update the flex endpoint manually.


OskarStark:
I think this is not needed anymore

PR #1417, opened 2022-07-07
made sentry config for prod only

### LINKED ISSUE #1115: [Sentry Bundle] Clarify config when using monolog
After recently upgrading an old Symfony 3 install to the latest 4.0 version, we ended up spending several days trying to debug why no PHP errors were being sent to our log files or the [Sentry](sentry.io) application monitoring service.

As it turns out, I had incorrectly assumed that the default flex recipe provided would setup Sentry correctly. 

With all our logs being blank, it was impossible to tell where the issue was. Further confused by the fact that some console errors were making their way through.

As per the Sentry docs and this PR, there are actually some extra steps to get Sentry to work correctly with Monolog: 
https://docs.sentry.io/platforms/php/guides/symfony/#monolog-integration

https://github.com/getsentry/sentry-symfony/pull/444

If I was to take a guess, I'd imagine most Symfony installs will rely on Monolog at some point in their lifetime. Similar to how other bundles do, would it perhaps make sense to have this commented out in the flex recipe like so:

```
sentry:
    dsn: '%env(SENTRY_DSN)%'   

# If you are using Monolog, you also need these additional configuration and services to log the errors correctly:
# https://docs.sentry.io/platforms/php/guides/symfony/#monolog-integration
    # register_error_listener: false

# monolog:
#    handlers:
#        sentry:
#            type: service
#            id: Sentry\Monolog\Handler

# services:
#    Sentry\Monolog\Handler:
#        arguments:
#           $hub: '@Sentry\State\HubInterface'
#            $level: !php/const Monolog\Logger::ERROR
```

Without this config step, all logs were getting consumed by Sentry and the environment became very hard to debug.

Hopefully this request is correct and please do shout if it needs any further clarification :)

robinlehrmann:
- added monolog example config to omit confusion
- made sentry config for prod env only

| Q             | A
| ------------- | ---
| License       | MIT

Should resolve #1115

### DIFF
--- a/sentry/sentry-symfony/3.0/config/packages/prod/sentry.yaml
+++ b/sentry/sentry-symfony/3.0/config/packages/prod/sentry.yaml
@@ -1,0 +1,22 @@
+sentry:
+    dsn: '%env(SENTRY_DSN)%'
+    options:
+        excluded_exceptions:
+            - Symfony\Component\HttpKernel\Exception\NotFoundHttpException
+            - Symfony\Component\Security\Core\Exception\AccessDeniedException
+
+    # If you are using Monolog, you also need these additional configuration and services to log the errors correctly:
+    # https://docs.sentry.io/platforms/php/guides/symfony/#monolog-integration
+    # register_error_listener: false
+
+    # monolog:
+    #    handlers:
+    #        sentry:
+    #            type: service
+    #            id: Sentry\Monolog\Handler
+
+    # services:
+    #    Sentry\Monolog\Handler:
+    #        arguments:
+    #           $hub: '@Sentry\State\HubInterface'
+    #            $level: !php/const Monolog\Logger::ERROR
--- a/sentry/sentry-symfony/3.0/config/packages/sentry.yaml
+++ b/sentry/sentry-symfony/3.0/config/packages/sentry.yaml
@@ -1,2 +1,0 @@
-sentry:
-    dsn: '%env(SENTRY_DSN)%'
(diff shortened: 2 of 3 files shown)


robinlehrmann:
Can someone tell me, why the job is failing 🤔 ?

vvasiloi:
@robinlehrmann probably the package is not compatible with PHP 8.1.

robinlehrmann:
> probably the package is not compatible with PHP 8.1.
@vvasiloi Yes, but why was this working before? I just changed the config in this MR 😅

vvasiloi:
Before when? Maybe it wasn't testing it with PHP 8.1.

vvasiloi:
Also, bear in mind that `when@prod`  was introduced in Symfony 5.3, so it won't work on versions prior to that, which I see are the target of Sentry bundle v3.

robinlehrmann:
> Also, bear in mind that `when@prod` was introduced in Symfony 5.3, so it won't work on versions prior to that, which I see are the target of Sentry bundle v3.

Thanks for the hint :). I changed it now and created a prod directory in `config/packages`

robinlehrmann:
> Before when? Maybe it wasn't testing it with PHP 8.1.

I guess it does not work, because it's [here](https://github.com/getsentry/sentry-symfony/blob/3.0.0-beta1/composer.json) not compatible with php 8 at all.

vvasiloi (approved):

vvasiloi:
Thanks, @robinlehrmann!

robinlehrmann:
You are welcome :) @vvasiloi

fabpot:
We've moved all recipes to not use the dub-directories anymore in favor of using the `when@prod` notation. I think this is what should be done here as well?
Not sure as it depends on the version of Symfony. /cc @nicolas-grekas

robinlehrmann:
@fabpot Just in case: https://github.com/symfony/recipes-contrib/pull/1419/files  :) 
I also added 4.0 because with php 8.1 it did not work anymore :/

PR #1418, opened 2022-07-09
Features/coka/notifier server bundle

CedrickOka:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |


fabpot (changes requested):
Entity classes should probably be removed.

PR #1421, opened 2022-07-12
add patchlevel event-sourcing 2.0 receipe

DavidBadura:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [patchlevel/event-sourcing-bundle](https://packagist.org/packages/patchlevel/event-sourcing-bundle)

Updated the recipe for patchlevel event-sourcing 2.0


DavidBadura:
hey,

I looked to see if I could somehow get the pipeline green so that the auto-merge from the bot would go through. Unfortunately, the auto-merge is only executed when the entire pipeline is green.

I don't get that because we support symfony 5, but not php7.4. That's why the test fails. 
Since the PR has also been open for a while, I don't know how I can push it further.

fabpot (approved):

DavidBadura:
That was quick, thank you very much!

PR #1428, opened 2022-07-23
update VichUploaderBundle recipe

garak:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [uploader-bundle](https://packagist.org/packages/vich/uploader-bundle)

Since version 1.13, UploaderBundle deprecated not using a namer. This new configuration reflects the new behaviour.
See also [UPGRADE.md](https://github.com/dustin10/VichUploaderBundle/blob/master/UPGRADE.md#upgrading-from-v1120-to-1130) on bundle repo.

### DIFF
--- a/vich/uploader-bundle/1.13/config/packages/vich_uploader.yaml
+++ b/vich/uploader-bundle/1.13/config/packages/vich_uploader.yaml
@@ -1,0 +1,8 @@
+vich_uploader:
+    db_driver: orm
+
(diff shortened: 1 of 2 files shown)


garak:
I don't get the exception that complains about symfony/translation missing. The bundle is requiring symfony/string, it should be enough...

fabpot (approved):

PR #1430, opened 2022-07-29
eduandebruijne/admin-bundle recipe

eduandebruijne:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [eduandebruijne/admin-bundle](https://packagist.org/packages/eduandebruijne/admin-bundle)


eduandebruijne:
Works like a charm

eduandebruijne:
@fabpot or someone else, what can I do to get this recipe merged? :) Thanks!

eduandebruijne/admin-bundle/9.1/manifest.json
```diff
@@ -0,0 +1,19 @@
...
+    },
+    "composer-scripts": {
+        "assets:install %PUBLIC_DIR%": "symfony-cmd",
+        "cache:clear": "symfony-cmd"
+    },
+    "env": {
+        "ADMIN_PATH": "admin",
+        "MEDIA_PATH": "/",
+        "GOOGLE_CLIENT_ID": "",
+        "GOOGLE_CLIENT_SECRET": ""
```
fabpot, line 17: I would prefix all these env vars to avoid any clash in case projects also have the same names (ADMIN_PATH looks very generic for instance).

fabpot (changes requested):

fabpot (changes requested):
All the PHP classes should be removed.
Entities won't work if the user is using PHP attributes instead of annotations for instance.
Such classes should be generated by the bundle or the maker bundle. Probably a step to add in a post install text.

eduandebruijne/admin-bundle/9.1/manifest.json
eduandebruijne: Added `EDB_` as prefix

eduandebruijne:
I removed all the PHP classes and added some installation instructions in the bundle with a hint in post-install text :+1:

fabpot (approved):

PR #1432, opened 2022-09-05
Adding 5.0

ThomasLandauer:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/codeception/codeception

Still missing:
This should be added to `composer.json`, right?:
```json
"autoload-dev": {
    "psr-4": {
        "Tests\\": "tests/"
    }
},
```


fabpot (changes requested):
Duplicate of #1435. See my comments there. Let's close ones of the PRs.

codeception/codeception/5.0/tests/Support/Helper/Acceptance.php
```diff
@@ -0,0 +1,10 @@
+<?php
+namespace Tests\Support\Helper;
+
+// here you can define custom actions
+// all public methods declared in helper class will be available in $I
+
+class Acceptance extends \Codeception\Module
```
Naktibalda, line 7: Please delete helper files.
Codeception 5 don't generate them automatically anymore because they are rarely used.
If user wants to use a helper, they can generate it using `./vendor/bin/codecept generate:helper` command.

Naktibalda:
I am maintainer of Codeception.

This pull request is a bit better because it correctly uses `.yml` extension for all `.suite.yml` files (Codeception doesn't use .yaml extension) and enables Symfony module in Functional.suite.yml file.
Also it correctly enables Helper files in module configuration.

Tester files must be added, but Helper files aren't necessary, I guess that they were added to reduce differences from the previous version of this recipe.

There is very little original work in either pull request  (manifest.json).
Other files can be generated by running `vendor/bin/codecept bootstrap` command.
`post-install.txt` file contains output of `codecept bootstrap` command.

Symfony module is no longer a part of `codeception/codeception` package since 4.0.
Is it possible to make `codeception/module-symfony` a dependency of this recipe or add suggestion to install it to post-install.txt ?

ThomasLandauer:
> Also it correctly enables Helper files in module configuration.

Is it OK to enable them, even if they don't exist? (as we just deleted them)

> Is it possible to make codeception/module-symfony a dependency of this recipe or add suggestion to install it to post-install.txt ?

A somewhat stupid question: This recipe is run for *all* Codeception installations, right? (Not just for Symfony users?) Then suggesting Symfony module is probably not the best idea...

From above: This needs to be added to `composer.json`, right?:
```json
"autoload-dev": {
    "psr-4": {
        "Tests\\": "tests/"
    }
},
```
Can this be done through this recipe or someway else?

TavoNiievez:
I'm [**`codeception/module-symfony`**](https://github.com/Codeception/module-symfony) maintainer.

I've submitted a PR that replaces the current one in #1583:

- I've added @ThomasLandauer as a co-author in the commit.
- Helpers are not necessary for a first run, it is better not to add them by default unless we want another first run error for the user.
- Running `vendor/bin/codecept run` already suggests the user to install the modules that have been imported but are not installed through composer. That's the way it has worked up to this point and I don't consider it a 5.0 recipe blocker, we can move on.
- This recipe is only for Symfony installations, not other frameworks. So I have taken the liberty to remove the previous step 1 in `post-install.txt` as the default url given corresponds to the default url in Symfony.
- Finally, there's no need for the test namespace to be named `Tests`, since we can not modify the `composer.json` file to add the patch that @ThomasLandauer suggests, the simplest thing is to leave things as they were and that `App/Tests` remains the default namespace in this recipe.

cc. @Naktibalda , @fabpot

ThomasLandauer:
OK, thanks @TavoNiievez, then I'm closing here.

PR #1436, opened 2022-09-27
adding recipe for networking/init-cms-bundle v5.4

ychadwick:
| Q             | A
| ------------- | ---
| License       | MIT

### DIFF
--- a/networking/init-cms-bundle/5.4/config/packages/dev/networking_init_cms.yaml
+++ b/networking/init-cms-bundle/5.4/config/packages/dev/networking_init_cms.yaml
@@ -1,0 +1,3 @@
+networking_init_cms:
+    cache:
+        activate: false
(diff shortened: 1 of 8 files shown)


ychadwick:
This bundle is not yet ready for symfony 6 due to one of the dependent packages which is not compatible wit symfony 6

ychadwick:
Bundle is ready for symfony 6, however I get a weird error when stablity in composer set to dev.

OskarStark:
Does it work now @ychadwick?

fabpot (changes requested):
All the PHP classes should be removed.
Entities won't work if the user is using PHP attributes instead of annotations for instance.
The bundle or the maker bundle should generate such classes. Probably a step to add in a post install text.

PR #1440, opened 2022-10-04
Sonata Page Bundle 4

eerison:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/sonata-project/page-bundle

### DIFF
--- a/sonata-project/page-bundle/4.0/config/packages/sonata_page.yaml
+++ b/sonata-project/page-bundle/4.0/config/packages/sonata_page.yaml
@@ -8,6 +8,7 @@ sonata_admin:
         extra_stylesheets:
             - bundles/sonatapage/app.css
 
+# https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/advanced_configuration/
 sonata_page:
     multisite: host
     use_streamed_response: false
@@ -31,6 +32,7 @@ sonata_page:
     ignore_uri_patterns:
         - ^/admin\/   # ignore admin route, ie route containing 'admin'
 
+    #https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/page_composer/
     default_template: default
     templates:
         default:
@@ -72,9 +74,3 @@ sonata_page:
     catch_exceptions:
         not_found: [404]    # render 404 page with "not_found" key (name generated: _page_internal_error_{key})
         fatal:     [500]    # so you can use the same page for different http errors or specify specific page for each error
-
-    class:
-        page: 'App\Entity\SonataPagePage'
-        snapshot: 'App\Entity\SonataPageSnapshot'
-        block: 'App\Entity\SonataPageBlock'
-        site: 'App\Entity\SonataPageSite'
--- a/sonata-project/page-bundle/4.0/config/packages/sonata_page.yaml
+++ b/sonata-project/page-bundle/4.0/config/packages/sonata_page.yaml
@@ -1,0 +1,80 @@
+sonata_block:
+    default_contexts: [sonata_page_bundle]
+
+sonata_admin:
+    assets:
+        extra_javascripts:
+            - bundles/sonatapage/app.js
+        extra_stylesheets:
+            - bundles/sonatapage/app.css
+
+sonata_page:
+    multisite: host
+    use_streamed_response: false
+
+    router_auto_register:
+        enabled: true
+        priority: 150
+    ignore_route_patterns:
+        - ^(.*)admin(.*)   # ignore admin route, ie route containing 'admin'
+        - ^_(.*)          # ignore symfony routes
+    ignore_routes:
+        - sonata_page_cache_esi
+        - sonata_page_cache_ssi
+        - sonata_page_js_sync_cache
+        - sonata_page_js_async_cache
+        - sonata_cache_esi
+        - sonata_cache_ssi
+        - sonata_cache_js_async
+        - sonata_cache_js_sync
+        - sonata_cache_apc
+    ignore_uri_patterns:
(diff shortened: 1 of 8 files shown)


sonata-project/page-bundle/4.0/src/Entity/SonataPageSnapshot.php
```diff
@@ -0,0 +1,29 @@
...
+ *     })
+ * })
+ */
+class SonataPageSnapshot extends BaseSnapshot
+{
+    /**
+     * @ORM\Id
+     * @ORM\Column(type="integer")
+     * @ORM\GeneratedValue(strategy="AUTO")
+     * // Serializer\Groups(groups={"sonata_api_read","sonata_api_write","sonata_search"})
```
jordisala1991, line 24:
```suggestion
```

sonata-project/page-bundle/4.0/src/Entity/SonataPagePage.php
```diff
@@ -0,0 +1,25 @@
...
+
+namespace App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Sonata\PageBundle\Entity\BasePage;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="page__page")
+ * @ORM\HasLifecycleCallbacks
```
jordisala1991, line 13: I don't think adding this is necessary. Did you find any problem without it?

jordisala1991: I don't remember having this line, and the life cycle callbacks were executed anyway.

eerison: the only thing that I found that probably use it, is: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Entity/BaseBlock.php#L21

But it works without `HasLifecycleCallbacks`

probably those line make this prePersist and update works: https://github.com/sonata-project/SonataPageBundle/blob/4.x/src/Resources/config/doctrine/BaseBlock.orm.xml#L11-L14


Then ok I'll remove it ;)

sonata-project/page-bundle/4.0/src/Entity/SonataPageSnapshot.php
eerison: Done

sonata-project/page-bundle/4.0/config/packages/sonata_page.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+    catch_exceptions:
+        not_found: [404]    # render 404 page with "not_found" key (name generated: _page_internal_error_{key})
+        fatal:     [500]    # so you can use the same page for different http errors or specify specific page for each error
+
+    class:
+        page: 'App\Entity\SonataPagePage'
+        snapshot: 'App\Entity\SonataPageSnapshot'
+        block: 'App\Entity\SonataPageBlock'
+        site: 'App\Entity\SonataPageSite'
```
VincentLanglet, line 80: I dunno if we're supposed to add this in the recipe:
- It force using DoctrineORM
- It force the naming of the entities
- It used annotation on entities when people might prefer attributes
- It arbitrary added index on some entities

I dunno how similar bundle/recipe works for these situations.

eerison:
> > Hey @jordisala1991
> > for page bundle works I need to add [SonataDoctrineOrmAdminBundle](https://docs.sonata-project.org/projects/SonataDoctrineORMAdminBundle/en/3.x/)
> > https://docs.sonata-project.org/projects/SonataPageBundle/en/4.x/reference/installation/
> > do you know anyway to check if this dependency was added if not ask if the user wants to install?
> 
> IMHO, it's the user job to require DoctrineORM so you don't need anything else.

What I was thinking is , do something to check if Orm or odm dependency exist, if not we add the dependency!

sonata-project/page-bundle/4.0/config/packages/sonata_page.yaml
VincentLanglet: It's not "necessary". I mean people can use another entity instead.
I'm not familiar about recipe and what should be provided or not.
Dunno if we can find a similar library recipe...

VincentLanglet (approved):

sonata-project/page-bundle/4.0/config/packages/sonata_page.yaml
eerison: should be good the recipe help the user make the sonata page bundle works without do too much!

I know it's used for ORM, But ATM I don't have any solution to make this works for both, and if I comment or remove this configuration, it will be even worst, because it will broken to ORM and ODM, then IMO we could let this like this and in the feature we can provide a new solution!

jordisala1991 (approved):

eerison:
Screenshot 2022-10-26 at 09 59 04


Hey @VincentLanglet and @jordisala1991 , after the user install Sonata Page bundle, it'll be the next steps to execute to have this working! What do you think?

VincentLanglet:
That's nice, let see if someone can merge this

VincentLanglet (approved):

eerison:
Screenshot 2022-10-26 at 10 39 54


The current one :D

eerison:
@jordisala1991 could you give your approval again pleases 🙏

VincentLanglet:
> @jordisala1991 could you give your approval again pleases 🙏

He's really busy atm ; and the approval will not really help moving this PR forward.
I think you just need someone with the rights to merge PR in this repository.

fabpot (approved):

PR #1444, opened 2022-10-19
phpzlc v2

CJayHe:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |

### DIFF
--- a/phpzlc/area-business/1.0/manifest.json
+++ b/phpzlc/area-business/1.0/manifest.json
@@ -1,0 +1,10 @@
+{
+    "copy-from-package": {
+        "routing/": "config/routing/",
+        "Business/": "src/Business/",
+        "Controller/": "src/Controller/",
+        "Entity/": "src/Entity/",
+        "Repository/": "src/Repository/",
+        "DataFixtures/": "src/DataFixtures/"
+    }
+}
--- a/phpzlc/phpzlc/2.0/manifest.json
+++ b/phpzlc/phpzlc/2.0/manifest.json
@@ -1,0 +1,17 @@
+{
+    "bundles": {
+        "PHPZlc\\PHPZlc\\Bundle\\PHPZlcBundle": ["all"]
+    },
+    "copy-from-package": {
+        "Doctrine/ORM/Rewrite/Templates/Repository.tpl.php": "vendor/symfony/maker-bundle/src/Resources/skeleton/doctrine/Repository.tpl.php",
+        "Doctrine/ORM/Rewrite/Hydration/AbstractHydrator.php": "vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/AbstractHydrator.php",
+        "Doctrine/ORM/Rewrite/Hydration/ObjectHydrator.php": "vendor/doctrine/orm/lib/Doctrine/ORM/Internal/Hydration/ObjectHydrator.php",
+        "Doctrine/ORM/Rewrite/MakeEntityRegenerate/ClassSourceManipulator.php": "vendor/symfony/maker-bundle/src/Util/ClassSourceManipulator.php",
+        "Doctrine/ORM/Rewrite/MakeEntityRegenerate/EntityRegenerator.php": "vendor/symfony/maker-bundle/src/Doctrine/EntityRegenerator.php",
+        "Bundle/phpzlc.yaml": "config/packages/phpzlc.yaml",
+        "PROJECT_README.md":"README.md"
+    },
+    "copy-from-recipe": {
+        "Business/":"src/Business/"
+    }
+}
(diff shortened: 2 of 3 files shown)


phpzlc/admin-business/2.0/manifest.json
```diff
@@ -0,0 +1,11 @@
+{
+    "copy-from-package": {
+        "routing/": "config/routing/",
+        "Business/": "src/Business/",
+        "Controller/": "src/Controller/",
+        "Entity/": "src/Entity/",
+        "Repository/": "src/Repository/",
+        "templates/": "templates/",
+        "DataFixtures/": "src/DataFixtures/"
+    }
```
fabpot, line 10: None of these directories exist, so the whole file can be removed.

phpzlc/area-business/2.0/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+    "copy-from-package": {
+        "routing/": "config/routing/",
+        "Business/": "src/Business/",
+        "Controller/": "src/Controller/",
+        "Entity/": "src/Entity/",
+        "Repository/": "src/Repository/",
+        "DataFixtures/": "src/DataFixtures/"
+    }
+}
```
fabpot, line 10: Same here

phpzlc/phpzlc/2.4/manifest.json
```diff
@@ -0,0 +1,17 @@
+{
+    "bundles": {
+        "PHPZlc\\PHPZlc\\Bundle\\PHPZlcBundle": ["all"]
+    },
+    "copy-from-package": {
+        "Doctrine/ORM/Rewrite/Templates/Repository.tpl.php": "vendor/symfony/maker-bundle/src/Resources/skeleton/doctrine/Repository.tpl.php",
```
fabpot, line 6: What if Marker bundle is not installed?

fabpot (changes requested):
I have a hard figuring out what you're trying to do here.
A recipe is mainly helpful when we want to add some configuration files.
We tend to avoid adding PHP classes as this is more for maker bundle.

phpzlc/admin-business/2.0/manifest.json
CJayHe: The purpose of copying here is to copy some written business code to the corresponding location of the project. These codes can be easily reused and adjusted according to the current project.

phpzlc/phpzlc/2.4/manifest.json
CJayHe: composer package ` phpzlc/phpzlc` will decide who to rely on

phpzlc/admin-business/2.0/manifest.json
fabpot: Oh sorry, I read too fast, that's a copy from **package**, not **recipe**.
I think that's something that will work long-term.
Just one example. Your entities are probably using annotations **or** attributes. Depending on what the user has configured, it will work or not.
So, instead, you should probably have a command that copies/generates/customizes those files. This command to be run can be advertized in a post install message (that's a recommendation I've made in the last few weeks in many similar PRs here).

fabpot (changes requested):

phpzlc/admin-business/2.0/manifest.json
CJayHe: I know what you mean, and I've thought about it too, but I think it will make the installation more complicated and inconvenient.
1.  Almost all of my packages need to do this, which means I have to execute extra commands every time.
2.  Executing copy is the same thing as symfony-flex, which means I have developed symfony-flex repeatedly.
3.  symfony-flex is great in my opinion.  It was also intended to be easy to install.
4.  These configurations of mine have to work, there is no alternative.

fabpot: Then, I fear that your packages are not a good fit for recipes :(
We want recipes that work for all use cases, not just one specific one.
To generate files, we have maker bundle, that's probably what you want to extend here.
Recipes are a way to configure things.
Another issue with copying PHP files that need to be heavily changed by the user: when one wants to update their recipes, they will get plenty of conflicts.

CJayHe: Thank you for your guidance.
But for now in my scenario, using symfy-flex is the best solution.
1.   If you change the method, you need to write a component like symfy-flex.   The functionality written is not as good as the Symfony-Flex experience, which as a symfony user I don't think is cool.
2.   Secondly, I think even if the method is changed, the problem will not be reduced, because the principle and purpose of the two are the same, so the situation is similar.

CJayHe: Thank you for your guidance.
But for now in my scenario, using symfy-flex is the best solution.
1.   If you change the method, you need to write a component like symfy-flex.   The functionality written is not as good as the Symfony-Flex experience, which as a symfony user I don't think is cool.
2.   Secondly, I think even if the method is changed, the problem will not be reduced, because the principle and purpose of the two are the same, so the situation is similar.

PR #1456, opened 2022-11-24
Recipe for wedrix/watchtower

Wedrix:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/wedrix/watchtower


Wedrix:
Intended for Symfony 6.1

OskarStark:
The PHP code should be part of the package, not the recipe IMHO.
A recipe is only meant for configuration of the package

Wedrix:
> The PHP code should be part of the package, not the recipe IMHO. A recipe is only meant for configuration of the package

Absolutely. I'll work on a proper bundle soon.

fabpot:
What about closing it for now as we won't merge this PR which contains mostly PHP code that should be part of a bundle instead?

fabpot (changes requested):
Recipes should not contain so many PHP classes. These should probably be moved to the bundle itself or generated by a command from the bundle.

Wedrix:
Closing for now.

PR #1473, opened 2023-01-08
Add recipe for numberninecms/cms 0.5

williarin:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/numberninecms/cms

Recipe for [numberninecms/cms](https://github.com/numberninecms/cms) 0.5


weaverryan:
Hi! Check the CI - it fails with: Uncaught Error: Class "NumberNine\NumberNineBundle" not found

Cheers!

williarin:
Hi! Yes I fixed it, can the job be re-run?

numberninecms/cms/0.5/manifest.json
```diff
@@ -0,0 +1,15 @@
...
+    "bundles": {
+        "NumberNine\\NumberNineBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "copy-from-package": {
+        "install/docker/": "docker/",
+        "install/public/": "%PUBLIC_DIR%/"
+    },
```
vvasiloi, line 11: Hey @williarin, I'm a bit confused about these instructions. I looked in the package and from the `docker` directory it copies some config files for NGINX and PHP, but those are useless without the Docker Compose configuration.
What's the idea behind it?

williarin: Hi, there's a docker-compose.yml in the bundle repo but it cannot be used in the recipe as it would be incompatible with the one Symfony creates. The yml is either copied by the installer (separate repo) or by a console command I didn't make yet.

vvasiloi: You could use the `docker-compose` configurator. Here's an example: https://github.com/symfony/recipes/blob/296e45d39a760ff9aca2fb9ffa4f6fa6a4b9bb99/doctrine/doctrine-bundle/2.4/manifest.json#L23-L39

williarin: Can the configurator delete a service created by another recipe? I didn't use this method because I need doctrine-bundle which creates this Postgre service, but also need MySQL and not Postgre.

vvasiloi: No, it cannot delete, but AFAIK it works per service, if you define another `database` service as Doctrine does, then it will update its config.

vvasiloi: And to maintain the compatibility, you can add a `mysql` alias to the `database` service.

williarin: I'll try that, thank you.

williarin: I updated the recipe with the docker-compose configurator. It creates duplicate services and volumes instead of replacing them, but it's not a big deal after all as the installer takes care of cleaning the file.

chalasr:
looks like the current build didn't run, please force-push again to trigger a new one

williarin:
Hello, thank you for noticing this problem! I forgot about it. Before I do that, is there a way to completely remove this recipe?
I'd like to use https://github.com/williarin/cook instead of Flex for recipe management but I'm not sure how to proceed for unregistering it from here.

chalasr:
You're welcome.
To remove a recipe, you have to submit a pull request removing the directory corresponding to the bundle for which the recipe is for. If the vendor only has 1 recipe (the removed one), then remove the directory corresponding to the vendor name as well (numberninecms here).

On a side note, Symfony Flex is widely deployed in the Symfony ecosystem so if adoption is something important for your package, keeping it on flex may be interesting.

fabpot:
Closing as there is no more activity.

PR #1479, opened 2023-01-17
prooph event store recipes

unixslayer:
event store interface should alias instance which wraps the internal event store client 
wrapper gets registered by bundle package due it is the proper place to define an alias

| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/prooph/event-store-symfony-bundle https://packagist.org/packages/prooph/pdo-event-store

### DIFF
--- a/prooph/pdo-event-store/1.15/manifest.json
+++ b/prooph/pdo-event-store/1.15/manifest.json
@@ -1,0 +1,13 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "copy-from-package": {
+        "scripts/": "%CONFIG_DIR%/scripts/"
+    },
+    "env": {
+        "EVENT_STORE_DSN": "pgsql:host=127.0.0.1;dbname=event_streams",
+        "EVENT_STORE_USER": "user",
+        "EVENT_STORE_PASSWORD": "password"
+    }
+}
--- a/prooph/pdo-event-store/1.15/post-install.txt
+++ b/prooph/pdo-event-store/1.15/post-install.txt
@@ -1,0 +1,4 @@
+  * Modify your EVENT_STORE_* configuration in <fg=green>.env</>
+
+  * Create event streams and projections tables using SQL scripts
+    in <comment>%CONFIG_DIR%/scripts/*</comment>
--- a/prooph/event-store-symfony-bundle/0.10/config/packages/prooph_event_store.yaml
+++ b/prooph/event-store-symfony-bundle/0.10/config/packages/prooph_event_store.yaml
@@ -1,0 +1,11 @@
+prooph_event_store:
+    stores:
+        default:
+            event_store: 'app.event_store.default'
+
+services:
+    _defaults:
+        public: false
+
+    Prooph\EventStore\EventStore: '@prooph_event_store.default'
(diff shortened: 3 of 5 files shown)


prooph/pdo-event-store/1.15/post-install.txt
```diff
@@ -0,0 +1,4 @@
+  * Modify your MYSQL_* configuration in <fg=green>.env</>
```
fabpot, line 1: Do you support PostgreSQL? If yes, it should be used instead (as this is what we're using by default in recipes.)

prooph/event-store-symfony-bundle/0.10/src/Command/CreateEventStreamCommand.php
```diff
@@ -0,0 +1,36 @@
...
+namespace App\Command;
+
+use Prooph\EventStore\EventStore;
+use Prooph\EventStore\Stream;
+use Prooph\EventStore\StreamName;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class CreateEventStreamCommand extends ContainerAwareCommand
```
fabpot, line 14: This class should probably be part of the bundle, not the recipe, right?

fabpot (changes requested):

prooph/pdo-event-store/1.15/post-install.txt
unixslayer: Yes, Prooph supports PostgreSQL. I'll change that.

prooph/event-store-symfony-bundle/0.10/src/Command/CreateEventStreamCommand.php
unixslayer: Not necessarily. Bundle helps to connect Symfony application with EventStore which can be managed somewhere else. This class is an example of how it can be done and is based on fixed data due most likely it is modified after bundle is installed. Making it part of a bundle is possible however it would require some changes in bundle itself.

unixslayer: I can plan those changes and eventually remove this class from the recipe. Although, this PR main focus is to fix the initial configuration which is, at the moment, not entirely correct after the bundle is installed.

fabpot: Let's remove this class for now then. I suppose there is some docs about this anyway.

prooph/pdo-event-store/1.15/manifest.json
```diff
@@ -0,0 +1,13 @@
+{
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "copy-from-package": {
+        "scripts/": "%CONFIG_DIR%/scripts/"
+    },
+    "env": {
+        "POSTGRES_DSN": "pgsql:host=127.0.0.1;dbname=event_streams",
```
fabpot, line 9: Sorry I missed that earlier, but you should probably prefix the env vars with something specific to your bundle.

PR #1480, opened 2023-01-30
phpzlc 2.5

CJayHe:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |

### DIFF
--- a/phpzlc/flex/2.0/manifest.json
+++ b/phpzlc/flex/2.0/manifest.json
@@ -1,0 +1,8 @@
+{
+    "bundles": {
+        "PHPZlc\\Flex\\Bundle\\FlexBundle": ["dev", "test"]
+    },
+    "copy-from-package": {
+        "Bundle/phpzlc-flex.yaml":  "config/packages/phpzlc-flex.yaml"
+    }
+}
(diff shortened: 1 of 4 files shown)


phpzlc/flex/2.0/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+    "bundles": {
+        "PHPZlc\\Flex\\Bundle\\FlexBundle": ["dev", "test"]
+    },
+    "copy-from-package": {
+        "Bundle/phpzlc-flex.yaml":  "config/packages/phpzlc-flex.yaml"
```
fabpot, line 6: Most (all?) recipes actually have their configuration in this repo instead of the package.
Maybe we should do the same here for better consistency?
I don't have any strong opinion.

CJayHe: I think configuration files are likely to be subject to frequent changes, and pr needs to be committed for each change. It doesn't look easy. Of course I would like to align myself with the community. What do you think I should do.

fabpot: The configuration is copied once at project creation. So, the frequent updates won't be visible to current projects. But if the configuration is here, then, recipes:update will get the new updates and propose it to current projects. So, I think having it here is much better from a UX perspective.

fabpot (approved):

PR #1488, opened 2023-02-27
Feature/mock server bundle recipe 1.0

yits-alexharpin:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/youritservices/mock-server-bundle

### DIFF
--- a/youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
+++ b/youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
@@ -1,0 +1,3 @@
+mock_server:
+    host_url: '%env(MOCK_SERVER_HOST)%'
+    mocks_directory: '%kernel.project_dir%/tests/apiMocks'
--- a/youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
+++ b/youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
@@ -1,3 +1,6 @@
+framework:
+    http_client:
+
 mock_server:
     host_url: '%env(MOCK_SERVER_HOST)%'
     mocks_directory: '%kernel.project_dir%/tests/apiMocks'
--- a/youritservices/mock-server-bundle/1.0/manifest.json
+++ b/youritservices/mock-server-bundle/1.0/manifest.json
@@ -1,0 +1,11 @@
+{
+    "bundles": {
+        "YourITServices\\MockServerBundle\\MockServerBundle": ["dev", "test"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "MOCK_SERVER_HOST": "http://mock-api:1080"
+    }
+}


yits-alexharpin:
@fabpot @weaverryan Any chance of a review on this one, when you've got time please.  I know these aren't are as high priority as normal work, so just when you've got chance.

fabpot (approved):

youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+    http_client:
```
alexander-schranz, line 2: This looks a little bit unexpected for me. Maybe the bundle should use the `prepend extension` here to set the `http_client` config: https://symfony.com/doc/current/bundles/prepend_extension.html

Else I'm unaware how this will be merged with project custom `http_client` configuration? /cc @fabpot

youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
```diff
@@ -0,0 +1,6 @@
...
+
+mock_server:
+    host_url: '%env(MOCK_SERVER_HOST)%'
+    mocks_directory: '%kernel.project_dir%/tests/apiMocks'
```
alexander-schranz, line 6: Would this not fail on `prod` and other non dev, test envs. think this should be:

```yaml
when@test: &test
    mock_server:
        host_url: '%env(MOCK_SERVER_HOST)%'
        mocks_directory: '%kernel.project_dir%/tests/apiMocks'

when@dev: *test
```

youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
fabpot: Sorry, my fault. It cannot work, I will revert this PR.

fabpot:
Reverted

youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
yits-alexharpin: @alexander-schranz @fabpot  I'll have a look into using that instead.  The original reason for adding this config in the first place, is that without it, the http_client wasn't being registered as a service when the bundle is added as a dev only dependency, which is the only environment it's intended to be used (it's after all designed to load API mocks for testing, not used in production).  I had presumed originally that the http_client service would be available purely via default  config / behaviour, as the project I was testing it in didn't have anything specifically configured, but this doesn't appear to be the case.

youritservices/mock-server-bundle/1.0/config/packages/mock_server.yaml
yits-alexharpin: @alexander-schranz Is this normally something that's required if the package is only used in dev and test and not prod?   If that's the case I'll add add the config for that and open a new PR.  The package is only intended for non-prod environment, as I couldn't see a reason you would want to mock remote API calls in a prod deployment.

alexander-schranz: the config is always loaded and if the bundle is only registered in `dev` and `test` it will show you an error. that not find who is responsible for `mock_server`:

Run:

```
bin/console cache:clear --env prod
```

or switch to `Prod` env. 

```bash
In FileLoader.php line 174:

  There is no extension able to load the configuration for "mock_server" (in "/tmp/project/config/packages/mock_server.yaml"). 
```

yits-alexharpin: @alexander-schranz Thanks for the above.  I was only testing this in a current project that's in not in prod at the moment, I'll get that added to the config and keep that in mind for next time.

PR #1517, opened 2023-05-29
Add mercuryseries/inertia-bundle recipe

mercuryseries:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [https://packagist.org/packages/mercuryseries/inertia-bundle](https://packagist.org/packages/mercuryseries/inertia-bundle)


mercuryseries/inertia-bundle/0.1/config/packages/mercuryseries_inertia.yaml
```diff
@@ -0,0 +1,38 @@
...
+    # | You can specify a custom SSR bundle path, or omit it to let Inertia
+    # | try and automatically detect it for you.
+    # |
+    # | Do note that enabling these options will NOT automatically make SSR work,
+    # | as a separate rendering service needs to be available. To learn more,
+    # | please visit https://inertiajs.com/server-side-rendering
+    ssr:
+        enabled: false
+        url: 'http://127.0.0.1:13714/render'
+        bundle: '%kernel.project_dir%/public/build-ssr/ssr.js'
```
weaverryan, line 18: Where does this file come from? Does the bundle make this available? Or is it something the user should add?

mercuryseries/inertia-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+    "bundles": {
+        "MercurySeries\\Bundle\\InertiaBundle\\MercurySeriesInertiaBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "gitignore": ["/%PUBLIC_DIR%/build-ssr/"],
+    "add-lines": [
```
weaverryan, line 9: Instead of `add-lines`, you should ship a new `config/routes/intertia.yaml` routes file with the routes inside. This configurator is meant as a last resort - but there is a built-in standard already for adding new routing files

weaverryan (changes requested):
A few minor questions/changes

mercuryseries/inertia-bundle/0.1/config/packages/mercuryseries_inertia.yaml
mercuryseries: The user needs to execute the `npm run build` command in order to generate the file. The presence of this file is already managed on the bundle side: https://github.com/mercuryseries/inertia-bundle/blob/main/src/Command/StartInertiaSsrCommand.php#L46-L59.

mercuryseries/inertia-bundle/0.1/manifest.json
mercuryseries: Done @weaverryan  thanks!
Is there a way to automatically detect which part to uncomment: https://github.com/symfony/recipes-contrib/blob/32f43347b449fa0d8c35d013621f86d3dbc0f91a/mercuryseries/inertia-bundle/0.1/config/routes/inertia.yaml?

mercuryseries/inertia-bundle/0.1/config/routes/inertia.yaml
```diff
@@ -0,0 +1,19 @@
...
+#             require: ['POST', 'PUT', 'PATCH', 'DELETE']
+
+# If you use annotations
+# controllers:
+#     resource: ../../src/Controller/
+#     type: annotation
+#     defaults:
+#         csrf:
+#             create: true
+#             require: ['POST', 'PUT', 'PATCH', 'DELETE']
```
weaverryan: You only need one of these - iirc, `type: annotations` will work for both.

However, if I'm understanding this correctly, then what you are REALLY doing is just adding the `defaults: csrf:` part to the user's existing route import, right? In that case, my previous comment was wrong... and you may indeed need to use `add-lines` to add this to the existing `config/routes.yaml`.

mercuryseries: Thanks @weaverryan for the feedback. Really appreciated!
I finally decided to handle this on the bundle side. So I think it can be merged now.

PR #1527, opened 2023-07-04
Add phpstan/phpstan recipe

gremo:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/phpstan/phpstan

### DIFF
--- a/phpstan/phpstan/1.0/phpstan.neon
+++ b/phpstan/phpstan/1.0/phpstan.neon
@@ -1,0 +1,8 @@
+parameters:
+    level: 6
+    paths:
+        - bin/
+        - config/
+        - public/
+        - src/
+        - tests/
--- a/phpstan/phpstan/1.0/manifest.json
+++ b/phpstan/phpstan/1.0/manifest.json
@@ -1,0 +1,5 @@
+{
+    "copy-from-recipe": {
+        "phpstan.neon": "phpstan.neon"
+    }
+}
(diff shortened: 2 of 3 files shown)


phpstan/phpstan/1.0/phpstan.neon
```diff
@@ -0,0 +1,8 @@
+parameters:
+    level: 6
+    paths:
+        - bin/
+        - config/
+        - public/
```
weaverryan, line 6: I'm not sure about `public/` and `bin/`, though I suppose they don't hurt.

phpstan/phpstan/1.0/phpstan.neon
```diff
@@ -0,0 +1,8 @@
+parameters:
+    level: 6
```
weaverryan, line 2: Why 6?

gremo:
> Having a recipe for this may make sense, but usually the PR would come from a maintainer of the project, or at least approved by one.

I see, but I dont’t think that the phpstan team have the interest or time to look at this. I can open an issue in their repository linking this PR if you want. The receipe is pretty trivial by the way!

phpstan/phpstan/1.0/phpstan.neon
gremo: Same as https://github.com/symfony/recipes-contrib/blob/main/squizlabs/php_codesniffer/3.6/phpcs.xml.dist#L13

phpstan/phpstan/1.0/phpstan.neon
gremo: Default 0 means almost no checks. Max is 10 and it’s even too strictly. The level 6 is on their website in the getting started guide, so a new developer would probably start with that level.

fabpot (approved):

norkunas:
Why not `phpstan.neon.dist` file was chosen for config? Because now it overrode my project's local config instead of dist config

PR #1536, opened 2023-08-24
Add recipe for luzrain/workerman-bundle

luzrain:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/luzrain/workerman-bundle

### DIFF
--- a/luzrain/workerman-bundle/0.1/config/packages/workerman.yaml
+++ b/luzrain/workerman-bundle/0.1/config/packages/workerman.yaml
@@ -1,12 +1,13 @@
 workerman:
-  servers:
-    - name: 'Symfony webserver'
-      listen: http://0.0.0.0:80
-      processes: 4
+    servers:
+        -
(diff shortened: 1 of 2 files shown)


luzrain:
Check failed because bundle requires implementation of the psr7/psr17 and depends only on PSR abstractions. User should manually install one of them, [nyholm/psr7](https://github.com/Nyholm/psr7) as an example.

luzrain/workerman-bundle/0.1/config/packages/workerman.yaml
```diff
@@ -0,0 +1,13 @@
+workerman:
+    servers:
+        -
+            name: 'Symfony webserver'
+            listen: http://0.0.0.0:80
+            processes: 4
+
+    relod_strategy:
```
OskarStark, line 8: is reload correct or should it be reload ?

fabpot (approved):

PR #1538, opened 2023-09-03
[phpstan/phpstan] phpstan.neon.dist config file usage

duboiss:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [https://packagist.org/packages/phpstan/phpstan](https://packagist.org/packages/phpstan/phpstan)

[https://phpstan.org/config-reference#config-file](https://phpstan.org/config-reference#config-file)
> The usual practice is to have phpstan.neon.dist or phpstan.dist.neon under version control, and allow the user to override certain settings in their environment (on their own computer or on a continuous integration server) by creating phpstan.neon that’s present in .gitignore file. See [Multiple files](https://phpstan.org/config-reference#multiple-files) for more details:


lcobucci:
Thanks @duboiss!
We're also having similar issues as @norkunas reported [here](https://github.com/symfony/recipes-contrib/pull/1527#issuecomment-1667188642) and this seems to be the solution.

@weaverryan @fabpot is there something we can help on getting this integrated?

fabpot (approved):

alexander-schranz:
I would suggest here to use `.dist.neon` file instead of `.neon.dist` as `.dist` files don't work well out of the box in different IDEs.

https://github.com/symfony/recipes-contrib/pull/1540

lcobucci:
> I would suggest here to use `.dist.neon` file instead of `.neon.dist` as `.dist` files don't work well out of the box in different IDEs: https://github.com/symfony/recipes-contrib/pull/1540

@alexander-schranz can you please share more details? I've been using `.neon.dist` for years without any issue... 🤔

alexander-schranz:
@lcobucci example PHPStorm (Neon Plugin installed) `neon.dist` requires manual configuration why `.dist.neon` works out of the box:

Bildschirmfoto 2023-09-11 um 19 25 50

lcobucci:
> @lcobucci example PHPStorm (Neon Plugin installed) `neon.dist` requires manual configuration why `.dist.neon` works out of the box:
> 
> Bildschirmfoto 2023-09-11 um 19 25 50
> 

Thanks for sharing, I think I'm so used to configure the `.dist` for other formats that I never realised 😅

I honestly don't have remarks beyond personal opinion (and many repos to change if Symfony decides to move to `.dist.neon`)

alexander-schranz:
> ... and many repos to change if Symfony decides to move to .dist.neon

That is already the case with this merge that many need to move from `phpstan.neon` to `phpstan.neon.dist` or `phpstan.dist.neon` because if the recipes update is executed the `phpstan.neon` file is added to the `.gitignore`, which are mostly more then the ones using dist files. So changing it now will have the same effects.

PR #1540, opened 2023-09-11
[phpstan/phpstan] Move phpstan.neon.dist to phpstan.dist.neon for better autocomplete

alexander-schranz:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | phpstan/phpstan

Why I understand that some want to overwrite it I suggest here to use `.dist.neon` instead of `.neon.dist` as that is better out of the box supported by different IDEs while `.neon.dist` mostly require manual configuration.

Related to: https://github.com/symfony/recipes-contrib/pull/1538


weaverryan (approved):
LGTM - https://github.com/phpstan/phpstan-src/pull/1006 has supported this for awhile

PR #1548, opened 2023-10-15
initial version for survos/command-bundle

tacman:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |  https://packagist.org/packages/survos/command-bundle

Add default configuration and routing to survos/command-bundle


survos/command-bundle/1.5/config/routes/survos_command.php
```diff
@@ -0,0 +1,16 @@
...
+use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
+
+return function (RoutingConfigurator $routes) {
+    $routes->add('survos_commands', '/admin/commands')
+        ->controller([CommandController::class, 'commands'])
+    ;
+
+    $routes->add('survos_command', 'admin/run-command/{commandName}')
+        ->controller([CommandController::class, 'runCommand'])
+    ;
```
weaverryan, line 14: It's pretty common to just have 1 route "import from inside the user's code, which then points to another file that adds the individual routes.

survos/command-bundle/1.5/config/packages/survos_command.php
```diff
@@ -0,0 +1,11 @@
...
+
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $container): void {
+    $container->extension('survos_command', [
+        'namespaces' => ['app', 'survos']
+    ]);
+};
+//survos_command:
+//namespaces: [app, project, survos]
```
weaverryan, line 11: I like PHP config too... though I'm not sure if we ever allow that in the recipes or not...

tacman: The PHP config seems to work as expected in the recipes, but I switch back to YAML since that's seems to be more common.  Although Tomas Votruba makes great arguments in https://tomasvotruba.com/blog/2020/07/16/10-cool-features-you-get-after-switching-from-yaml-to-php-configs

survos/command-bundle/1.5/config/routes/survos_command.php
tacman: I've made that change now, I think it follows the standard conventions.

tacman:
How do I set this PR up so that it's one PR for each bundle?  I have the command-bundle and crawler-bundle, should they each be in separate branches in my fork?

survos/command-bundle/1.5/config/routes/survos_command.php
```diff
@@ -0,0 +1,12 @@
+<?php
+// config/routes.php
+
+use Symfony\Component\Routing\Loader\Configurator\RoutingConfigurator;
+
+return static function (RoutingConfigurator $routes): void {
+    $routes->import('@SurvosCommandBundle/config/routes.php')
+        ->prefix('/admin') // consider adding this route to the access_control key in security
```
weaverryan, line 8:
```suggestion
        ->prefix('/admin') // consider adding this path to the access_control key in security
```

survos/command-bundle/1.5/config/routes/survos_command.php
```diff
@@ -0,0 +1,12 @@
...
+        ->schemes(['https'])
```
weaverryan, line 9: We don't usually enforce https at this level. The user should definitely have this, but it's on them.

survos/command-bundle/1.5/config/routes/survos_command.php
weaverryan: Btw, another option for you in the bundle is to, in your controller, always run a check for something like `ROLE_SURVOS_COMMAND`. Then tell the user to make sure their user's have this or put it in role_hierarchy. Then the controllers are locked down by default.

survos/command-bundle/1.5/config/routes/survos_command.php
```diff
@@ -0,0 +1,12 @@
+<?php
+// config/routes.php
```
weaverryan, line 2: Remove this command - and this is also more commonly written in YAML, not PHP

survos/command-bundle/1.5/config/routes/survos_command.php
tacman: interesting idea, I'll play with that for version 2 (which will support Symfony 7).

survos/command-bundle/1.5/config/routes/survos_command.php
tacman: done

tacman:
Done!  So I'll close this.

PR #1552, opened 2023-10-17
Add recipe for rekalogika/gotenberg-pdf-bundle

priyadi:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/rekalogika/gotenberg-pdf-bundle


rekalogika/gotenberg-pdf-bundle/1.0/config/packages/rekalogika_gotenberg_pdf.yaml
```diff
@@ -0,0 +1,2 @@
+rekalogika_gotenberg_pdf:
+    url: 'http://%env(GOTENBERG_HOST)%:%env(GOTENBERG_PORT)%'
```
weaverryan, line 2: Might it be simpler to have a single `GOTENBERG_URL` variable that is set to the full `https://...`? With the two variables, it doesn't let me use HTTP easily.

priyadi: That's what I had in mind, however, as far as I can tell, _HOST and _PORT are necessary for [docker integration](https://symfony.com/doc/current/setup/symfony_server.html#docker-integration).

But how about my latest commit?

weaverryan (approved):

PR #1557, opened 2023-10-26
Add recipe for simply-stream/twitch-api-bundle:0.1

aaricdev:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |  https://packagist.org/packages/simply-stream/twitch-api-bundle

Versions in packagist currently have the prefix "v" (it's my first contribution, will this interfere?) but I'll remove it in due time.


fabpot:
Can you fix the missing empty lines?
Having `v` as a prefix for your versions is not an issue.

aaricdev:
Should be added now, but it looks like I need to set it up on a clean environment to fix the last errors.

fabpot (approved):

PR #1561, opened 2023-11-10
Add omines/antispam-bundle

curry684:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/omines/antispam-bundle

### DIFF
--- a/omines/antispam-bundle/0.1/config/packages/antispam.yaml
+++ b/omines/antispam-bundle/0.1/config/packages/antispam.yaml
@@ -39,7 +39,6 @@ antispam:
 
 when@test:
     antispam:
-        # In automated tests the bundle and included components should by default run in passive mode,
-        # to ensure your form submissions will not get rejected due to being exactly what they are -
-        # an automated bot filling in your site forms in a distinctly non-human way.
-        passive: true
+        # In automated tests the bundle and included components are by default disabled. You can still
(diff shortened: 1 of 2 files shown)


curry684:
QA check on PHP 7.4 is failing because the bundle requires 8.1 or later.

omines/antispam-bundle/0.1/config/packages/antispam.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+            # Reject text fields that consist for more than 40% of Cyrillic (Russian) characters
+#            banned_scripts:
+#                scripts: [ cyrillic ]
+#                max_percentage: 40
+
+            # Reject fields that contain more than 3 URLs, or repeat a single URL more than once
+#            url_count:
+#                max: 3
+#                max_identical: 1
```
kbond, line 38:
```suggestion
#                max_identical: 1

when@test:
    antispam:
        # disable in tests
        passive: true
```

I know passive mode via config is not yet implemented but what about adding this for when it is?

curry684: That is actually a brilliant idea. Thnx!

kbond (approved):
Looks good!

PR #1569, opened 2023-11-24
Added alms/cycle-bundle

AlessandroAmos:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/alms/cycle-bundle


AlessandroAmos:
It seems that your package may not pass the second test because it runs only on PHP 8. It's a fork that you plan to maintain, integrating another ORM similar to Doctrine called Cycle ORM. The issue is that the cycle/orm package has PHP 8 as its limit.

alms/cycle-bundle/0.1/config/packages/cycle.yaml
```diff
@@ -0,0 +1,35 @@
+cycle:
+    dbal:
+        databases:
+            default:
+                connection: default_connection
+
```
fabpot, line 6: You should remove one of the newline.

fabpot (approved):

PR #1570, opened 2023-11-27
Use SymLinksIfOwnerMatch by default.

ostark:
| Q             | A
| ------------- | ---
| License       | MIT

`Options +SymLinksIfOwnerMatch` is more secure than `Options +FollowSymlinks`.

With the `FollowSymlinks` option enabled, you could easily mistakenly create a symlink to a file outside the web root folder and Apache would follow it. The owner of files for the web application usually differs from other system files. This additional owner check reduces the risk of exposing files.


frank-laemmer:
May I add some context:

We (fortrabbit) offer a standardized shared hosting solution with Apache. Our Symfony install [guide suggests to install the apache-pack](https://help.fortrabbit.com/install-symfony#toc-quick-start). That is not working any more, because of the open rule to follow any symlink.

We are aware that this may have an performance impact. Yet the increased security level might be worth it.

fabpot (approved):

PR #1572, opened 2023-11-29
Json rpc api v1.7

OtezVikentiy:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |


OtezVikentiy:
@fabpot Hello! Is there any opportunity to skip tests for symfony 5.4 with php 7.4? My bundle does not support these versions and is created extending the AbstractBundle as is written in the documentation. How can i solve this problem? Or to pass tests there is only one way - create backward compability with 5.4 version and php 7.4?

PR #1583, opened 2024-01-08
Codeception 5 update

TavoNiievez:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [codeception/codeception](https://packagist.org/packages/codeception/codeception)

### DIFF
--- a/codeception/codeception/5.0/codeception.yml
+++ b/codeception/codeception/5.0/codeception.yml
@@ -1,0 +1,19 @@
+namespace: App\Tests
+support_namespace: Support
+paths:
+    tests: tests
+    output: tests/_output
+    data: tests/Support/Data
+    support: tests/Support
+    envs: tests/_envs
+actor_suffix: Tester
+extensions:
+    enabled:
+        - Codeception\Extension\RunFailed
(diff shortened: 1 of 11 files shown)


codeception/codeception/5.0/codeception.yml
```diff
@@ -0,0 +1,19 @@
+namespace: App\Tests
```
W0rma, line 1: IIRC codeception itself uses `Tests` as default namespace (without `App\`).

Should the recipe use the same default namespace?

TavoNiievez: Hi @W0rma !

It was already discussed at https://github.com/symfony/recipes-contrib/pull/1432 , please read that thread.
In short, `Tests` is a **_should_**, not a _**must**_. If you find a way to use `Tests` without the user getting errors the first time he tries to run `vendor/bin/codecept run` you are welcome to modify the recipe.

W0rma: I see, thank you for the clarification!

fabpot (approved):

PR #1586, opened 2024-01-11
Update the Sentry recipe

javiereguiluz:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/sentry/sentry-symfony

Updated the recipe according to:

* https://docs.sentry.io/platforms/php/guides/symfony/performance/instrumentation/automatic-instrumentation/
* https://docs.sentry.io/platforms/php/guides/symfony/

~I know that `tracing.enabled: true` is the default option and we try to not add the default values to recipes to make them simpler. But this is a very important option that can impact the performance of your application, so I think it's fair to show this option prominently.~ Update: we set this option to `false` by default

### DIFF
--- a/sentry/sentry-symfony/4.6/config/packages/sentry.yaml
+++ b/sentry/sentry-symfony/4.6/config/packages/sentry.yaml
@@ -5,7 +5,7 @@ when@prod:
         # automatic instrumentation (there might be some performance penalty)
         # https://docs.sentry.io/platforms/php/guides/symfony/performance/instrumentation/automatic-instrumentation/
         tracing:
-            enabled: true
+            enabled: false
 
 #        If you are using Monolog, you also need this additional configuration to log the errors correctly:
 #        https://docs.sentry.io/platforms/php/guides/symfony/#monolog-integration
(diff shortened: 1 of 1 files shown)


sentry/sentry-symfony/4.6/config/packages/sentry.yaml
```diff
@@ -1,6 +1,11 @@
 when@prod:
     sentry:
         dsn: '%env(SENTRY_DSN)%'
+        # this hooks into critical paths of the framework (and vendors) to perform
+        # automatic instrumentation (there might be some performance penalty)
+        # https://docs.sentry.io/platforms/php/guides/symfony/performance/instrumentation/automatic-instrumentation/
+        tracing:
+            enabled: true
```
fabpot, line 8: I would set it to false by default as it impacts performance (and we've been hit by that significantly in the past for some of our projects).

```suggestion
            enabled: false
```

fabpot (changes requested):

sentry/sentry-symfony/4.6/config/packages/sentry.yaml
javiereguiluz: OK, I've just set it to `false` by default.

fabpot (approved):

PR #1611, opened 2024-04-26
add patchlevel/event-sourcing 3.0

DavidBadura:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/patchlevel/event-sourcing

A new release for patchlevel/event sourcing has been published.


DavidBadura:
The bundle only supports Symfony 6 and 7. And the test fails because there is no database running.

What can I do to make the CI happy?

chalasr:
Can you double-check that the recipe can be installed on a blank application (fresh install) without failing on the same error?

DavidBadura:
Yes, we also noticed yesterday that we had a problem with new installations. 
We fixed it this morning and now the DB connection is no longer needed. 👍

chalasr (approved):

PR #1617, opened 2024-05-17
Add kocal/biome-js-bundle recipe

Kocal:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/kocal/biome-js-bundle

This PR add the recipe for the fresh package https://github.com/Kocal/BiomeJsBundle :)

The "Run updated recipes" fails because the package does not support PHP 7.4, but works fine with PHP >= 8.1


fabpot (approved):

Kocal:
Thanks!

PR #1624, opened 2024-06-21
Feature/add gotenberg bundle

Neirda24:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/sensiolabs/gotenberg-bundle

This adds the minimal configuration required to use the sensiolabs/gotenberg-bundle package.

cc @Jean-Beru @StevenRenaux


sensiolabs/gotenberg-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,26 @@
...
+            "services": [
+                "gotenberg:",
+                "  image: 'gotenberg/gotenberg:8'"
+            ]
+        },
+        "compose.override.yaml": {
+            "services": [
+                "gotenberg:",
+                "  ports:",
+                "    - \"3000:3000\""
```
fabpot, line 22: In other services, we don't force the local port so that you can work on several projects with the same service in // without conflict.

fabpot (changes requested):

sensiolabs/gotenberg-bundle/0.1/manifest.json
Neirda24: The override file is serving that purpose am I wrong ? It is not required though to expose. Depends if developers are using full docker network or the Symfony cli. Let me know if you want us to remove it (on mobile ATM so can't check for sure but I think I picked this idea from another recipe...)

Neirda24: Fixed according to other examples in recipes.

fabpot (approved):

PR #1628, opened 2024-06-28
Rayenbou ticket bundle

RayenBou:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/rayenbou/ticket-bundle


rayenbou/ticket-bundle/1.0/config/packages/rayenbou_ticket.yaml
```diff
@@ -0,0 +1,7 @@
+rayen_bou_ticket:
+    authentication:
+        url: "%env(RAYENBOU_TICKET_URL)%"
+        username: "%env(RAYENBOU_TICKET_USERNAME)%"
+        password: "%env(RAYENBOU_TICKET_PASSWORD)%"
+    settings:
```
Neirda24, line 6: empty ? or does the next should be indented ?

RayenBou: I'm currently building a bigger side of this bundle that may need other settings , so i've indented to put more key in future

RayenBou:
first of all i apologize if the many close/reopen  put a little mess. I've tried to test several way to pass those test by removing requirement. But i'm in front of a wall right now:

since i've built my bundle under Symfony 7 , using the AbstractBundle class and the configure/loadExtension directly in my main class , i have no way to make it working under symfony 6.4/php 8.1. But one test require it to work on older version.

thanks in advence

Neirda24:
@RayenBou : maybe just close it and re-open it once you are ready.

PR #1629, opened 2024-07-04
Added Nerd4ever Kaya OAuth2 Module

sobrito:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |  https://packagist.org/packages/nerd4ever/kaya-oauth2-bundle


nerd4ever/kaya-oauth2-bundle/1.0/config/kaya_oauth2.yaml
```diff
@@ -0,0 +1,18 @@
+kaya_oauth2:
+    idp:
+        # O endpoint base do servidor IDP (Identity Provider)
```
fabpot, line 3: Can you use English for comments?

fabpot (changes requested):

nerd4ever/kaya-oauth2-bundle/1.0/config/kaya_oauth2.yaml
sobrito: Thank you for the assistance. I will correct all the errors and conduct tests in a private repository before opening another PR, as the project is still very premature and contains many errors. As requested, I will also translate the comments into English. However, making the code compatible with PHP 7.4 is not feasible due to syntax differences, as the project is currently on the PHP 8.1 standard. Downgrading would not be advisable.

PR #1631, opened 2024-07-09
Add softspring components

javihgil:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/softspring/components


fabpot (changes requested):
If the config for version 5.2 is the same as the one for 5.1, no need to duplicate it. You can remove it.

fabpot (approved):

PR #1642, opened 2024-07-10
Add softspring permissions-bundle

javihgil:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/softspring/permissions-bundle


javihgil:
Hi everyone:

First of all, apologize me because of many PRs, I'm automating the installation of many of our bundles with flex. Besides most of them are failing (most of them because of same problem as this, others depend on each other to be configured, and others are other reasons).

I think this config is valid, but it's failing because bundle requires php >=8.1 and the actions are testing with 7.4. 

Could I skip this check step or test it with 8.1?

Thanks

fabpot (approved):

PR #1648, opened 2024-07-15
Refactor 'dahromy_glide' to 'glide' across project

dahromy:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://github.com/dahromy/glide-symfony

The term 'dahromy_glide' has been replaced with 'glide' in routes, services, config, and dependency injections. This results in a more concise and standard naming convention across the entire codebase. This refactoring does not introduce any functionality changes.

### DIFF
--- a/dahromy/glide-symfony/1.0/config/packages/glide.yaml
+++ b/dahromy/glide-symfony/1.0/config/packages/glide.yaml
@@ -1,0 +1,3 @@
+glide:
+    source: '%kernel.project_dir%/public'
+    cache: '%kernel.project_dir%/var/cache/glide'
--- a/dahromy/glide-symfony/1.0/config/routes/glide.yaml
+++ b/dahromy/glide-symfony/1.0/config/routes/glide.yaml
@@ -1,0 +1,3 @@
+glide:
+    resource: '@DahRomyGlideBundle/Resources/config/routes.php'
+    type: php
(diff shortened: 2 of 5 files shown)


.idea/.gitignore
fabpot, line 1: The full `.idea` directory must be removed.
Can you also squash your commits after the removal of this directory?

fabpot (changes requested):

.idea/.gitignore
dahromy: Thank you for your feedback. I have made the requested changes:

1. The `.idea` directory has been completely removed from the repository.
2. I've added `.idea/` to the `.gitignore` file to prevent it from being accidentally reintroduced.
3. I've squashed the relevant commits, including the one about adding newlines to YAML configuration files, into a single commit to keep the history clean.

The branch has been updated with these modifications. Could you please review these changes and let me know if any further adjustments are needed?

fabpot: Please remove the `.gitignore` as well. We're not adding things that are specific to a local config.

fabpot: And it doesn't look like you squashed the commits.

dahromy: I apologize for the confusion earlier. I have made the requested changes:

1. The `.gitignore` file has been removed as requested. I understand now that we should not add local-specific configurations.
2. I have re-squashed the commits into a single one, which includes the removal of the `.idea` directory and the removal of the `.gitignore` file.

I have updated the branch with these new changes. Could you please review these modifications? I apologize for any confusion and thank you for your patience.

fabpot (approved):

PR #1649, opened 2024-07-18
Add recipe for bitbag/wishlist-plugin

hmfilar:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/bitbag/wishlist-plugin

### DIFF
--- a/bitbag/wishlist-plugin/4.4/post-install.txt
+++ b/bitbag/wishlist-plugin/4.4/post-install.txt
@@ -1,0 +1,2 @@
+  If you want to contact us, the best way is to fill the form on our website or send us an e-mail to <comment>hello@bitbag.io</comment> with your question(s).
+  We guarantee that we answer as soon as we can!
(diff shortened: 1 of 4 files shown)


bitbag/wishlist-plugin/4.4/post-install.txt
```diff
@@ -0,0 +1,2 @@
+  If you want to contact us, the best way is to fill the form on our website or send us an e-mail to <comment>hello@bitbag.io</comment> with your question(s).
```
fabpot, line 1: This post-install note should be removed.

hmfilar: This is a standard we use in our recipes: https://github.com/symfony/recipes-contrib/tree/main/bitbag

Why do you think it shouldn't be included here?

fabpot (approved):

PR #1673, opened 2024-09-19
feat: add shopware/app-bundle 4.0 variant

shyim:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/shopware/app-bundle

We changed the schema of the yaml, therefore need a new recipe version


shyim:
Whats the best way here? I made doctrine optional as we support dynamodb in that bundle now too. So the new require flow is like: `composer require shopware/app-bundle doctrine/orm symfony/doctrine-bridge`

shyim:
Now it's fine, we don't support Symfony 5

fabpot (approved):

PR #1680, opened 2024-09-23
feat: add storage to auto for shopware app bundle

shyim:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/shopware/app-bundle

Instead of using hard-code in-memory, we let auto detect the storage so it's more convenient for users to install


shyim:
Symfony 5 failure is fine, we don't support it

fabpot (approved):

PR #1687, opened 2024-10-09
ecphp/cas-bundle: add recipe for version 3.0

drupol:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/ecphp/cas-bundle


drupol:
The build issue make sense since we let the user choosing with which `psr7` implementation they want to work with.

Installing this bundle is usually done by doing: `composer require nyholm/psr7 ecphp/cas-bundle`

fabpot (approved):

drupol:
Cool, thanks!

PR #1689, opened 2024-10-09
mv spatie/symfony-ignition-bundle/0.0/config/packages{,/dev}/ignition.yaml

n0099:
fix `There is no extension able to load the configuration for "ignition" (in "config/packages/ignition.yaml"). Looked for namespace "ignition", found "framework", "doctrine", "twig".` when under environments other than `dev`

| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/spatie/symfony-ignition-bundle


diimpp (changes requested):
Hi @n0099, thank you for your contribution!

I would suggest an alternative approach to wrap config with `when@dev` like so (instead of moving config to `config/packages/dev`):

```yaml
when@dev:
    ignition:
        application_path: '%kernel.project_dir%'
        dark_mode: false
        should_display_exception: '%kernel.debug%'
```

This approach is available since [symfony 5.3](https://symfony.com/blog/new-in-symfony-5-3-configure-multiple-environments-in-a-single-file) and `spatie/symfony-ignition-bundle` minimal requirement is [5.4](https://packagist.org/packages/spatie/symfony-ignition-bundle#0.0.1), so we should be safe to use it.

And it's currently preferred way to do it, as shown in example https://github.com/symfony/demo/blob/f39ffa0c5983cd50f73a8d6ef1026422ca4ec355/config/packages/monolog.yaml#L5-L24

I've opened PR with same fix at https://github.com/spatie/symfony-ignition-bundle/pull/29

diimpp (approved):

PR #1690, opened 2024-10-10
Fix method switching

ajgarlag:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [symfony/apache-pack](https://packagist.org/packages/symfony/apache-pack)

Fix method switching when the path starts with `/index.php` and `mod_rewrite` is available.

Related #165


diimpp (commented):
Change looks okay, referenced PR explains the issue and [mdn](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) says similar stuff, but I don't feel qualified to approved it.

@Nyholm would you mind take a look since you've approved similar issue previously.

nicolas-grekas (approved):

PR #1694, opened 2024-10-17
added neo4j-bundle  1.0

AkshatSW:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/neo4j/neo4j-bundle

All tests are passing except the test for PHP 7.4. We are unable to support that version due to upstream limitations of the neo4j-php-client library. Yet, our bundle should work with Symfony 5.4, if the user uses PHP 8.1. We found in the CI definition that this test is safe to ignore: 

https://github.com/symfony/recipes/blob/802e06040c6bce727b9298d1eee0e0ae9002f486/.github/workflows/callable-qa.yml#L266C26-L266C123

We would love to have this recipe added to the contrib, as the definition for version 0.4 in this project contains a breaking configuration, potentially confusing new users before they are even started when they run `composer install neo4j/neo4j-bundle`.

Kindly let us know if there's anything we can do.


diimpp (approved):
Thank you.

On unrelated note, at your bundle, either `default_driver` property should have been called `default_connection` or `connections` as `drivers`. Those terms used interchangeably, which is confusing at glance reading.

image

diimpp:
P.S. Follow-up on unrelated note, it's error at bundle docs to call it `connections` instead of `drivers`.

https://github.com/neo4j-php/neo4j-symfony/pull/86

transistive:
Thank you so much for being so thorough. We are excited to see it merged! :partying_face:

PR #1696, opened 2024-10-21
[vincentlanglet/twig-cs-fixer] add default config and add message

JohJohan:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | vincentlanglet/twig-cs-fixer

add default config and add message

### DIFF
--- a/vincentlanglet/twig-cs-fixer/3.0/post-install.txt
+++ b/vincentlanglet/twig-cs-fixer/3.0/post-install.txt
@@ -1,0 +1,4 @@
+<bg=green;fg=white></>
+<bg=green;fg=white> You can create a `.twig-cs-fixer.dist.php` and configure to your needs </>
+<bg=green;fg=white> See https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md </>
+<bg=green;fg=white> For configuration explanation. </>
--- a/vincentlanglet/twig-cs-fixer/3.0/manifest.json
+++ b/vincentlanglet/twig-cs-fixer/3.0/manifest.json
@@ -1,0 +1,5 @@
+{
+    "gitignore": [
+        "/.twig-cs-fixer.cache"
+    ]
+}


vincentlanglet/twig-cs-fixer/0.6/post-install.txt
```diff
@@ -0,0 +1,4 @@
+<bg=green;fg=white></>
+<bg=green;fg=white> You can edit the `.twig-cs-fixer.php` to your needs see</>
```
fabpot, line 2:
```suggestion
<bg=green;fg=white> You can edit the `.twig-cs-fixer.php` to your needs </>
```

vincentlanglet/twig-cs-fixer/0.6/post-install.txt
```diff
@@ -0,0 +1,4 @@
...
+<bg=green;fg=white> https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md </>
```
fabpot, line 3:
```suggestion
<bg=green;fg=white> See https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md </>
```

fabpot (changes requested):

vincentlanglet/twig-cs-fixer/0.6/post-install.txt
JohJohan: Thank you for the feedback i implemented it :+1:

vincentlanglet/twig-cs-fixer/0.6/post-install.txt
JohJohan: Thank you for the feedback i implemented it :+1:

vincentlanglet/twig-cs-fixer/0.6/.twig-cs-fixer.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+$ruleset = new TwigCsFixer\Ruleset\Ruleset();
+
+// You can start from a default standard
+$ruleset->addStandard(new TwigCsFixer\Standard\TwigCsFixer());
+
+//// And then add/remove/override some rules
```
fabpot, line 8:
```suggestion
// And then add/remove/override some rules
```

maybe?

fabpot:
@VincentLanglet Maybe you can chime in here?

vincentlanglet/twig-cs-fixer/0.6/.twig-cs-fixer.php
```diff
@@ -0,0 +1,21 @@
...
+//$ruleset->addRule(new TwigCsFixer\Rules\File\FileExtensionRule());
+//$ruleset->removeRule(TwigCsFixer\Rules\Whitespace\EmptyLinesRule::class);
+//$ruleset->overrideRule(
+//    new TwigCsFixer\Rules\Punctuation\PunctuationSpacingRule(
+//        ['}' => 1],
+//        ['{' => 1],
+//    )
```
VincentLanglet, line 15: Those classes doesn't exists in 0.6 version cf
https://github.com/VincentLanglet/Twig-CS-Fixer/tree/0.6.0/src

I don't think it's worth to have a recipe for < 3.0.0 since I only maintain the latest major.
So it seems better to only create
```
vincentlanglet/twig-cs-fixer/3.0/
```
files, no ?

vincentlanglet/twig-cs-fixer/0.6/post-install.txt
```diff
@@ -0,0 +1,4 @@
+<bg=green;fg=white></>
+<bg=green;fg=white> You can edit the `.twig-cs-fixer.php` to your needs </>
+<bg=green;fg=white> See https://github.com/VincentLanglet/Twig-CS-Fixer/blob/main/docs/configuration.md </>
+<bg=green;fg=white> For configuration explanation: </>
```
VincentLanglet, line 4:
```suggestion
<bg=green;fg=white> For configuration explanation. </>
```
?

vincentlanglet/twig-cs-fixer/0.6/manifest.json
```diff
@@ -1,5 +1,8 @@
 {
     "gitignore": [
         "/.twig-cs-fixer.cache"
-    ]
+    ],
+    "copy-from-recipe": {
+        ".twig-cs-fixer.php": ".twig-cs-fixer.php"
```
VincentLanglet, line 6: I dunno if we should provide this default config file.

Without config file, I believe my tool is already using `TwigCsFixer` standard by default. Which means that if you add this config file like this without changing anything in it, it's just a useless file.

So maybe the post-install message is enough ?

diimpp (changes requested):
Hi, thank you for contribution.

I assume you've chosen 0.6 due to the last version supporting PHP 7.4, but because of changes in the `twig-cs-fixer` config file, this recipe will crash for users on 2.0/3.0; hence, you will need to add `0.6`, `2.0` (`Standard` class name is `Twig`) and `3.0`. (I didn't check if 0.6 and 1.0 are any different in the config file syntax)

vincentlanglet/twig-cs-fixer/0.6/.twig-cs-fixer.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+$ruleset = new TwigCsFixer\Ruleset\Ruleset();
+
+// You can start from a default standard
+$ruleset->addStandard(new TwigCsFixer\Standard\TwigCsFixer());
```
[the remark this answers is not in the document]
JohJohan: Will revert for 0.6 see: https://github.com/symfony/recipes-contrib/pull/1696#discussion_r1808952120

vincentlanglet/twig-cs-fixer/0.6/manifest.json
VincentLanglet: Since it can work without the config file, I don't know if it's a good idea to add automatically a config.
I'm unsure how other tools work (and if they can work without a config file)...

But the `post-install.txt` is great

JohJohan: Ah i think its a good addition as you can indicate you can make changes to the config like adding more rule, if you dont change anything you still get the same results. So i'm for it.

As for others i see php-cs-fixer has config: https://github.com/symfony/recipes/blob/main/friendsofphp/php-cs-fixer/3.0/.php-cs-fixer.dist.php but this doesn't work without it i assume

diimpp:
Folks @VincentLanglet @JohJohan , what do we decide about it? Does this package needs recipe or not?

VincentLanglet:
Recipe would be useful to update gitignore file for example.

But providing a `[.twig-cs-fixer.dist.php](https://github.com/symfony/recipes-contrib/pull/1696/files#diff-ecb7c9f7faa5beb8f9b73f6a911fd2a734c3ff63233aa5aa94de0aebd96c2d5f)` file by default is not needed.

JohJohan:
> Recipe would be useful to update gitignore file for example.
> 
> But providing a [.twig-cs-fixer.dist.php](https://github.com/symfony/recipes-contrib/pull/1696/files#diff-ecb7c9f7faa5beb8f9b73f6a911fd2a734c3ff63233aa5aa94de0aebd96c2d5f) file by default is not needed.

Okay i will remove the `.twig-cs-fixer.dist.php` file

vincentlanglet/twig-cs-fixer/3.0/post-install.txt
```diff
@@ -0,0 +1,4 @@
+<bg=green;fg=white></>
+<bg=green;fg=white> You can edit the `.twig-cs-fixer.dist.php` to your needs </>
```
VincentLanglet, line 2: You can create a `.twig-cs-fixer.dist.php` or something like this ?

JohJohan: Update this line :)

diimpp (approved):

VincentLanglet (approved):

fabpot (approved):

PR #1697, opened 2024-10-23
Add 21torr/hosting recipe for 3.2+

apfelbox:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/21torr/hosting

Update config for version 3.2+ of the package.

### DIFF
--- a/21torr/hosting/3.2/config/packages/hosting.yaml
+++ b/21torr/hosting/3.2/config/packages/hosting.yaml
@@ -1,0 +1,3 @@
+hosting:
+    tier: '%env(HOSTING_TIER)%'
+    installation: "%env(HOSTING_INSTALLATION)%"
--- a/21torr/hosting/3.2/config/routes/hosting.yaml
+++ b/21torr/hosting/3.2/config/routes/hosting.yaml
@@ -1,0 +1,3 @@
+hosting:
+    resource: '@HostingBundle/config/routes.yaml'
+    prefix: /hosting
--- a/21torr/hosting/3.2/manifest.json
+++ b/21torr/hosting/3.2/manifest.json
@@ -1,0 +1,15 @@
+{
+    "bundles": {
+        "Torr\\Hosting\\HostingBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "HOSTING_TIER": "development",
+        "HOSTING_INSTALLATION": ""
+    },
+    "gitignore": [
+        "/.build-info.json"
+    ]
+}


apfelbox:
Tests seem to fail as no PHP 8.3 is supported.

Just did a manual installation check with the guide above and everything looked fine. 👍

21torr/hosting/3.2/config/packages/hosting.yaml
```diff
@@ -0,0 +1,4 @@
+hosting:
+    tier: '%env(HOSTING_TIER)%'
+    # @todo add a unique identifier of your installation here
```
diimpp:
```suggestion
    # Add a unique identifier of your installation here.
```

21torr/hosting/3.2/config/packages/hosting.yaml
```diff
@@ -0,0 +1,4 @@
+hosting:
```
diimpp, line 1: Unrealted to current PR, but while your bundle is young, I would suggest to set proper namespacing, for example

```suggestion
torr_hosting:
```

or

```suggestion
21torr_hosting:
```

21torr/hosting/3.2/config/routes/hosting.yaml
```diff
@@ -0,0 +1,3 @@
+_import.hosting:
```
diimpp, line 1: I think according to current bundle name it should be plain
```suggestion
hosting:
```

21torr/hosting/3.2/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+    "bundles": {
+        "Torr\\Hosting\\HostingBundle": ["all"]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "HOSTING_TIER": "development",
+        "HOSTING_INSTALLATION": "todo-add-your-unique-installation-key-here"
```
diimpp, line 10: I think either value should be commented out or have valid value.

```suggestion
        "HOSTING_INSTALLATION": ""
```

21torr/hosting/3.2/config/routes/hosting.yaml
apfelbox: The keys of import routes are removed, right? Otherwise we might have an issue with name collisions.

21torr/hosting/3.2/config/packages/hosting.yaml
apfelbox: Thanks for the suggestion, I will think about that for v4.

apfelbox: I just did the change for v4 of the bundle.

fabpot (approved):

PR #1698, opened 2024-10-24
Add 21torr/storyblok 

apfelbox:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/21torr/storyblok

For v3 it's only env vars and the config file.


apfelbox:
The CI fails as it doesn't support PHP 8.3

I ran the test above manually and everything is looking good. 👍

fabpot (approved):

PR #1703, opened 2024-11-08
[Yousign\ZddMessageBundle] Move configuration file under config/packages/

Kocal:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [](https://packagist.org/packages/yousign/zdd-message-bundle)

I've been a bit surprised when upgrading the package, so see the configuration file under `config/*`, we expect it to be under `config/packages/*` like every other packages configuration.


Kocal:
I do not think failing checks are caused by this PR:
image

fabpot (approved):

PR #1713, opened 2024-11-18
Survos crawler

tacman:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | removing the recipe


fabpot:
This PR looks very empty 😄

tacman:
Yeah, I wanted to remove the recipe, as I'm experimenting with a custom installer that works from within the bundle.  While of course it doesn't do as much as recipes does, it copies the config files and updates .gitignore and .env, so it works for almost everything I want, and testing it is much easier.

https://github.com/survos/installer/

diimpp:
Hi @tacman, going to close it for now since it's empty.

tacman:
How do I remove a recipe?

diimpp:
> How do I remove a recipe?

Should be the same as for adding new recipes -- Remove files that no longer needed, commit the changes and create a new PR.

PR #1714, opened 2024-11-22
rename "database" service into "eventstore" to avoid conficts with doctrine

DavidBadura:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/patchlevel/event-sourcing-bundle

If you use doctrine and event-sourcing recipes, you get a conflict in the docker-compose.yaml.

This is now avoided by giving the event store its own DB service. This also makes more sense, since it is usually not the same DB that you use for event-sourcing.


DavidBadura:
The bundle only supports Symfony >= 6

PR #1715, opened 2024-11-28
Fix gotenberg bundle default DSN

norkunas:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/sensiolabs/gotenberg-bundle

### DIFF
--- a/sensiolabs/gotenberg-bundle/0.1/manifest.json
+++ b/sensiolabs/gotenberg-bundle/0.1/manifest.json
@@ -6,7 +6,7 @@
         "config/": "%CONFIG_DIR%/"
     },
     "env": {
-        "GOTENBERG_DSN": "http://localhost:3000"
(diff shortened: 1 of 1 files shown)


diimpp (commented):
Hi @norkunas, `GotenbergBundle` maintainers seems to be okay with localhost as default.

https://github.com/sensiolabs/GotenbergBundle?tab=readme-ov-file#with-symfony-flex

image

You only need `gotenberg` host if you're calling your php script from the same docker setup, from host point of view it will be `localhost`.

Could you please start with GotenbergBundle readme and after it will be accepted we can merge it here?

norkunas:
Ok, I didn't read the readme :D one of our devs installed gotenberg bundle and complained that it didn't work. if the docker image is added by flex then it'd be expected to get the http://gotenberg url instead of localhost. As it's explained in the readme (where I should have pointed our dev) thefore I can close this

PR #1717, opened 2024-11-29
Add MongoDB extension and service to Docker

GromNaN:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/doctrine/mongodb-odm-bundle

Configuration inspired by [`doctrine/doctrine-bundle`](https://github.com/symfony/recipes/blob/main/doctrine/doctrine-bundle/2.9/manifest.json).

### DIFF
--- a/doctrine/mongodb-odm-bundle/4.4/manifest.json
+++ b/doctrine/mongodb-odm-bundle/4.4/manifest.json
@@ -11,7 +11,39 @@
         "env(MONGODB_DB)": ""
     },
     "env": {
+        "#1": "Format described at https://www.mongodb.com/docs/php-library/current/connect/connection-options/",
+        "#2": "MONGODB_URL=\"mongodb://username:password@localhost:27017/?authSource=auth-db\"",
+        "#3": "MONGODB_URL=\"mongodb+srv://username:password@YOUR_CLUSTER_NAME.YOUR_HASH.mongodb.net/?retryWrites=true&w=majority\"",
         "MONGODB_URL": "mongodb://localhost:27017",
         "MONGODB_DB": "symfony"
+    },
+    "dockerfile": [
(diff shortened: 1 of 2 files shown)


doctrine/mongodb-odm-bundle/4.4/manifest.json
```diff
@@ -11,7 +11,39 @@
...
+        "#2": "MONGODB_URL=\"mongodb://username:password@localhost:27017/?authSource=auth-db\"",
+        "#3": "MONGODB_URL=\"mongodb+srv://username:password@YOUR_CLUSTER_NAME.YOUR_HASH.mongodb.net/?retryWrites=true&w=majority\"",
         "MONGODB_URL": "mongodb://localhost:27017",
         "MONGODB_DB": "symfony"
+    },
+    "dockerfile": [
+        "RUN install-php-extensions mongodb"
+    ],
+    "docker-compose": {
+        "docker-compose.yml": {
```
OskarStark, line 24: General flex question:
State of the art is now compose.yaml file, does it work also? 
cc @nicolas-grekas @fabpot

GromNaN: Good point, it looks like this work but should be updated to `compose.yaml`: https://github.com/symfony/flex/blob/92f4fba342161ff36072bd3b8e0b3c6c23160402/src/Configurator/DockerComposeConfigurator.php#L158

OskarStark: Best way would be to update both files or make it configurable for both files

OskarStark: Looks like it does

fabpot (approved):

PR #1719, opened 2024-12-04
fix: default invalid key

martingold:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |  https://packagist.org/packages/fluffydiscord/roadrunner-symfony-bundle

### DIFF
--- a/fluffydiscord/roadrunner-symfony-bundle/3.0/config/packages/fluffy_discord_road_runner.yaml
+++ b/fluffydiscord/roadrunner-symfony-bundle/3.0/config/packages/fluffy_discord_road_runner.yaml
@@ -66,5 +66,6 @@ fluffy_discord_road_runner:
         keypair_path: bin/keypair.key
 
 when@dev:
-    fluffy_discord_road_runner:
-        early_router_initialization: false
+    http:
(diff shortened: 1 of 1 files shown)


fluffydiscord/roadrunner-symfony-bundle/3.0/config/packages/fluffy_discord_road_runner.yaml
```diff
@@ -66,5 +66,6 @@ fluffy_discord_road_runner:
         keypair_path: bin/keypair.key
 
 when@dev:
-    fluffy_discord_road_runner:
-        early_router_initialization: false
+    http:
+        fluffy_discord_road_runner:
+            early_router_initialization: false
```
diimpp, line 71:
```suggestion
    fluffy_discord_road_runner:
        http:
            early_router_initialization: false
```

diimpp (changes requested):
Hi @martingold , `http` part cannot be on top, so I assume it should be the other way around.

image

martingold:
You are correct, fixed. Thanks!

fabpot (approved):

PR #1734, opened 2024-12-24
Fix boshurik/telegram-bot-bundle

Evilkeeper:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [boshurik/telegram-bot-bundle](https://packagist.org/packages/boshurik/telegram-bot-bundle)

yml file was deleted in [this commit](https://github.com/BoShurik/TelegramBotBundle/commit/e5f16a8be354a6689081333e0faaa4ea7129d9e9#diff-bea85dabca61a35bc31e4ed5664459656d1db5912749510152e2e40690dbfb0b), it generates bad config since then

@BoShurik


diimpp (changes requested):
Hi, `routing.yml` was replaced with `routing.php` at 6.0 release, but 3,4,5 releases still have it.

https://github.com/BoShurik/TelegramBotBundle/blob/5.0.2/src/Resources/config/routing.yml

Hence there is no bug in older versions and you should add new 6.0 related recipe instead to handle new file structure.

diimpp (approved):
@Evilkeeper thank you :+1:

PR #1736, opened 2025-01-03
use yaml instead of php for config

tacman:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/survos/crawler-bundle

Use yaml instead of PHP for bundle configuration


tacman:
My bundle doesn't support php7.  So why is it complaining?

fabpot (approved):

PR #1737, opened 2025-01-04
fix call to open file

tacman:


survos/command-bundle/1.5/post-install.txt
```diff
@@ -6,4 +6,4 @@
     1. Configure /config/packages/survos_command.yaml to expose the namespaces (command prefixes)
```
fabpot, line 6:
```suggestion
    1. Configure config/packages/survos_command.yaml to expose the namespaces (command prefixes)
```

fabpot:
The description is missing the standard table (which is why the tests don't pass).

fabpot (approved):

PR #1742, opened 2025-01-16
Update Inspector Recipe to 1.5

ilvalerione:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     |


fabpot:
If I read the PR correctly, there is no difference between 1.4 and 1.5. If that's the case, plese remove the files for 1.5 as they are not needed.

ilvalerione:
Done

PR #1763, opened 2025-03-19
add recipe for synolia/sylius-mail-tester-plugin@v3

maxperei:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/synolia/sylius-mail-tester-plugin


synolia/sylius-mail-tester-plugin/3.0/config/routes/sylius_mail_tester.yaml
```diff
@@ -0,0 +1,3 @@
+synolia_mail_tester:
```
diimpp, line 1: Just a suggestion since imported route file is specifically `admin`

```suggestion
synolia_mail_tester_admin:
```

maxperei: hi @diimpp, that's interesting but this plugin is not meant to add/alter a shop route so that's why i left like this

diimpp (approved):

PR #1769, opened 2025-03-30
Remove the Bref recipe

mnapoli:
We're moving away from Flex recipes as this is too unpractical to maintain.

See https://github.com/brefphp/bref/pull/1934

| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/bref/symfony-bridge


leob (approved):
@mnapoli Okay, so we're deleting them here because we'll move them to the Bref repo ... approving this!

mnapoli:
Thank you!

PR #1775, opened 2025-04-06
update php-redis-om version to 0.4

clementtalleu:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/talleu/php-redis-om


fabpot:
This change is not needed. `0.3` is the min version supported by the recipe.
A new recipe is only needed when the config changes.

PR #1782, opened 2025-04-18
altcha bundle

tito10047:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [huluti/altcha-bundle](https://packagist.org/packages/huluti/altcha-bundle)

### DIFF
--- a/huluti/altcha-bundle/1.0/config/routes/huluti_altcha.yaml
+++ b/huluti/altcha-bundle/1.0/config/routes/huluti_altcha.yaml
@@ -1,0 +1,2 @@
+live_component:
+    resource: '@HulutiAltchaBundle/config/routes.yml'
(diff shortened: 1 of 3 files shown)


huluti/altcha-bundle/1.0/config/routes/huluti_altcha.yaml
```diff
@@ -0,0 +1,2 @@
+live_component:
```
diimpp, line 1: Please update routes import, as it should be named after your bundle. Additionally, `live_component` is reserved for ux live component package.

```suggestion
huluti_altcha:
```

diimpp (changes requested):

huluti/altcha-bundle/1.0/config/routes/huluti_altcha.yaml
tito10047: done. thanks

diimpp (approved):

PR #1796, opened 2025-05-11
Create recipes for adrienlbt/hexagonal-maker-bundle

AdrienWac:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/adrienlbt/hexagonal-maker-bundle

Recipes creation for the first version of symfony bundle, Hexagonal Maker Bundle.


AdrienWac:
Hi @fabpot , if i understand correctly my bundle must be compatible with symfony:^6.4 and symfony:^5.4 to be available for recipes ?

diimpp:
> Hi @fabpot , if i understand correctly my bundle must be compatible with symfony:^6.4 and symfony:^5.4 to be available for recipes ?

Hi, no, it's not required. The CI is configured this way, but it doesn't block the package from being merged.

adrienlbt/hexagonal-maker-bundle/1.2/config/packages/hexagonal_maker_bundle.yaml
```diff
@@ -0,0 +1,4 @@
+hexagonal_maker:
+    application_path: 'Application'
+    domain_path: 'Domain'
+    infrastructure_path: 'Infrastructure'
```
diimpp, line 4: Since your bundle is only available in `dev` env, this configuration will fail in `prod` and needs to be wrapped in `when@dev`.

```suggestion
when@dev:
    hexagonal_maker:
        application_path: 'Application'
        domain_path: 'Domain'
        infrastructure_path: 'Infrastructure'
```

diimpp (approved):

PR #1974, opened 2026-04-30
Add recipe for codevenom/fakturownia-bundle 0.1 (#1974)

levan-cdv:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/codevenom/fakturownia-bundle

Adds a Symfony Flex recipe for `codevenom/fakturownia-bundle`.

The recipe enables the bundle and installs a minimal default configuration using environment variables for the Fakturownia API connection and seller information.


codevenom/fakturownia-bundle/0.1/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+    "bundles": {
+        "Codevenom\\FakturowniaBundle\\CodevenomFakturowniaBundle": ["all"]
```
levan-cdv, line 3: Is it possible to skip the QA check that installs the recipe into a `symfony/skeleton:^6` project?

This bundle currently does not support Symfony 6 because it depends on `symfony/mcp-bundle`, which is not compatible with Symfony 6. The bundle is intended to support Symfony 7+ only.

I can keep the bundle constraints explicit in `composer.json`, for example requiring Symfony `^7.0 || ^8.0`, but the current QA matrix still tries to install it into a Symfony 6 skeleton and fails during Composer dependency resolution.
Here is the `composer.json` - https://github.com/codevenom-co/fakturownia-bundle/blob/023d75cc20bebab768d0fec5e0c86a33aacc53e3/composer.json#L8


Is there a recommended way for Symfony 7+ only bundles/recipes to handle this in `recipes-contrib`?

diimpp: Hi,

QA checks don't need to fully pass in order for PR to be merged. If package doesn't support SF6 -- that's totally fine.

diimpp (approved):

codevenom/fakturownia-bundle/0.1/manifest.json
levan-cdv: Thanks

PR #1978, opened 2026-05-07
Add recipe for lubomirfiala/sulu-preview-nav

lubomirfiala:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/lubomirfiala/sulu-preview-nav

Adds a Flex recipe for lubomirfiala/sulu-preview-nav, a Symfony bundle that enables click-to-edit block navigation in the Sulu 3 CMS admin preview. The recipe auto-registers the bundle in config/bundles.php.

### DIFF
--- a/lubomirfiala/sulu-preview-nav/0.1/manifest.json
+++ b/lubomirfiala/sulu-preview-nav/0.1/manifest.json
@@ -1,0 +1,5 @@
+{
+    "bundles": {
+        "Lubomirfiala\\SuluPreviewNav\\SuluPreviewNavBundle": ["all"]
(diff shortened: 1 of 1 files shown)


lubomirfiala:
| Q       | A   |
|---------|-----|
| License | MIT

diimpp:
@lubomirfiala Hi, I think you need to push another commit for QA check to understand MIT change, simple re-run still fails.

fabpot:
The recipe is not needed as this is the default already when installing a bundle without a recipe.

PR #1981, opened 2026-05-18
fix(php_codesniffer): Ignore `config/reference.php`

ebitkov:
| Q             | A
| ------------- | ---
| License       | MIT

Since Symfony v7.4/v8.0, a `config/reference.php` file is auto-generated. CodeSniffer produces many false positives for this file, which can safely be ignored.

### DIFF
--- a/squizlabs/php_codesniffer/3.6/phpcs.xml.dist
+++ b/squizlabs/php_codesniffer/3.6/phpcs.xml.dist
@@ -16,4 +16,6 @@
     <file>src/</file>
     <file>tests/</file>
 
+    <exclude-pattern>config/reference.php</exclude-pattern>
+
 </ruleset>
--- a/squizlabs/php_codesniffer/3.6/phpcs.xml.dist
+++ b/squizlabs/php_codesniffer/3.6/phpcs.xml.dist
@@ -16,6 +16,6 @@
     <file>src/</file>
     <file>tests/</file>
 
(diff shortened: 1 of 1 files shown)


diimpp:
@jrfnl Hi, please advise on this change.

New `config/reference.php` file is part of new symfony configuration format.


> These shapes are generated dynamically based on the bundles installed in your application. The generated file is called `reference.php` and is stored in the config/ directory. You should commit it to your repository and optionally add a new entry in the autoload section of composer.json (for example, "classmap": ["config/"]).

https://symfony.com/blog/new-in-symfony-7-4-better-php-configuration

squizlabs/php_codesniffer/3.6/phpcs.xml.dist
```diff
@@ -16,4 +16,6 @@
     <file>src/</file>
     <file>tests/</file>
 
+    <exclude-pattern>config/reference.php</exclude-pattern>
```
jrfnl:
```suggestion
    <exclude-pattern>config/reference\.php$</exclude-pattern>
```

jrfnl:
> @jrfnl Hi, please advise on this change.

Not sure what kind of advise you expect ? The suggestion appears valid, though it could also be considered to make the auto-generated output comply with PSR12 to begin with by running `phpcbf` over the auto-generated output ?

diimpp (approved):

PR #1989, opened 2026-05-28
Update routes config for SonataAdminBundle to PHP

willemverspyck:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/sonata-project/admin-bundle

The XML config of SonataAdminBundle is deprecated and replaces by PHP config.

### DIFF
--- a/sonata-project/admin-bundle/4.0/config/routes/sonata_admin.yaml
+++ b/sonata-project/admin-bundle/4.0/config/routes/sonata_admin.yaml
@@ -1,5 +1,5 @@
 admin_area:
-    resource: "@SonataAdminBundle/Resources/config/routing/sonata_admin.xml"
+    resource: "@SonataAdminBundle/Resources/config/routing/sonata_admin.php"
     prefix: /admin
 
 _sonata_admin:


sonata-project/admin-bundle/4.0/config/routes/sonata_admin.yaml
```diff
@@ -1,5 +1,5 @@
 admin_area:
-    resource: "@SonataAdminBundle/Resources/config/routing/sonata_admin.xml"
+    resource: "@SonataAdminBundle/Resources/config/routing/sonata_admin.php"
```
diimpp, line 2: sonata_admin.`php` only exists starting from version `4.39`, but this recipe applies to all versions since `4.0`. You would need to create a new recipe version; otherwise, this change will break installations running older versions.

https://github.com/sonata-project/SonataAdminBundle/blob/4.39.0/src/Resources/config/routing/sonata_admin.php

diimpp (changes requested):

diimpp:
Done in https://github.com/symfony/recipes-contrib/pull/1993

PR #2015, opened 2026-07-23
Add fluffydiscord/roadrunner-symfony-bundle 7.0 and 7.1 recipes

FluffyDiscord:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/fluffydiscord/roadrunner-symfony-bundle

I'm the bundle author.

### 7.0 recipe

v7.0.0 removed the `http.early_router_initialization` option (the 2.3/3.0 recipe default config throws an `InvalidConfigurationException` on v7) and replaced it with a zero-config worker pre-warming system configured under a new top-level `warmup` node.

The 7.0 recipe drops the removed option (including its `when@dev` override) and documents the new `warmup` options with their defaults. Manifest is unchanged.

Bundle release: https://github.com/FluffyDiscord/roadrunner-symfony-bundle/releases/tag/v7.0.0

### 7.1 recipe

v7.1.0 adds an `http.request_factory` option (`auto`/`native`/`psr7`) that selects how the RoadRunner request is converted into a Symfony request; the new `native` default roughly halves the per-request conversion cost.

It is documented in a separate 7.1 recipe rather than being added to the 7.0 one, because the option does not exist in 7.0.x — writing it into a 7.0 install fails with `Unrecognized option "request_factory" under "fluffy_discord_road_runner.http"`, which is the same breakage this PR set out to fix. Both recipes were verified against their respective releases' configuration tree.

Bundle release: https://github.com/FluffyDiscord/roadrunner-symfony-bundle/releases/tag/v7.1.0

### DIFF
--- a/fluffydiscord/roadrunner-symfony-bundle/7.0/config/packages/fluffy_discord_road_runner.yaml
+++ b/fluffydiscord/roadrunner-symfony-bundle/7.0/config/packages/fluffy_discord_road_runner.yaml
@@ -1,0 +1,77 @@
+fluffy_discord_road_runner:
+    # Optional
+    # Specify relative path from "kernel.project_dir" to your RoadRunner config file
+    # if you want to run cache:warmup without having your RoadRunner running in background,
+    # e.g. when building Docker images. Default is ".rr.yaml"
+    rr_config_path: ".rr.yaml"
+
+    # https://docs.roadrunner.dev/http/http
+    http:
+        # Optional
+        # -----------
+        # This decides when to boot the Symfony kernel.
+        #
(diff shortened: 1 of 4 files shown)


diimpp (changes requested):
@FluffyDiscord  Hi, please add MIT license to the PR before it can be processed further.

Here is example,
```markdown
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/fluffydiscord/roadrunner-symfony-bundle
```

Image

diimpp (approved):

PR #2021, opened 2026-08-07
Adding GearSite JSON API Recipe for V1.0.0

ErvinSabic:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/gearsite/api


ErvinSabic:
Let me verify with certainty that it will work with Symfony 8. But I don't suspect there will be any problems.

ErvinSabic:
That's about as far as I'll take it. If we require PHP 8.1 support, I'll ensure we have it but as it stands the library is at 8.2. 8.1 is EOL.

diimpp (approved):

PR #2026, opened 2026-08-15
Add zestly/dev-login-bundle recipe for 0.1

andrewbrereton:
| Q             | A
| ------------- | ---
| License       | MIT

Adds a recipe for [`zestly/dev-login-bundle`](https://packagist.org/packages/zestly/dev-login-bundle) — password-free login to your own application in the `dev` environment, so browser-driven testing (and coding agents doing it) don't stop to ask a human to type a password.

- Repository: https://github.com/ZestlyDigital/dev-login-bundle
- Requires PHP 8.2+, Symfony ^7.0 || ^8.0

### Why a recipe is needed

The bundle can't self-register its routes, and they must land in `config/routes/dev/` — that environment-scoped directory is one of its safety gates, ensuring the URLs don't exist in a production router at all. Without the recipe every user has to write that import by hand.

The recipe registers the bundle for `dev` only, and writes two files:

- `config/routes/dev/zestly_dev_login.yaml` — the routes import.
- `config/packages/dev/zestly_dev_login.yaml` — `identities`, the list of accounts offered by the discovery endpoint.

Per the Best Practices in the README, the package config contains only the one setting that has no sensible default and must be filled in by the application. The bundle's other options all have working defaults and are left to its documentation rather than shipped here.

### Checks

Validated locally against `symfony-tools/recipes-checker` before opening: `lint:manifests --contrib` and `lint:yaml` both exit 0, no aliases, all files listed under `copy-from-recipe`, 4-space indentation, trailing newlines, underscore notation, no `: ~` nulls, no `parameters:`.

I also installed the package into a fresh `symfony/skeleton:^8` with `extra.symfony.allow-contrib true`, applied this recipe, and confirmed the container compiles, both routes register and a login works end to end. That exercise found a bug in the bundle — `Security::login()` throws on the skeleton's authenticator-less `main` firewall — which is fixed in v0.1.1, released before opening this PR.

### DIFF
--- a/zestly/dev-login-bundle/0.1/config/packages/dev/zestly_dev_login.yaml
+++ b/zestly/dev-login-bundle/0.1/config/packages/dev/zestly_dev_login.yaml
@@ -1,0 +1,16 @@
+zestly_dev_login:
+    # The identities offered by GET /_dev/login and `bin/console dev:login`.
+    #
+    # There is no sensible default for this — it depends entirely on your fixtures — which is
+    # why it is the one setting this recipe writes. Everything else the bundle exposes already
+    # has a working default and is documented in the README.
+    #
+    # This is a convenience menu, not a whitelist: any identifier your application's user
+    # provider accepts will work, and listing one here grants nothing on its own.
+    #
+    # To build the list from your fixtures, a repository, or per-subdomain in a multi-tenant
+    # app, implement Zestly\DevLoginBundle\Identity\IdentityProviderInterface and alias it —
+    # it receives the current Request.
+    identities: []
+        # - { identifier: 'admin@example.com', label: 'Admin', roles: ['ROLE_ADMIN'] }
+        # - { identifier: 'user@example.com',  label: 'Regular user' }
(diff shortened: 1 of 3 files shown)


andrewbrereton:
The one red check is the expected Symfony 6 case, which the workflow itself annotates as ignorable.

| Job | Result |
| --- | --- |
| skeleton ^6 / PHP 8.1 | ❌ — package requires `php >=8.2` |
| skeleton ^7 / PHP 8.2 | ✅ |
| skeleton ^7 / PHP 8.4 | ✅ |
| skeleton ^8 / PHP 8.4 | ✅ |

The failure is purely the platform requirement:

```
- zestly/dev-login-bundle[v0.1.0, ..., v0.1.1] require php >=8.2
```

The package requires PHP 8.2+ and `symfony/*: ^7.0 || ^8.0`, so it cannot install on Symfony 6 by design. Every version it does claim to support installs and applies the recipe cleanly.

`Run checks` is green.

diimpp (approved):

andrewbrereton:
Thanks for the `when@dev:` pointer @diimpp — you're right that the env-scoped directory is the outdated approach. Applied in #2030, which moves both files to flat paths under a `when@dev:` guard.

One small deviation: the packages file is guarded too, rather than left as a bare root key. The bundle is registered for `dev` only, so in `prod` there'd be no extension able to load a `zestly_dev_login:` section and `cache:warmup` would fail.

PR #2027, opened 2026-08-16
Add tito10047/progressive-image-bundle 2.0

tito10047:
https://github.com/tito10047/progressive-image-bundle

| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [tito10047/progressive-image-bundle](https://packagist.org/packages/tito10047/progressive-image-bundle)


tito10047:
The remaining "Run updated recipes" failure is on the PHP 8.2 + Symfony 7 leg, which — unlike the PHP 8.1 leg — has no tolerance in the checker's own script (`callable-qa.yml`'s "You can ignore this error if your package does not support Symfony 6" only wraps the 8.1 step).

`tito10047/progressive-image-bundle` requires `php: ">=8.3"`, transitively forced by `intervention/image` (`^4.2`, itself `"php": "^8.3"`), so this leg will always fail here — it's not a bug this recipe/package can route around without a major (and out of scope) downgrade of the image-processing dependency.

Everything else in that job passes cleanly: all three PHP 8.4 legs (Symfony 6/7/8) show `cache:clear [OK]`, `assets:install [OK]`, and `progressive-image:generate-custom-css [OK]`. That includes a real regression we found and fixed via this PR — the `asset_mapper` resolver had a hard `Reference('asset_mapper')` that broke container compilation on any project without `symfony/asset-mapper` installed (exactly what a bare `create-project` here does). It's now an optional reference, with a clear `LogicException` at first actual use instead of a compile-time failure.

diimpp (approved):

PR #2030, opened 2026-08-19
Update zestly/dev-login-bundle recipe to use when@dev

andrewbrereton:
| Q             | A
| ------------- | ---
| License       | MIT

Follow-up to the review comment on #2026, which landed after the PR was merged:

> FYI, that's an outdated approach, modern one is
> ```yaml
> when@dev:
>     zestly_dev_login:
>         resource: '@ZestlyDevLoginBundle/config/routes.php'
> ```

Thanks — applied. Both files move out of their env-scoped directories to flat paths with a `when@dev:` guard:

- `config/routes/dev/zestly_dev_login.yaml` → `config/routes/zestly_dev_login.yaml`
- `config/packages/dev/zestly_dev_login.yaml` → `config/packages/zestly_dev_login.yaml`

One small deviation from the suggestion: the packages file is guarded by `when@dev:` as well, rather than left as a bare `zestly_dev_login:` root key. The bundle is registered for `dev` only, so in `prod` there is no extension able to load that section and `cache:warmup` would fail.

`manifest.json` needed no `copy-from-recipe` change — it copies `config/` wholesale; only its post-install text referred to the old paths.

Staying in the `0.1` directory: the config shape is unchanged, so applications installed with the previous version of the recipe need no migration.

Checked locally with `symfony-tools/recipes-checker`: `lint:manifests --contrib` and `lint:yaml` both pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01FtiTUbCskmCiy9k78vCL19


andrewbrereton:
CI is green apart from the two `Create-project with skeleton ^6` steps, which is the same expected failure as on #2026 and the one the workflow itself annotates:

```
- zestly/dev-login-bundle[v0.1.0, ..., v0.1.1] require php >=8.2
```

The package requires PHP 8.2+ and Symfony `^7.0 || ^8.0`, so it cannot install on the PHP 8.1 job at all. The steps that can install it all pass — skeleton `^7` on PHP 8.2, and `^7` and `^8` on PHP 8.4 — so the `when@dev:` recipe applies and the application builds on every supported combination.

diimpp (approved):

PR #2033, opened 2026-08-24
Improve JanePHP recipes first-run experience

Korbeil:
| Q             | A
| ------------- | ---
| License       | MIT

## Description

Fixes the broken first-run experience reported in https://github.com/janephp/janephp/issues/860.

The JanePHP recipes currently:
- point `config/jane/*.php` to specification files (`open-api.yaml`, `json-schema.json`) that the recipe never creates, so running the provided binary right after install crashes with a raw PHP warning and a misleading fatal error;
- run generation through a `NullOutput`, hiding success output entirely and letting exceptions surface as unformatted PHP stack traces;
- always exit with code 0, even when generation fails;
- list post-install steps in an order that suggests running the generator before any specification file exists.

### Changes (applied identically to `jane-php/open-api-common/6.0`, `7.0`, `jane-php/json-schema/6.0` and `7.0`)
- **config/jane/*.php**: documented every configuration key and added an explicit note that the specification file must exist before running the generator.
- **bin/*-generate**: replaced `NullOutput` with `ConsoleOutput`, render thrown errors via `Application::renderThrowable()` and propagate exit codes.
- **post-install.txt**: placing the specification file is now step 1, running the generator moved to step 4, and the vague "remove line comments" step now explains it enables autowiring of the generated normalizer.
- **config/packages/jane.yaml**: comments explaining what uncommenting the service does and that its namespace must match the config file.

The `json-schema/5.0` recipe (Jane 5, EOL) was intentionally left untouched.

### DIFF
--- a/jane-php/json-schema/6.0/bin/jane-json-schema-generate
+++ b/jane-php/json-schema/6.0/bin/jane-json-schema-generate
@@ -4,12 +4,20 @@
 require __DIR__ . '/../vendor/autoload.php';
 
 use Jane\JsonSchema\Console\Command\GenerateCommand;
-use Symfony\Component\Console\Input\ArrayInput;
-use Symfony\Component\Console\Output\NullOutput;
 use Jane\JsonSchema\Console\Loader\ConfigLoader;
 use Jane\JsonSchema\Console\Loader\SchemaLoader;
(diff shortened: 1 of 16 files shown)


lyrixx (approved):
Thank you Baptiste

PR #2034, opened 2026-08-27
Add recipe for sulu/mcp-bundle 1.0

wachterjohannes:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/sulu/mcp-bundle

[SuluMcpBundle](https://github.com/sulu/SuluMcpBundle) turns a Sulu installation into an MCP server. The bundle needs configuration and routes, so it needs a recipe.

The recipe registers the bundle and `symfony/mcp-bundle`, which has no recipe of its own, writes the routes, and writes `config/packages/sulu_mcp.yaml`.

That file carries three sections. `sulu_mcp.server_url` is the one bundle setting without a sensible default. `mcp.http.allowed_hosts` names the public host on the transport: left unset, the MCP SDK's DNS rebinding protection accepts localhost only and a server under its own domain answers every request with `403 Forbidden: Invalid Host header.` after a successful OAuth handshake. `league_oauth2_server.scopes.default` is required by league, and the bundle deliberately contributes only `scopes.available` itself.

The two foreign sections live in `sulu_mcp.yaml` rather than in `mcp.yaml` and `league_oauth2_server.yaml`, because Symfony merges configuration per extension and the latter file belongs to the recipe of `league/oauth2-server-bundle`. Scope lists append across files, so `scopes.default` adds the MCP scopes to what the league recipe wrote instead of replacing it.

Precedent for registering `Symfony\AI\McpBundle\McpBundle` from another package's recipe: `sylius/admin-mcp-server-plugin`.

The setup no configurator can express (the OAuth key pair, the grants, the migration, the security firewall) is printed as post-install output.


wachterjohannes:
`Run updated recipes` is red. Here is what is behind each step, so it is not mistaken for a broken recipe.

The job creates a bare `symfony/skeleton` and runs `composer require sulu/mcp-bundle`.

* **skeleton ^6 (PHP 8.1 and 8.4):** the bundle requires PHP >= 8.2 and Symfony ^7.3, so there is nothing to resolve. The workflow prints its own hint for this case ("You can ignore this error if your package does not support Symfony 6").
* **skeleton ^8:** Symfony 8 support is not there yet. It is tracked in [sulu/sulu#8216](https://github.com/sulu/sulu/issues/8216) and planned before the 1.0 release.
* **skeleton ^7 (PHP 8.2 and 8.4):** resolution succeeds and the recipe is applied. `cache:clear` then fails with `The service "Sulu\Mcp\Infrastructure\Sulu\Security\EntryPoint\OAuthAuthorizeEntryPoint" has a dependency on a non-existent service "sulu_security.authentication_entry_point"`. The bundle extends a Sulu installation, `sulu/sulu` has no Flex recipe, so none of its bundles are registered in the test skeleton and the decorated service never exists. A bare skeleton cannot boot this bundle, with or without the recipe.

The first run of this job did report a real bug in the recipe. It shipped `mcp.http.allowed_hosts`, an option `symfony/mcp-bundle` only has since 0.11, while the released `sulu/mcp-bundle` 1.0.0-RC1 requires `^0.6`. That broke `cache:clear` on every install. It is fixed in the current head: the option moved into the post-install output, with a note on the version it needs.

wachterjohannes:
RC2 is out and requires `symfony/mcp-bundle ^0.12`, so `mcp.http.allowed_hosts` is back in the shipped config, together with `league_oauth2_server.scopes.default`. The current run installs `sulu/mcp-bundle 1.0.0-RC2` with `symfony/mcp-bundle v0.12.0` and gets past extension configuration, so both settings validate. What remains red is the structural part described above.

sulu/mcp-bundle/1.0/manifest.json
```diff
@@ -0,0 +1,73 @@
...
+        ],
+        "Sulu\\Mcp\\Infrastructure\\Symfony\\HttpKernel\\SuluMcpBundle": [
+            "all"
+        ]
+    },
+    "copy-from-recipe": {
+        "config/": "%CONFIG_DIR%/"
+    },
+    "env": {
+        "SULU_MCP_SERVER_URL": "https://localhost"
```
diimpp, line 14:
```suggestion
        "SULU_MCP_SERVER_URL": "http://localhost"
```

diimpp (approved):

PR #2035, opened 2026-08-27
Add lgarret/health-check-bundle 1.0 recipe

LouisGarret:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | [lgarret/health-check-bundle](https://packagist.org/packages/lgarret/health-check-bundle)

[lgarret/health-check-bundle](https://github.com/LouisGarret/health-check-bundle) exposes a `/health` endpoint reporting the status of the application and its dependencies. Doctrine connections and AssetMapper are auto-detected; custom checks implement `HealthCheckInterface`. It supports Symfony 6.4, 7.x and 8.x.

The recipe registers the bundle, imports the route — which the bundle serves through its own routing loader, so the path stays configurable — writes `config/packages/health_check.yaml` with the defaults made explicit, and provisions `HEALTH_CHECK_SECRET`.

`HEALTH_CHECK_SECRET` gets the `!ChangeMe!` placeholder rather than an empty value on purpose: the controller compares it with `hash_equals()`, so an empty secret would match an empty `Authorization` header and hand the detailed check results to any caller.

Source: https://github.com/LouisGarret/health-check-bundle

### DIFF
--- a/lgarret/health-check-bundle/1.0/config/packages/health_check.yaml
+++ b/lgarret/health-check-bundle/1.0/config/packages/health_check.yaml
@@ -1,0 +1,11 @@
+health_check:
+    path: '/health'
+    secret: '%env(HEALTH_CHECK_SECRET)%'
+    header: 'Authorization'
+    timeout: 5
+    cache:
+        enabled: true
+        ttl: 300
+    checks:
+        doctrine: true
+        asset_mapper: true
--- a/lgarret/health-check-bundle/1.0/config/routes/health_check.yaml
+++ b/lgarret/health-check-bundle/1.0/config/routes/health_check.yaml
@@ -1,0 +1,3 @@
+health_check:
+    resource: .
+    type: health_check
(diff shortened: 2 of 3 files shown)


LouisGarret:
Note on the red `Run updated recipes` job: only the `skeleton ^6 / PHP 8.1` step fails, and it fails at dependency resolution, before the recipe is ever applied:

```
- lgarret/health-check-bundle[v1.0.1, ..., v1.3.1] require php >=8.2
```

The package requires PHP >= 8.2 because three of its DTOs are `final readonly class`, which is 8.2 syntax. That step pins PHP 8.1, so it is unreachable for this package — this is the case the job's own `You can ignore this error if your package does not support Symfony 6` message covers.

The four other steps pass and do exercise the recipe: `skeleton ^7 / PHP 8.2`, and `skeleton ^6`, `^7`, `^8` on PHP 8.4.

diimpp (approved):

PR #2041, opened 2026-08-31
Add gplanchat/durable-bundle recipe

gplanchat:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/gplanchat/durable-bundle

[`gplanchat/durable-bundle`](https://packagist.org/packages/gplanchat/durable-bundle) brings durable execution to Symfony: long-running business logic written as a plain PHP class, journalled and replayed so it survives deploys, restarts and crashes. Workflow resumes and activity runs travel over Symfony Messenger, so the bundle needs transports and routing before it does anything at all — which is what this recipe sets up.

Docs: https://durable.rocks

### What the recipe writes

`config/packages/durable.yaml`, with the backend the bundle already defaults to (`in_memory` — no infrastructure, correct for a fresh install and for tests), plus the Messenger transports and the routing of the five message classes the bundle dispatches.

Two env vars, both defaulting to `sync://`. That is deliberate: `in-memory://` collects messages without handling them, so a fresh install would silently do nothing, and `doctrine://` would assume a transport dependency the project may not have. `sync://` handles them in-process, which matches the `in_memory` backend the config starts from. The post-install output says what to change when moving off memory.

### On the `framework.messenger` block

The best practices say a recipe should not modify another bundle's configuration, with one exception — appending to a config collection, the given example being a new DoctrineBundle connection. The transports and routing entries here are exactly that: additions to `framework.messenger.transports` and `framework.messenger.routing`, under names prefixed `durable_`.

Without them the bundle installs and does nothing, since every workflow resume and activity dispatch goes through Messenger. If you would rather this stayed out of the recipe, say so and I will drop it and move the block to the post-install output instead.

### `bundles`

The package is `"type": "symfony-bundle"`, so Flex registers it on its own when no recipe exists. The `bundles` key is here so that adding this recipe does not take that away.

### DIFF
--- a/gplanchat/durable-bundle/0.1/config/packages/durable.yaml
+++ b/gplanchat/durable-bundle/0.1/config/packages/durable.yaml
@@ -1,0 +1,23 @@
+durable:
+    # in_memory keeps the journal in the PHP process: right for tests, lost on restart.
+    # Use dbal to keep it in your database, or set temporal.dsn to reach a Temporal cluster.
+    event_store:
+        type: in_memory
+    workflow_metadata:
+        type: in_memory
+    activity_transport:
+        type: messenger
+        transport_name: durable_activities
+
+framework:
+    messenger:
+        transports:
+            durable_workflows: '%env(MESSENGER_DURABLE_WORKFLOW_DSN)%'
+            durable_activities: '%env(MESSENGER_DURABLE_ACTIVITY_DSN)%'
+
+        routing:
+            Gplanchat\Durable\Transport\ResumeWorkflowMessage: durable_workflows
+            Gplanchat\Durable\Transport\ActivityMessage: durable_activities
+            Gplanchat\Durable\Transport\FireWorkflowTimersMessage: sync
+            Gplanchat\Durable\Transport\DeliverWorkflowSignalMessage: sync
+            Gplanchat\Durable\Transport\DeliverWorkflowUpdateMessage: sync
--- a/gplanchat/durable-bundle/0.1/manifest.json
+++ b/gplanchat/durable-bundle/0.1/manifest.json
@@ -1,0 +1,27 @@
+{
+    "bundles": {
+        "Gplanchat\\Durable\\Bundle\\DurableBundle": ["all"]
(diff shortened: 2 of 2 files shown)


gplanchat:
About the red `Run updated recipes` job, so a reviewer does not have to dig through the log: the only failing combination is `sf6-php81`, and it fails before the recipe is reached.

```
- gplanchat/durable-bundle require php >=8.2
```

`callable-qa.yml` prints its own note on that exact step — *"You can ignore this error if your package does not support Symfony 6"* — and the package requires PHP 8.2, so that combination cannot resolve.

The other four all install and apply the recipe: `sf7-php82`, `sf6-php84`, `sf7-php84`, `sf8-php84`. The log shows `Configuring gplanchat/durable-bundle` four times, with the config files written and the post-install output rendering as intended.

Happy to adjust anything in the recipe itself.

gplanchat:
Tested end to end against the PR endpoint, on a fresh `symfony/skeleton:^7` with PHP 8.2 — the `sf7-php82` combination from the QA matrix.

`composer require gplanchat/durable-bundle:^0.1` writes:

**`config/bundles.php`**
```php
Gplanchat\Durable\Bundle\DurableBundle::class => ['all' => true],
```

**`.env`**
```
MESSENGER_DURABLE_WORKFLOW_DSN=sync://
MESSENGER_DURABLE_ACTIVITY_DSN=sync://
```

**`config/packages/durable.yaml`** — the `durable:` block plus the `framework.messenger` transports and routing, verbatim as in the recipe.

And it holds up once the container is built:

- `bin/console cache:clear` → OK, no configuration error
- `debug:config framework messenger` shows both `durable_workflows` and `durable_activities` transports resolved, and the five message classes routed
- `bin/console messenger:consume durable_workflows durable_activities` — the command the post-install output tells you to run — starts and consumes against both transports

So the recipe produces a working install with nothing left to edit by hand, which is what the `sync://` defaults were chosen for.

diimpp (approved):

PR #2048, opened 2026-09-09
Add recipe for indexnowkit/symfony-bundle

somework:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/indexnowkit/symfony-bundle

Recipe for [indexnowkit/symfony-bundle](https://github.com/indexnowkit/php/tree/main/packages/symfony-bundle) (0.15): a bundle that tells search engines (Yandex, Bing, Naver, Seznam, ...) through [IndexNow](https://www.indexnow.org) which pages changed when a Doctrine entity is committed.

What the recipe does:

- registers `IndexNowKit\SymfonyBundle\IndexNowKitBundle` for all environments;
- `config/packages/indexnowkit.yaml` with the two values the bundle cannot guess: the key and the public base URL, both from env;
- `config/routes/indexnowkit.yaml` importing the bundle's key-file route (`/<key>.txt`, which the engines fetch to verify ownership);
- `.env`: `INDEXNOW_KEY` generated by Flex (the key is served publicly at `/<key>.txt`, it is not a secret; the protocol accepts 8–128 chars of `[A-Za-z0-9-]`, so a generated hex secret is a valid key) and `INDEXNOW_BASE_URL` as a placeholder.

### DIFF
--- a/indexnowkit/symfony-bundle/0.15/config/packages/indexnowkit.yaml
+++ b/indexnowkit/symfony-bundle/0.15/config/packages/indexnowkit.yaml
@@ -1,0 +1,3 @@
+indexnowkit:
+    key: '%env(INDEXNOW_KEY)%'
+    base_url: '%env(INDEXNOW_BASE_URL)%'
--- a/indexnowkit/symfony-bundle/0.15/config/routes/indexnowkit.yaml
+++ b/indexnowkit/symfony-bundle/0.15/config/routes/indexnowkit.yaml
@@ -1,0 +1,2 @@
+indexnowkit:
+    resource: '@IndexNowKitBundle/config/routes.php'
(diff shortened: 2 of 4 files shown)


somework:
The only red step is "Create-project with skeleton ^6" on PHP 8.1: the bundle requires PHP ^8.2 (Symfony 6.4 itself is supported — the same skeleton ^6 step passes on PHP 8.4). Per the job's own note this can be ignored.

Tested the recipe on a fresh `symfony/skeleton` with `SYMFONY_ENDPOINT` pointing at `flex/pull-2048`: bundle registered, `config/packages/indexnowkit.yaml` and `config/routes/indexnowkit.yaml` copied, `.env` gets `INDEXNOW_KEY=` (dry run outside production until `indexnow:key:generate --write-env` fills `.env.local`) and `INDEXNOW_BASE_URL`; `GET /<key>.txt` answers 200 through the front controller.

diimpp (approved):

PR #2049, opened 2026-09-11
Add recipe for fyrst/shopware-cd

ROBJkE:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/fyrst/shopware-cd

## Package
- Packagist: https://packagist.org/packages/fyrst/shopware-cd (`1.1.0`+)
- GitHub: https://github.com/fyrst-dev/shopware-cd
- License: MIT

## What the recipe does
Flex metadata only: `copy-from-package` of `overlay/` → shop root, plus the bundle and env keys. CI, Compose, edge, and `.env.example` live in the Composer package (`fyrst/shopware-cd` `overlay/`), not in this recipe.

Same idea as `shopware/docker`, except the overlay is copied from the package instead of from the recipe.

Shopware projects typically already have `extra.symfony.allow-contrib: true` and `flex://defaults`.

## Discoverability
While this PR is in review, fyrst also keeps a private Flex endpoint that uses the same thin `copy-from-package` manifest:

https://raw.githubusercontent.com/fyrst-dev/recipes/flex/main/index.json

This contrib recipe is the preferred path once merged and indexed. The private endpoint is optional and independent of this PR.


diimpp (changes requested):
@ROBJkE Hi, please use [copy-from-package](https://github.com/symfony/recipes#copy-from-package-configurator)  and place all those scripts in your package.

Recipes are intended only to make installation easier, rather than to contain large chunks of the package.

ROBJkE:
@diimpp Thanks

The overlay now lives in the Composer package, not in this recipe:

- Package: https://packagist.org/packages/fyrst/shopware-cd (`1.1.0`+)
- Files: `overlay/` in https://github.com/fyrst-dev/shopware-cd (CI, Compose, edge, `.env.example`)
- This recipe is Flex metadata only: `copy-from-package` of `overlay/` → shop root, plus the existing bundle and env keys
- The old `root/` tree (including the deploy shell scripts) is gone from the PR

`manifest.json` is now:

```
{
    "copy-from-package": {
        "overlay/": ""
    }
}
```

Ready for another look.

diimpp (approved):

PR #2050, opened 2026-09-11
Add recipe for jpmmartin/sylius-nmi-plugin

jpmmartin:
| Q             | A
| ------------- | ---
| License       | MIT
| Packagist     | https://packagist.org/packages/jpmmartin/sylius-nmi-plugin

Recipe for the NMI payment gateway plugin for Sylius 2.2 (`type: sylius-plugin`), version 1.0. It registers the bundle, copies the configuration import and the route imports, adds the plugin's script to the store's shop entrypoint through the `@vendor` alias a Sylius Standard store defines (the same line the `flux-se/sylius-stripe-plugin` recipe adds), adds the ignore rules for a generated encryption key, and prints what no recipe can do. On purpose it does not change the store's encryption key path: a store may already hold other plugins' credentials encrypted with the current key.

The recipe is kept in the plugin's repository under the same layout as this one, compiled with `symfony-tools/recipes-checker` and applied to a fresh Sylius Standard store in the plugin's own CI on every change, after the same `lint:manifests` and `lint:yaml` this repository runs: https://github.com/jpmmartin/SyliusNmiPlugin/blob/main/.github/workflows/install.yaml

### DIFF
--- a/jpmmartin/sylius-nmi-plugin/1.0/config/packages/jpm_martin_sylius_nmi.yaml
+++ b/jpmmartin/sylius-nmi-plugin/1.0/config/packages/jpm_martin_sylius_nmi.yaml
@@ -1,0 +1,2 @@
+imports:
+    - { resource: "@JpmMartinSyliusNmiPlugin/config/config.yaml" }
--- a/jpmmartin/sylius-nmi-plugin/1.0/config/routes/jpm_martin_sylius_nmi.yaml
+++ b/jpmmartin/sylius-nmi-plugin/1.0/config/routes/jpm_martin_sylius_nmi.yaml
@@ -1,0 +1,18 @@
+jpm_martin_sylius_nmi_shop:
+    resource: "@JpmMartinSyliusNmiPlugin/config/routes/shop.yaml"
+
+jpm_martin_sylius_nmi_admin:
+    resource: "@JpmMartinSyliusNmiPlugin/config/routes/admin.yaml"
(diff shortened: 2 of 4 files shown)


jpmmartin:
For the reviewer, about the red `Run updated recipes` job: it fails here the way it does for every Sylius plugin recipe — see #1992 (flux-se/sylius-stripe-plugin) and #2044 (odiseoteam/sylius-rbac-plugin), both merged with the same result. That job installs the package into bare Symfony skeletons: the PHP 8.1 leg cannot satisfy `php ^8.2`, the Symfony 8 leg cannot satisfy `symfony/intl ^6.4 || ^7.4` (Sylius 2.2 runs on Symfony 7.4), and on the Symfony 7 legs the install succeeds but `cache:clear` fails because Sylius's bundles are never registered in a skeleton, so the `main` Messenger transport that Sylius core defines does not exist for the plugin's routing.

In a Sylius Standard store the recipe applies cleanly. It is tested that way on every change of the plugin, compiled with `symfony-tools/recipes-checker` and installed into a fresh `sylius/sylius-standard` store (the `recipe` leg of https://github.com/jpmmartin/SyliusNmiPlugin/blob/main/.github/workflows/install.yaml), and I verified it against this PR's `flex/pull-2050` endpoint as well: bundle registered, files copied, import added, `composer remove` leaves nothing behind.

diimpp (approved):

PR #2057, opened 2026-09-17
feat(ux-datatables): add the 1.0 recipe

pentiminax:
| Q | A |
| --- | --- |
| License | MIT |
| Packagist | https://packagist.org/packages/pentiminax/ux-datatables |

Adds a versioned recipe for UX DataTables 1.0, which renames the bundle entry point to
`PentiminaxDataTablesBundle` and the bundle resource alias to `@PentiminaxDataTablesBundle`.

The existing 0.43 recipe remains unchanged for pre-1.0 releases.

Related upstream change: https://github.com/pentiminax/ux-datatables/pull/468


pentiminax:
The structural QA check passes. The installation job currently fails because Composer cannot resolve pentiminax/ux-datatables:^1.0 yet; Packagist only exposes releases through v0.90.0. This job can be rerun after the 1.0 release is published.

